PR #16, opened 2025-03-22
Test Semantic Layer code. Add Levenshtein similarity metric.

DevonFulcher:
In this PR I refactored the Semantic Layer code so that it is easier to test, and I added a few unit and integration tests. I also added a Levenshtein similarity metric to identify when the LLM or user misspells a metric or dimension and suggest metrics and dimensions that are similar to the misspelling.

### DIFF
--- a/dbt_mcp/semantic_layer/client.py
+++ b/dbt_mcp/semantic_layer/client.py
@@ -1,0 +1,226 @@
+from contextlib import contextmanager
+from functools import cache, lru_cache
+from typing import Any, Generator, LiteralString, Protocol
+
+from dbtsl import SemanticLayerClient
+from dbt_mcp.config.config import Config
+from dbt_mcp.semantic_layer.levenshtein import get_misspellings
+from dbt_mcp.semantic_layer.types import DimensionToolResponse, MetricToolResponse
+from dbtsl.models.dimension import Dimension
+from dbtsl.models.metric import Metric
+import time
+import requests
+
+class SemanticLayerClientProtocol(Protocol):
+    @contextmanager
+    def session(self) -> Generator[Any, Any, None]:
+        ...
+
+    def metrics(self) -> list[Metric]:
+        ...
+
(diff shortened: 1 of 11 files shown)


dbt_mcp/semantic_layer/client.py
```diff
@@ -0,0 +1,226 @@
...
+        limit_section = f"limit: {limit}" if limit else ""
+
+        # Build create query mutation with direct string construction
+        mutation = f"""
+        mutation {{
+        createQuery(
+            environmentId: "{self.config.environment_id}"
+            metrics: [{metric_list}]
+            {group_by_section}
+            {limit_section}
```
alan-andrade, line 135: nit: maybe from working in JS too much I got used to the ternaries within the usage spot. I do think it _could_ be nice to have the param name present here, versus embbeded in the strings above. I think it could be good to have an idea of the shape of the gqp query.

```
{ "groupBy: " + ", ".join(groups) if groups else "" }
```

tests/unit/test_semantic_layer.py
```diff
@@ -0,0 +1,24 @@
...
+            queryable_granularities=[],
+            queryable_time_granularities=[],
+            label="Revenue",
+            requires_metric_time=False
+        )],
+    )
+    fetcher = SemanticLayerFetcher(mock_client, "fake", MagicMock())
+    result = fetcher.validate_query_metrics_params(["revehue"], None)
+    assert result is not None
+    assert "revenue" in result
```
alan-andrade, line 24: :NICE:

dbt_mcp/semantic_layer/client.py
DevonFulcher: This code is just copied and pasted from the preexisting implementation. I plan to remove all of it and use the SL Python SDK instead. I'm just working on fixing a small issue in the SDK first. I don't want to mess with it too much since it is working and will be removed soon anyway.

PR #25, opened 2025-04-01
configure and run ruff and fix mypy errors

reneruck-dbt:
all formatting, no real changes

### DIFF
--- a/dbt_mcp/semantic_layer/client.py
+++ b/dbt_mcp/semantic_layer/client.py
@@ -9,15 +9,6 @@ from dbt_mcp.semantic_layer.levenshtein import get_misspellings
 from dbt_mcp.semantic_layer.types import DimensionToolResponse, MetricToolResponse
 
 
-# class SemanticLayerClientProtocol(Protocol):
-#     @contextmanager
-#     def session(self) -> Generator[Any, Any, None]: ...
-
-#     def metrics(self) -> list[Metric]: ...
-
-#     def dimensions(self, metrics: list[str]) -> list[Dimension]: ...
-
-
 class SemanticLayerFetcher:
     def __init__(self, sl_client: SemanticLayerClient, host: str, config: Config):
         self.sl_client = sl_client
--- a/uv.lock
+++ b/uv.lock
@@ -82,6 +82,15 @@ wheels = [
 ]
 
 [[package]]
+name = "cfgv"
+version = "3.4.0"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


.github/actions/setup-python/action.yml
```diff
@@ -0,0 +1,33 @@
...
+  using: "composite"
+  steps:
+    - name: Setup python
+      id: setup-python
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: setup ssh
+      uses: dbt-labs/internal-actions/private-repo-ssh-setup@v1
```
DevonFulcher, line 21: This repo will be public. So, we should remove this.

DevonFulcher: Also this part:
```
  ssh-private-key:
    description: Private Key required for installing private packages
    required: true
 ```

File: .pre-commit-config.yaml [hunk 2]
```diff
@@ -0,0 +1,34 @@
+exclude: |
+  (?x)^(
+      .mypy_cache/
+      | .devspace/
```
DevonFulcher, line 4: We can remove `.devspace` from this list.

.github/actions/setup-python/action.yml
reneruck-dbt: oh, right

pyproject.toml
```diff
@@ -7,9 +7,16 @@ requires-python = ">=3.12"
 dependencies = [
   "dbt-sl-sdk[sync]>=0.8.0",
   # We are using some internal functions, so pinning this dependency.
-  "mcp[cli]==1.5.0",
+  "mcp[cli]==1.6.0",
   "pytest>=8.3.5",
   "pytest-asyncio>=0.25.3",
```
DevonFulcher, line 12: Whoops, I should have put these in dev dependencies too.

dbt_mcp/semantic_layer/client.py
```diff
@@ -1,35 +1,20 @@
...
 from dbtsl import SemanticLayerClient
-from dbtsl.models.dimension import Dimension
-from dbtsl.models.metric import Metric
 
 from dbt_mcp.config.config import Config
 from dbt_mcp.semantic_layer.levenshtein import get_misspellings
 from dbt_mcp.semantic_layer.types import DimensionToolResponse, MetricToolResponse
 
 
-class SemanticLayerClientProtocol(Protocol):
```
DevonFulcher, line 16: I think I added this for tests, but if they pass without it, even better!

PR #27, opened 2025-04-01
Support getting entities from semantic layer

DevonFulcher:
New get_entities tool.

image

### DIFF
--- a/dbt_mcp/semantic_layer/tools.py
+++ b/dbt_mcp/semantic_layer/tools.py
@@ -1,6 +1,6 @@
 from dbt_mcp.config.config import Config
 from dbt_mcp.semantic_layer.client import get_semantic_layer_fetcher
-from dbt_mcp.semantic_layer.types import MetricToolResponse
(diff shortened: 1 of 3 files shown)


dbt_mcp/semantic_layer/tools.py
```diff
@@ -25,10 +25,20 @@ def get_dimensions(metrics: list[str]) -> list[DimensionToolResponse]:
...
         Args:
-            metrics: List of metric names or a single metric name
+            metrics: List of metric names
         """
         return semantic_layer_fetcher.get_dimensions(metrics=metrics)
 
+    @dbt_mcp.tool()
+    def get_entities(metrics: list[str]) -> list[EntityToolResponse]:
+        """
+        Get entities for a metric
```
reneruck-dbt, line 35: Is this enough docstring for the LLM to use it when not directly asked about metric entities?

DevonFulcher: Ya, good call. This docstring is similar to others in this repo. We should go through all of them and add more details.

PR #28, opened 2025-04-01
Publishing to pypi

DevonFulcher:
After release, I will trigger the release workflow, and we will have our first version published on PyPi! More context [here](https://dbt-labs.slack.com/archives/C05FWBP9X1U/p1743534235487449).


pyproject.toml
```diff
@@ -1,15 +1,22 @@
...
 readme = "README.md"
+license = { file = "LICENSE" }
 requires-python = ">=3.12"
+classifiers = [
+  "License :: OSI Approved :: Apache Software License",
+  "Programming Language :: Python :: 3",
+  "Operating System :: OS Independent",
+  "Development Status :: 4 - Beta",
+  "Intended Audience :: Developers",
+  "Operating System :: OS Independent",
```
emmyoop: Duplicated os

```suggestion
  "Operating System :: OS Independent",
  "Development Status :: 4 - Beta",
  "Intended Audience :: Developers",
```

pyproject.toml
```diff
@@ -1,15 +1,22 @@
 [project]
 name = "dbt-mcp"
 version = "0.1.0"
-description = "Add your description here"
+description = "A MCP (Model Context Protocol) server for interacting with dbt resources."
```
emmyoop, line 4: let's add author too
```suggestion
description = "A MCP (Model Context Protocol) server for interacting with dbt resources."
authors = [
  {name = "dbt Labs"}
]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install go-task
+        run: sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
+      - name: Run check
+        run: task check
+      - name: Run tests
+        run: task test:unit
+      - name: Build
+        run: task build
+      - name: Publish
+        run: task publish
```
emmyoop: I don't think this will work.  The trusted publisher is specifically setup to work with GitHub actions.  You need to use the `pypa/gh-action-pypi-publish` action.  I've never used uv to do this though and the [docs](https://docs.astral.sh/uv/guides/package/#publishing-your-package) aren't clear to me so maybe I'm wrong?

emmyoop: Additionally, you need to set the [environment](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) to `pypi` (production) or `pypiTest` to match what we set up the trusted publisher to use.

emmyoop: So for pypi test it would look like this.  You can parameritize this so if you're doing a test release it goes to pypiTest, otherwise to pypi.

```
    publish:
        runs-on: ubuntu-24.04
        environment:
            name: pypiTest
            url: https://test.pypi.org/legacy/
...
      - name: Publish
        uses: pypa/gh-action-pypi-publish@release/v1
```

DevonFulcher: Okay, got it, thanks. I misunderstood the uv docs. As you mentioned, they aren't very clear.

Taskfile.yml
```diff
@@ -79,3 +79,13 @@ tasks:
...
+
+  build:
+    desc: "Build the package"
+    cmds:
+      - uv build
+
+  publish:
+    desc: "Publish the package to PyPI"
+    cmds:
+      - uv publish
```
emmyoop, line 91: Does this do anything now?

DevonFulcher: Good call! I'll remove it.

dist/dbt_mcp-0.1.0-py3-none-any.whl
emmyoop, line 1: To be slightly more accurate I'd rename this as `dbt_mcp-0.1.0a1-py3-none-any.whl` (same for the next file) to indicate 0 guarantee anything works.  Then you're also not using up minor version 1.

DevonFulcher: Thank you!

PR #37, opened 2025-04-10
Update README and add arch diagram

b-per:
Add a bit more context at the top of the README as well as a diagram (generated from d2) showing how the tools interact with dbt


docs/diagram.d2
```diff
@@ -0,0 +1,64 @@
...
+    build
+    compile
+    docs
+    ls
+    parse
+    run
+    test
+  }
+
+  discovery: Discovery\nMetadata API or local artifacts {
```
DevonFulcher, line 29: I'm not sure what you are referring to as "local artifacts" here. The only local components are the dbt Core tools.

docs/diagram.d2
```diff
@@ -0,0 +1,64 @@
...
+    label.near: outside-right-center
+    class: container
+    get_mart_models
+    get_all_models
+    get_model_details
+    get_model_parents
+  }
+
+  sl: Semantic Layer\ndbt Cloud SL gateway {
```
DevonFulcher, line 38: I think it would be a bit more accurate to simply have this as `dbt Semantic Layer. '

docs/diagram.d2
jasnonaz: We don't have this right now but would love if we could look into using the local artifacts. Probably this should be set in the env vars - either use the local artifacts or disco api for this. May also be separate tools - local artifacts if its a dev use case or disco API if its a data consumption use case.

PR #40, opened 2025-04-10
Prompt updates and evals for semantic layer

DevonFulcher:
This PR improves the semantic layer query experience by adding an improved prompt as well as some updates to the SL SDK and how it is used.

Additionally, I included some evals here. These evals should be fleshed out some more. They often don't pass. However, in my testing with Claude the SL queries perform much better.

In the `dbt_mcp/semantic_layer/sdk` directory, I copy-pasted some code from the SL SDK. I am going to sync with Lucas when he gets back from PTO on how we can move these changes into the SDK itself.


dbt_mcp/semantic_layer/tools.py
```diff
@@ -40,7 +53,12 @@ def get_dimensions(metrics: list[str]) -> list[DimensionToolResponse]:
     @dbt_mcp.tool()
     def get_entities(metrics: list[str]) -> list[EntityToolResponse]:
         """
-        Get entities for a metric
+        Get the entities for specified metrics
+
+        Entities are real-world concepts in a business such as customers,
```
alan-andrade, line 58: nit/comment: Makes me wonder if this would bias the model at all, or if it can deduce "real-world" correctly 💭

DevonFulcher: I got that language from our docs. I can remove it.

alan-andrade: It's just a thought, we need to check against the results

PR #48, opened 2025-04-11
Releasing to pypi package

DevonFulcher:
This should allow us to run the `release.yml` action and release a version to pypi.


tanlen0 (approved):
LGTM

PR #52, opened 2025-04-14
Updated arch image from relative path to github asset

tanlen0:


DevonFulcher (approved):
Awesome thanks! That URL works for me

README.md
```diff
@@ -9,7 +9,7 @@ In its current form, it allows users to:
 
 ## Architecture
 
-![architecture diagram of the dbt MCP server](docs/d2.png)
```
DevonFulcher, line 12: Should the text within `[ ]` stay the same?

tanlen0: Good call, updated

PR #62, opened 2025-04-17
Add configuration for user id and develop environment id

DevonFulcher:
Support for calling the SQL execution API requires a develop environment ID and user ID. This PR adds these as environment variables and passes them as headers through remote tools.


alan-andrade:
This will also be necessary for the authz work. Thanks Devon!

PR #68, opened 2025-04-18
Fix install script one-liner

reneruck-dbt:


alan-andrade (approved):
can confirm it points to a usable install.sh

PR #75, opened 2025-04-21
Add development environment

legomb:
This PR adds a portable development environment to the repo. This means there is no need to set up python, uv, etc. in your local machine.

- Add [devbox](https://www.jetify.com/devbox) support
- Add [direnv](https://direnv.net/) support for automatic use of devbox when you land on this dir
- Add VSCode devcontainer support

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+FROM jetpackio/devbox:latest
+
+# Installing your devbox project
+WORKDIR /code
+USER root:root
+RUN mkdir -p /code && chown ${DEVBOX_USER}:${DEVBOX_USER} /code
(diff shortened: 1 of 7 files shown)


DevonFulcher:
Hey @legomb, thanks a bunch for opening this PR! It looks very thorough and well-structured. Have you had a problem with the existing developer experience described in the README.md and CONTRIBUTING.md files? I'd be happy to help you troubleshoot any problems you encounter.

While I understand the intent of using devbox to create isolated dev environments, I am worried about adding and maintaining too many separate developer workflows. You can see in this repo the `.tool-versions` file, which I use with [asdf](https://asdf-vm.com/). asdf and devbox provide similar functionality. Would it help if we added these files to the `.gitignore` so that you can keep your developer workflow uninterrupted?

PR #76, opened 2025-04-21
Stateless Remote MCP support

DevonFulcher:
The current implementation of the Python SDK is stateful. There is work underway to fix this: https://github.com/modelcontextprotocol/python-sdk/issues/443. However, currently we need to eject from the MCP framework to use stateless requests. This is to avoid issues described in this issue: https://github.com/modelcontextprotocol/python-sdk/issues/520, for example. I tested this in Claude by listing and making a request to a remote tool.


alan-andrade (approved):
:shipit:

PR #82, opened 2025-04-22
Support any Python version >= 3.12

DevonFulcher:
Previously, we couldn't allow for Python version 3.13 because the SL SDK didn't allow it. The SL SDK now supports Python 3.13, so this PR updates the install script to allow for it.


tanlen0 (approved):
LGTM

PR #84, opened 2025-04-22
Pin all dependencies

DevonFulcher:
Our current install process is with pip in `install.sh`. `install.sh` makes for an easy installation process, but using pip can mean that users can get different installations of dbt-mcp. The best way to fix this would be to install the app from the `uv.lock` file, not with pip. However, this would require that users have uv installed. In the future, we can use [uv export](https://github.com/astral-sh/uv/issues/12584?utm_source=chatgpt.com) and install from pylock.toml with pip, but [that isn't supported yet](https://github.com/pypa/pip/issues/13334). In the meantime, pinning all dependencies should help provide for a more consistent install.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,12 +14,13 @@ classifiers = [
   "Intended Audience :: Developers",
 ]
 dependencies = [
+  # Pinning all dependencies because this app is installed with pip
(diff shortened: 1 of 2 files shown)


alan-andrade (approved):
📌

PR #91, opened 2025-04-23
Make package version based on git tag

b-per:
With this change we shouldn't need to update the version by hand anymore and doing a release in GH will publish the package with the version being the same as the tag used

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,11 @@
 [project]
 name = "dbt-mcp"
-version = "0.1.3"
 description = "A MCP (Model Context Protocol) server for interacting with dbt resources."
 authors = [{ name = "dbt Labs" }]
 readme = "README.md"
 license = { file = "LICENSE" }
 requires-python = ">=3.12"
+dynamic = ["version"]
 classifiers = [
   "License :: OSI Approved :: Apache Software License",
   "Programming Language :: Python :: 3",
@@ -43,8 +43,12 @@ Source = "https://github.com/dbt-labs/dbt-mcp"
 dbt-mcp = "dbt_mcp.main:main"
 
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
 
 [tool.hatch.build.targets.sdist]
 include = ["src/dbt_mcp/**/*", "README.md", "LICENSE"]
+
+[tool.hatch.version]
+source = "vcs"
+
--- a/uv.lock
+++ b/uv.lock
@@ -152,7 +152,6 @@ wheels = [
 
 [[package]]
 name = "dbt-mcp"
-version = "0.1.3"
 source = { editable = "." }
 dependencies = [
     { name = "dbt-sl-sdk", extra = ["sync"] },


reneruck-dbt:
What is setting the git tag and by what pattern?
How does this change the current release process?

reneruck-dbt:
What is setting the git tag and by what pattern?
How does this change the current release process?

b-per:
Like we do in other libraries, for example, [dbt-jobs-as-code](https://github.com/dbt-labs/dbt-jobs-as-code/releases)

We do releases on GitHub, and at the same time set some tags on the commit for the given release. In `dbt-jobs-as-code` for example, with [this GH action](https://github.com/dbt-labs/dbt-jobs-as-code/blob/main/.github/workflows/release.yml) we then get a new version of the package pushed to PyPi every time we do a new release in GH.

b-per:
I added a few workflow files to mimic the process used in https://github.com/dbt-labs/semantic-layer-sdk-python

The flow for a release then becomes:
- someone would trigger the action `create-release-pr` mentioning if it should be a patch minor or major
- this will merge all the changie messages into a new changelog and create a PR with the new file
- when the PR is created, it triggers `create-release-tag` to set the git tag to the new version
- when a tag is created, we push a new version of the package to PyPi

b-per:
This whole process wouldn't create a GH release automatically, but we could use the `gh` CLI if we wanted with something like

```
export VERSION=$(changie latest)
gh release create "$VERSION" --notes-file .changes/$VERSION.md
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      # TODO: Uncomment this when we have tests
+      # - name: Run tests
+      #   run: task test:unit
+      
+      - name: Build
+        run: task build
+
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
DevonFulcher: In order to publish to pypi. I think this action needs to be called release.yml like the existing release action. This is a security feature of pypi.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,65 @@
+name: Publish dbt-mcp to PyPI
+on:
+  workflow_dispatch:
```
DevonFulcher: I would expect there to be only one `workflow_dispatch` action to trigger a release, right?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: setup python
+        uses: ./.github/actions/setup-python
+        id: setup-python
+
+      - name: Install go-task
+        run: sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
+
+      # TODO: Uncomment this when we have tests
+      # - name: Run tests
+      #   run: task test:unit
```
DevonFulcher: We have tests now, so we can uncomment this.

.github/workflows/publish.yml
DevonFulcher: I think this action also needs this environment to be able to publish as well: https://github.com/dbt-labs/dbt-mcp/blob/main/.github/workflows/release.yml#L21

b-per: I just deleted the old `release.yml` then

b-per:
Not too sure what is going on with "Check styling (pull_request)" as it all passes locally...

.github/workflows/release.yml
```diff
@@ -1,49 +1,65 @@
...
         run: task test:unit
+      
       - name: Build
         run: task build
-      - name: Publish to test PyPI
-        if: inputs.pypi-server == 'test-pypi'
-        uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc
+
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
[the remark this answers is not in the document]
DevonFulcher: non-blocking: you will want to use the git SHA to dismiss this security warning instead of `release/v1`, like in the original version.

.github/workflows/release.yml
```diff
@@ -1,49 +1,65 @@
...
           packages-dir: ./dist
-          repository-url: https://test.pypi.org/legacy/
-      - name: Publish to PyPI
-        if: inputs.pypi-server == 'pypi'
-        uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc
+
+  create-github-release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # required to create a releas
```
DevonFulcher:
```suggestion
      contents: write # required to create a release
```

.github/workflows/release.yml
```diff
@@ -1,49 +1,65 @@
...
+          version: latest
+          args: latest
+
+      - name: Create GitHub Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.changie-latest.outputs.version }}
+        run: |
+          echo "Creating release for version: $VERSION"
+          gh release create "$VERSION" \
```
DevonFulcher, line 63: Is this already automatically installed in the action?

DevonFulcher (approved):
Amazing job!

.github/workflows/release.yml
b-per: Yes

> GitHub CLI is preinstalled on all GitHub-hosted runners. For each step that uses GitHub CLI, you must set an environment variable called GH_TOKEN to a token with the required scopes.

.github/workflows/release.yml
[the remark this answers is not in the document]
b-per: We could update all the actions, yes, but let's do it after we see if the release flow works

PR #99, opened 2025-04-28
dropping the installer script

reneruck-dbt:


DevonFulcher:
We would also need to update the JSON in the readme too, right? Also, in this case, we would need to bring back the git clone instructions from the CONTRIBUTING.md file.

DevonFulcher:
We would also need to update the JSON in the readme too, right?

PR #103, opened 2025-05-06
Add new tool to get children models

### LINKED ISSUE #101: Feature - Add a tool called `get_model_children`
(from the community Slack)

We currently provide a way to retrieve the parents of models but not the other way around.

Having the ability to get children would allow identifying "dangling" models that don't fee any downstream mart tables.

b-per:
Fixes #101


b-per:
Screenshot from Claude Desktop with the new tool
image

reneruck-dbt (approved):

Issue #107, opened 2025-05-12
Feature: support for cli args


Skorch:
Currently only the `dbt show` cli command supports args (inline sql)

I would like to see all of the commands supports a `select` arg at the very least.

This is critical for agentic coding - where you need to build/test single models as quickly as possible.

Example:
```python
    @mcp.tool(description=get_prompt("dbt_cli/test"))
    async def test(
        selector: Optional[str] = Field(
            default=None,
            description=get_prompt("dbt_cli/args/selector")
        ) -> str
```
(note:  the use of the pydantic `Field` type ensures the the MCP server (at least in cline/roo) pick up parameter descriptions.)
See: https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/tools/base.py

DevonFulcher:
@Skorch Thanks for the idea! @b-per has implemented this for the `--select` argument. Are there other arguments that would be beneficial, or should we close this task?

DevonFulcher:
I'm going to close this for now as it is completed. Please let us know if you have any more ideas.

PR #109, opened 2025-05-14
POC to add the ability to use selectors + other small things

b-per:
A few changes here, likely not to be merged but can serve as a base for future dev
- tries to implement #107 
- add output to "OK" even in quiet mode
- remove `--quiet` if a selector is provided (likely need to be tweaked, see code comments)

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -1,22 +1,38 @@
 import subprocess
+from typing import Optional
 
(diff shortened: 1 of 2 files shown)


BrJan:
Love seeing this @b-per! This closes a big current gap in the MCP.

DevonFulcher:
@b-per I believe this was completed in a separate PR, right? Should we close this one?

b-per:
Yes, the `--quiet` removal part wasn't added, but we can just close it for now.

PR #113, opened 2025-05-19
Refactor config for modularity

DevonFulcher:
This PR refactors the Config class so that each tool type has its own config class. This helps with modularity, and it makes it more apparent which fields are required for each type of tool. I tested each tool type in Claude.


alan-andrade (approved):
This looks great, easier to understand the configs for each tool. Seems like we just need some test udpates ✅

PR #116, opened 2025-05-21
Usage tracking scaffolding

DevonFulcher:
This PR sets up the codebase to emit usage tracking. This PR doesn't emit events yet, but it should be straightforward to add that soon. This is implemented by overriding the `call_tool` method and emitting events after each tool call. This method is also used to catch exceptions, so I was also able to remove some redundant try-except blocks.

### DIFF
--- a/src/dbt_mcp/mcp/server.py
+++ b/src/dbt_mcp/mcp/server.py
@@ -1,15 +1,24 @@
 import asyncio
 import logging
-from collections.abc import AsyncIterator
-from contextlib import asynccontextmanager
+from collections.abc import AsyncIterator, Sequence
+from contextlib import (
+    asynccontextmanager,
+)
+from typing import Any
 
 from mcp.server.fastmcp import FastMCP
+from mcp.types import (
+    EmbeddedResource,
+    ImageContent,
+    TextContent,
+)
 
 from dbt_mcp.config.config import load_config
 from dbt_mcp.dbt_cli.tools import register_dbt_cli_tools
 from dbt_mcp.discovery.tools import register_discovery_tools
 from dbt_mcp.remote.tools import register_remote_tools
 from dbt_mcp.semantic_layer.tools import register_sl_tools
+from dbt_mcp.tracking.tracking import UsageTracker
 
 config = load_config()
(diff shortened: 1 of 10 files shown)


src/dbt_mcp/mcp/server.py
```diff
@@ -27,7 +36,45 @@ async def app_lifespan(server: FastMCP) -> AsyncIterator[None]:
...
+                    text=str(e),
+                )
+            ]
+        self.usage_tracker.emit_tool_called_event(
+            config=config.tracking_config,
+            tool_name=name,
+            arguments=arguments,
+            error_message=None,
+        )
+        return result
```
alan-andrade, line 74: NIce!

src/dbt_mcp/semantic_layer/tools.py
```diff
@@ -23,39 +23,27 @@ def register_sl_tools(dbt_mcp: FastMCP, config: SemanticLayerConfig) -> None:
...
-            end_time = time()
-            logger.info(f"get_entities took {end_time - start_time} seconds")
-            return result
-        except Exception as e:
-            logger.error(f"Error getting entities: {e}")
-            return str(e)
+        start_time = time()
+        result = semantic_layer_fetcher.get_entities(metrics=metrics)
+        end_time = time()
+        logger.info(f"get_entities took {end_time - start_time} seconds")
```
alan-andrade, line 45: super nitpick: Feels like these 4 instances of timing measurement blocks could be refactored? not a big deal though

alan-andrade (approved):
👌 Nice!

src/dbt_mcp/semantic_layer/tools.py
DevonFulcher: Do you mean to put them in the `call_tool` function? Originally, this timing was very focused just on timing SL latency to help debug a specific issue. It may be worth removing them now that the issue has been fixed.

alan-andrade: I mean using a function/context/we that takes care of the timing the operation. If it's not going to stay for long, then we don't bother.

PR #117, opened 2025-05-22
Increase remote tools timeout

DevonFulcher:
The execute_sql & text_to_sql tools can experience some high-latency requests. Just increasing the timeout here.


b-per (approved):
I haven't tested remote tools personally but the change makes sense

PR #120, opened 2025-05-27
Add selector for dbt commands

b-per:
Replaces #109 

Works locally
image

To be discussed:
- should we add tests on it, if so how/which

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -1,15 +1,21 @@
 import subprocess
+from typing import Optional
 
 from mcp.server.fastmcp import FastMCP
+from pydantic import Field
 
 from dbt_mcp.config.config import DbtCliConfig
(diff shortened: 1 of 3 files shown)


src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -26,10 +35,14 @@ def _run_dbt_command(command: list[str]) -> str:
...
     @dbt_mcp.tool(description=get_prompt("dbt_cli/build"))
-    def build() -> str:
+    def build(
+        selector: Optional[str] = Field(
+            default=None, description=get_prompt("dbt_cli/args/selectors")
+        ),
+    ) -> str:
         return _run_dbt_command(["build"])
 
     @dbt_mcp.tool(description=get_prompt("dbt_cli/compile"))
```
DevonFulcher, line 48: Should selector be added to dbt compile too?

PR #121, opened 2025-05-28
Install hatch in release process

DevonFulcher:
Trying to fix this error: https://github.com/dbt-labs/dbt-mcp/actions/runs/15288639277/job/43003917936

### DIFF
--- a/.changes/unreleased/Under the Hood-20250528-080437.yaml
+++ b/.changes/unreleased/Under the Hood-20250528-080437.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Install hatch in release process
+time: 2025-05-28T08:04:37.075036-05:00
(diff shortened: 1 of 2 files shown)


b-per:
I got the original action from [here](https://github.com/dbt-labs/semantic-layer-sdk-python/blob/6c61837d8409f392bda0fc57e4471484032f19ec/.github/workflows/create-release-pr.yaml) and yes, their `setup-python-env` step installs hatch. 

This should solve this issue.

b-per:
And, to be honest, I think that running `hatch version "$VERSION_NO_V"` is not required as:
- we get the new version from changie [here](https://github.com/dbt-labs/dbt-mcp/blob/3592fb2c4dd7e1d92af03e1127342527978f5bc3/.github/workflows/create-release-pr.yml#L37C15-L42)
- we define our tool as `dynamic` [here](https://github.com/dbt-labs/dbt-mcp/blob/3592fb2c4dd7e1d92af03e1127342527978f5bc3/pyproject.toml#L8)

But we can revisit when we get a successful release process end to end.

PR #122, opened 2025-05-28
Update diagram to local one

b-per:
Right now we are pointing at an old version that was copy/paster instead of the latest image generated

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This MCP (Model Context Protocol) server provides tools to interact with dbt. Re
 
 ## Architecture
 
-![architecture diagram of the dbt MCP server](https://github.com/user-attachments/assets/89b8a24b-da7b-4e54-ba48-afceaa56f956)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ This MCP (Model Context Protocol) server provides tools to interact with dbt. Re
 
 ## Architecture
 
-![architecture diagram of the dbt MCP server](https://github.com/user-attachments/assets/89b8a24b-da7b-4e54-ba48-afceaa56f956)
+![architecture diagram of the dbt MCP server](./docs/d2.png)
```
DevonFulcher: I don't think this will appear in pypi correctly. Can we continue hosting the image in `user-attachments`?

DevonFulcher (approved):

README.md
b-per: I didn't think about it. I'll do some research to see how people handle it

b-per: I updated the image to point to the http location

PR #129, opened 2025-05-28
Update instrcuctions for running the server

b-per:
Update instructions so that we don't require cloning the repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,23 +8,8 @@ This MCP (Model Context Protocol) server provides tools to interact with dbt. Re
 
 ## Setup
 
-1. Clone the repository:
-```shell
-git clone https://github.com/dbt-labs/dbt-mcp.git
-cd dbt-mcp
-```
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,19 +54,18 @@ This configuration will be added to the respective client's config file. Be sure
  {
   "mcpServers": {
     "dbt-mcp": {
-      "command": "<path-to-mcp-executable>",
+      "command": "uvx",
       "args": [
-        "run",
-        "<path-to-this-directory>/src/dbt_mcp/main.py"
+        "--env-file",
```
DevonFulcher, line 59: I think we would want to use a syntax with `env` like here:

```
{
  "mcpServers": {
    "server-name": {
      "command": "python",
      "args": ["mcp-server.py"],
      "env": {
        "API_KEY": "value"
      }
    }
  }
}
```

DevonFulcher: Actually, I think the way you have done it is okay for now

PR #130, opened 2025-05-28
Logging usage events

DevonFulcher:
This PR sends ToolCalled usage events to our Vortex eventing system. I also cleaned up some unnecessary logs for SL tools.

I tested this by running Vortex locally and sending events to our prod instance as well.

### DIFF
--- a/src/dbt_mcp/mcp/server.py
+++ b/src/dbt_mcp/mcp/server.py
@@ -1,11 +1,13 @@
 import asyncio
 import logging
+import time
 from collections.abc import AsyncIterator, Sequence
 from contextlib import (
     asynccontextmanager,
 )
 from typing import Any
 
+from dbtlabs_vortex.producer import shutdown
 from mcp.server.fastmcp import FastMCP
 from mcp.types import (
     EmbeddedResource,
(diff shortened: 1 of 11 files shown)


src/dbt_mcp/mcp/server.py
```diff
@@ -65,10 +75,14 @@ async def call_tool(
                     text=str(e),
                 )
             ]
+        end_time = int(time.time() * 1000)
+        logger.info(f"Tool {name} called successfully in {end_time - start_time}ms")
         self.usage_tracker.emit_tool_called_event(
```
reneruck-dbt, line 80: What happens if this call takes very long or fails all together?
Should that fail the whole tool call?

DevonFulcher: So, emit_tool_called_event is a thin wrapper around log_proto which already catches all exceptions here: https://github.com/dbt-labs/vortex/blob/main/clients/python/dbtlabs_vortex/producer/client.py#L180

DevonFulcher: I'll add another catch within emit_tool_called_event just in case.

reneruck-dbt: I'm wondering if we want to push that operation into a co-routine and don't wait for its completion.
The request would be completely out of the hot path.

DevonFulcher: So, `log_proto` just queues up events to be processed by a background thread. So, there isn't an async version, and the code doesn't block on the request.

reneruck-dbt (approved):

PR #142, opened 2025-06-03
Add missing selector and update prompt

b-per:
- fix bug with missing selector
- update prompt for building selectors as it was missing `--select` sometimes

### DIFF
--- a/src/dbt_mcp/prompts/dbt_cli/args/selectors.md
+++ b/src/dbt_mcp/prompts/dbt_cli/args/selectors.md
@@ -1,4 +1,5 @@
-It is important to add the `--select` before the selector if any selector needs to be provided
+It is important to add the `--select` before the selector if any selector needs to be provided.
+If you provide a selector, DO NOT forget adding `--select` in front!
 
 - to select all models, just do not provide a selector.
 - to select a particular model, use the selector `--select <model_name>`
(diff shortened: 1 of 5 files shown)


b-per:
We should do a new patch release with this fix when it is merged

src/dbt_mcp/prompts/dbt_cli/args/selectors.md
```diff
@@ -1,4 +1,5 @@
-It is important to add the `--select` before the selector if any selector needs to be provided
+It is important to add the `--select` before the selector if any selector needs to be provided.
+If you provide a selector, DO NOT forget adding `--select` in front!
```
reneruck-dbt, line 2: I'm not sure about the use of `DO NOT` here.

DevonFulcher: Logically, I think it is correct. However, they may be better language or we may change up our prompt/implementation to not require that the LLM provides the `--select` prefix.

b-per: I created a follow-up issue for this. I will merge it as-is for now as the current implementation just doesn't work.

dyanarose: peaking in on ` to not require that the LLM provides the --select prefix.`

As we were testing the MCP server yesterday (and found the missing selector arg bug), it *felt* wrong/redundant that the `selectors` field needed to have the `--select` flag in it.

b-per: @dyanarose , what application and model are you using to integrate with the MCP server?

I had similar issues where `--select` was not added, but it is quite random. I will run some checks, adding the `--select` in the code rather than in the prompt to see if it works better.

PR #146, opened 2025-06-03
Added argparse support to dbt_mcp main.py with env_file argument

### LINKED ISSUE #145: Fix: --env-file arg is not parsed
The current README.md states that `--env-file` should be passed as an arg; however, that arg is not currently parsed

appleseed-iii:
fixes #145

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -57,8 +57,8 @@ class Config:
     semantic_layer_config: SemanticLayerConfig | None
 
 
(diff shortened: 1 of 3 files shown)


DevonFulcher:
Thanks for the contribution! However, see my comment here https://github.com/dbt-labs/dbt-mcp/issues/145. I don't think we need to add this CLI argument.

DevonFulcher:
Thanks again for this. I'm going to close it for now.

PR #150, opened 2025-06-04
FIX - check disable_discovery flag before discovery config

appleseed-iii:
a quick one:
I'm assuming ya'll intend to check the `disable_discovery` arg before the discovery_config rather than `disable_dbt_cli`.

The reason this came up is that I don't have access to the discovery api since I don't have an enterprise account. I'd still like to keep the dbt_cli enabled, but do to the logic below i was also getting all the discovery tools enabled even though they aren't accessible on my account.

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -149,7 +149,7 @@ def load_config() -> Config:
         )
 
     discovery_config = None
-    if not disable_dbt_cli and host and actual_prod_environment_id and token:
+    if not disable_discovery and host and actual_prod_environment_id and token:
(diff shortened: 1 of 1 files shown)


DevonFulcher (approved):
Awesome! Thanks for the bug fix

DevonFulcher:
You will want to install [changie](https://changie.dev/), run `changie new`, and commit the file that is generated.

DevonFulcher:
Thanks for the contribution @appleseed-iii! I think this is an important fix, and I wanted to get it in quickly, so I already merged a fix here: https://github.com/dbt-labs/dbt-mcp/pull/156. Thanks for bringing this up!

PR #159, opened 2025-06-10
Instruct LLM to add limit as an argument instead of SQL

DevonFulcher:
Unfortunately, the LLM isn't following instructions to use `--limit` instead of adding a limit to the SQL query. Here is an excerpt from my tests with the client:
```
user > Add a limit
[06/10/25 12:15:27] INFO     HTTP Request: POST https://api.openai.com/v1/responses "HTTP/1.1 200 OK"                                        _client.py:1025
Calling tool: show with arguments: {"sql_query":"select * from orders LIMIT 5"}
```
Even so, I think we can commit this change to try encouraging the LLM to use `--limit`. We may seek to encorporate Fusion in the future to determine if the LLM adds limit to the SQL query .

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -1,5 +1,4 @@
 import subprocess
-from typing import Optional
 
 from mcp.server.fastmcp import FastMCP
 from pydantic import Field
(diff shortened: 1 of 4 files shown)


DevonFulcher:
Addresses this feedback: https://getdbt.slack.com/archives/C08NGJYD563/p1749040282450049

reneruck-dbt (approved):

Issue #162, opened 2025-06-12
DBT Show does not allow you to use the LIMIT keyword


afoong-8s:
When the AI model generates a SQL query for the show tool e.g. 
`{
  "sql_query": "SELECT abc, count(*) as row_count\nFROM {{ ref('XXX') }}\nWHERE date >= current_date - 3\nGROUP BY abc\nORDER BY abc LIMIT 10"
}`

It returns an error:
Invalid type for parameter 'limit' in tool show

DevonFulcher:
This occurs because dbt show often adds a limit to the query in the background. If the SQL query also has a limit, then the query that is sent to the data platform has two limits, which is invalid. That PR should address this issue. Let us know if you see more issues related to this.

PR #163, opened 2025-06-12
Fix use of limit in dbt show

DevonFulcher:
Unfortunately, the LLM isn't listening to the instructions I provided [here](https://github.com/dbt-labs/dbt-mcp/pull/159). This approach is a bit hacky, but should provide a better experience most of the time.


DevonFulcher:
Addresses: https://github.com/dbt-labs/dbt-mcp/issues/162

PR #164, opened 2025-06-12
Add `--target` support to tools

hanslemm:
Added `--target` support to the tools.

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -8,13 +8,17 @@ from dbt_mcp.prompts.prompts import get_prompt
 
 
 def register_dbt_cli_tools(dbt_mcp: FastMCP, config: DbtCliConfig) -> None:
-    def _run_dbt_command(command: list[str], selector: str | None = None) -> str:
+    def _run_dbt_command(command: list[str], selector: str | None = None, target: str | None = None) -> str:
         # Commands that should always be quiet to reduce output verbosity
         verbose_commands = ["build", "compile", "docs", "parse", "run", "test"]
 
         if selector:
             selector_params = str(selector).split(" ")
             command = command + ["--select"] + selector_params
+
+        if target:
+            target_params = str(target).split(" ")
+            command = command + ["--target"] + target_params
 
         full_command = command.copy()
         # Add --quiet flag to specific commands to reduce context window usage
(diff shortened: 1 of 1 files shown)


uv.lock
DevonFulcher, line 1: uv.lock shouldn't need to be regenerated. Do you mind deleting this?

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -41,54 +71,143 @@ def build(
...
+
+                    default_target = profile_config.get(
+                        "target", list(targets.keys())[0] if targets else None
+                    )
+                    result["profiles"][profile_name] = {
+                        "targets": targets,
+                        "default_target": default_target,
+                    }
+
+            import json
```
DevonFulcher: I think it would be better to import this at the top of the file.

src/dbt_mcp/prompts/dbt_cli/get_profiles.md
```diff
@@ -0,0 +1,16 @@
...
+For each target, the tool returns:
+- Target name
+- Database adapter type (e.g., postgres, snowflake, bigquery, redshift)
+- Database name/catalog
+
+The profiles.yml file is typically located in:
+
+- The project directory (as specified by DBT_PROJECT_DIR)
+- The user's home directory at ~/.dbt/profiles.yml
+- Or as specified by the DBT_PROFILES_DIR environment variable
```
DevonFulcher: I wonder if we need this part in the prompt.

src/dbt_mcp/prompts/dbt_cli/args/target.md
```diff
@@ -0,0 +1,3 @@
+A target is an argument to dbt commands that allows using different data warehouse connection configurations for the same dbt project, such as different users, hosts, or even completely different databases.
```
DevonFulcher: I think it would be best to include how the target can be found. For instance, `Use the get_profiles tool to find available profiles.`.

docs/diagram.d2
```diff
@@ -26,6 +26,7 @@ tools: Tools {
     build
     compile
     docs
+    get_profiles
```
DevonFulcher, line 29: @b-per should be able to help get the updated diagram uploaded so it appears in the README and pypi.

DevonFulcher (commented):
Looks great! I just had a few minor comments.

hanslemm:
Thanks @DevonFulcher for the feedback. 

I've addressed the comments, please check latest changes.

b-per:
Hi @hanslemm !

I have been testing the PR, and it technically works 🚀 , but I am wondering what exactly you are trying to achieve/solve with it (and/or if we should slightly update the tool).

With this PR, we can
- get the different profiles
- for each profile get the targets
- specify a target for the current project

What I think is missing though is that today we can't know what is the profile used for the current dbt project. So in my case, the LLM tells me that I have 10+ profiles and several targets for each, but most of those targets won't work for my current profile.

I can think of 2 ways to solve this:
- update `get_profiles` to `get_profiles_and_targets` and allow the LLM to scope down the targets to the current profile only
- or create a new tool called `get_targets` that would get the target for the current profile only

What do you think?

hanslemm:
Good point, since I only have one profile in my case, I didn't think about multiple profiles + targets.

The main point of this PR is allowing the MCP to run on different targets. It is nice to have the chance to tell the LLM which target it should use when reaching the MCP to run the dbt related commands.

b-per:
Thanks!

In that case, is it OK to rename the tool to `get_targets`, as this reflects more what you are after, and update the logic it so that it gets the targets of the current profile?

And if it sounds good to you. Let me know if you want to take care of it, otherwise, I can give it a go.

Issue #180, opened 2025-06-18
SL metrics are not orderable


WilliamDee:
## Context

There's an issue when trying to order by metrics, this is due to https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/semantic_layer/client.py#L199 being pinned to a `OrderByGroupBy` object, which means whenever ordering by a Metric it fails. We need to figure out how to pass in a `OrderByMetric` when necessary so that it works properly.

PR #181, opened 2025-06-24
Fix order by to allow for ordering by metrics

WilliamDee:
## Context
Issue: #180. Anytime anyone wants to order by a metric, it fails as we hard pinned the parameter being passed to the SDK as a `OrderByGroupBy` which won't work for metrics. Adding support by parsing out the queried metrics/group-by and properly parsing it into the correct order by object.


DevonFulcher (approved):
🚀 Looks great!

PR #186, opened 2025-07-01
Fix issue with relative path for some dbt CLI

### LINKED ISSUE #176: DBT_PROJECT_DIR conflicts with dbt -- Issue when setting it to a relative path
Setting the `DBT_PROJECT_DIR` makes `dbt-mcp` to change directory (`cd $DBT_PROJECT_DIR`).

But `dbt` also interprets `DBT_PROJECT_DIR` value, and then tries to look for files in _that_ directory again.

E.g. setting `DBT_PROJECT_DIR` to `foo` will result in:

```
"Usage: dbt list [OPTIONS]
Try 'dbt list -h' for help.

Error: Invalid value for '--project-dir': Path 'foo' does not exist.
```

Because dbt CLI actually tries to access `foo/foo`.

I question the need for MCP to take over `DBT_PROJECT_DIR` where dbt CLI already natively supports this option.

b-per:
Closes #176 

Not the cleanest fix, but today there is a different behavior between dbt core/Fusion and the dbt Cloud CLI in relation to the env var `DBT_PROJECT_DIR`.

If they are all aligned one day we could just remove the `cwd` part.

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 
 from mcp.server.fastmcp import FastMCP
@@ -30,9 +31,14 @@ def register_dbt_cli_tools(dbt_mcp: FastMCP, config: DbtCliConfig) -> None:
         # Make the format json to make it easier to parse for the LLM
         full_command = full_command + ["--log-format", "json"]
 
+        # We change the path only if this is an absolute path, otherwise we can have
+        # problems with relative paths applied multiple times as DBT_PROJECT_DIR
+        # is applied to dbt Core and Fusion as well (but not the dbt Cloud CLI)
+        cwd_path = config.project_dir if os.path.isabs(config.project_dir) else None
+
         process = subprocess.Popen(
             args=[config.dbt_path, *full_command],
-            cwd=config.project_dir,
+            cwd=cwd_path,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             text=True,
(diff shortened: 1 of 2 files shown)


reneruck-dbt (approved):

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -30,9 +31,14 @@ def _run_dbt_command(
         # Make the format json to make it easier to parse for the LLM
         full_command = full_command + ["--log-format", "json"]
 
+        # We change the path only if this is an absolute path, otherwise we can have
+        # problems with relative paths applied multiple times as DBT_PROJECT_DIR
+        # is applied to dbt Core and Fusion as well (but not the dbt Cloud CLI)
+        cwd_path = config.project_dir if os.path.isabs(config.project_dir) else None
```
DevonFulcher, line 37: I'm not sure how absolute vs relative paths change the behavior here, but it seems like moltar is on to something here https://github.com/dbt-labs/dbt-mcp/issues/176#issuecomment-3023523270. Is the solution just to allow `DBT_PROJECT_DIR` to be None and for the user to just not set it in the .env file? I don't think I was aware that `DBT_PROJECT_DIR` was already used by dbt when I added that as an env var.

b-per: I don't really agree with the comment you linked as it wouldn't work for the Cloud CLI (which the person doesn't use but that we want to support with the MCP server).

Relative vs absolute is because, 

if `DBT_PROJECT_DIR='foo'` ; then, without the change we 
- cd to the subfolder `foo`
- because `DBT_PROJECT_DIR='foo'`, then we look for the subfolder `foo` if we run dbt Core or Fusion
- so we are looking in `foo/foo` ❌

if `DBT_PROJECT_DIR='/abs/path/to/foo'` ; then, we 
- cd to `/abs/path/to/foo`
- because `DBT_PROJECT_DIR='/abs/path/to/foo'`, then we look for the project in `/abs/path/to/foo`, which is where we already are

DevonFulcher: Got it! Ya, that makes sense.

DevonFulcher (approved):
Awesome!

PR #188, opened 2025-07-01
Return errors when the dbt command didn't select any node

### LINKED ISSUE #179: Run swallows all feedback
```
⏺ dbt:run (MCP)(selector: "foo")
```

Executes:

```sh
dbt run --quiet --select foo --log-format json
```

The `--quiet` then produces no output.

But the output contains:

> The selection criterion 'foo' does not match any enabled nodes

Yet the feedback from the MCP client was:

> ⏺ The Foo model has been successfully executed in dbt.

I think `--quiet` and `--log-format json` are also contradictory in it's intent.

b-per:
Closes #179 

Working in Claude Code
image


README.md
```diff
@@ -39,10 +39,11 @@ The MCP server takes the following environment variable configuration:
...
+| ------------------------ | ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `DBT_PROJECT_DIR`        | -                                              | The path to where the repository of your dbt Project is hosted locally. This should look something like `/Users/firstnamelastname/reponame`                                                               |
+| `DBT_PATH`               | -                                              | The path to your dbt Core, dbt Cloud CLI, or dbt Fusion executable. You can find your dbt executable by running `which dbt`                                                                               |
+| `DBT_WARN_ERROR_OPTIONS` | `'{"error": ["NoNodesForSelectionCriteria"]}'` | The dbt warnings that should return errors. When using LLMs we likely want to return an error if the dbt command didn't select any mode [docs](https://docs.getdbt.com/reference/global-configs/warnings) |
```
reneruck-dbt:
```suggestion
| `DBT_WARN_ERROR_OPTIONS` | `'{"error": ["NoNodesForSelectionCriteria"]}'` | The dbt warnings that should return errors. When using LLMs we likely want to return an error if the dbt command didn't select any node [docs](https://docs.getdbt.com/reference/global-configs/warnings) |
```

README.md
```diff
@@ -39,10 +39,11 @@ The MCP server takes the following environment variable configuration:
...
+| ------------------------ | ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `DBT_PROJECT_DIR`        | -                                              | The path to where the repository of your dbt Project is hosted locally. This should look something like `/Users/firstnamelastname/reponame`                                                               |
+| `DBT_PATH`               | -                                              | The path to your dbt Core, dbt Cloud CLI, or dbt Fusion executable. You can find your dbt executable by running `which dbt`                                                                               |
+| `DBT_WARN_ERROR_OPTIONS` | `'{"error": ["NoNodesForSelectionCriteria"]}'` | The dbt warnings that should return errors. When using LLMs we likely want to return an error if the dbt command didn't select any node [docs](https://docs.getdbt.com/reference/global-configs/warnings) |
```
DevonFulcher: I think we should add this to `.env.example`, right?

src/dbt_mcp/config/config.py
```diff
@@ -74,6 +74,11 @@ def load_config() -> Config:
     disable_remote = os.environ.get("DISABLE_REMOTE", "true") == "true"
     multicell_account_prefix = os.environ.get("MULTICELL_ACCOUNT_PREFIX", None)
 
+    # set default warn error options if not provided
+    if not os.environ.get("DBT_WARN_ERROR_OPTIONS"):
+        warn_error_options = '{"error": ["NoNodesForSelectionCriteria"]}'
+        os.environ["DBT_WARN_ERROR_OPTIONS"] = warn_error_options
```
DevonFulcher, line 80: I wonder if we want to edit the env at the call to `subprocess.Popen` instead of here. This should work fine anyway though!

DevonFulcher (approved):
I'm hesitant to add more user config burden in the README. I wonder if this is just something we do by default and the user can override it if they really know what they are doing. Either way, this is a great change, thank you!

README.md
b-per: I think that it relates to your other commend about adding stuff to the README.

In the end, this env var is not specific to the MCP server, it is an out of the box dbt one ; but it makes sense that in a MCP context we set the value that I listed. I'd expect the vast majority of people won't need to change it and will be OK with this default.

Instead of adding a whole new row and adding it to the example env var file, we could maybe just add a comment in the README saying that:
- all dbt CLI env vars are supported
- by default we set `DBT_WARN_ERROR_OPTIONS` to the value mentioned as this provides a better experience

DevonFulcher: Sounds great!

src/dbt_mcp/config/config.py
b-per: We can change it after if we want. It works as-is.

Issue #178, opened 2025-06-18
Error: MCP tool "list" response exceeds maximum allowed tokens


moltar:
Our project is not even medium-sized; it's on the small side.

Yet the list continues to fail, even when it attempts to recover and apply filters.

```
⏺ dbt:list (MCP)(selector: "")
  ⎿  Error: MCP tool "list" response (171718 tokens) exceeds maximum allowed tokens (25000). Please use pagination, filtering, or limit parameters to reduce the response size.
```

I think it's perhaps because of JSON output?

In any case, at this point the tool is unusable.

b-per:
Hi @moltar . Thanks for raising this.

What MCP client are you using? Is it Claude Code? I searched a bit online and it seems that users of Claude Code report having this issue sometimes with different tools.

In your case, what is the ballpark size of your project (models + tests)?

We could trim the JSON to the relevant fields, but it looks like for your project to work we'd have to divide the size of the JSON  by at least seven.

moltar:
Yes, it's CC.

Models: 85
Tests: ~ 700 (across all models)

DevonFulcher:
Hey @moltar, https://github.com/dbt-labs/dbt-mcp/pull/189 should improve this issue. Please let us know if this doesn't work for you or if you have any other comments. I will close this issue for now.

Issue #175, opened 2025-06-17
dbt:list is not working correctly


moltar:
Instead of printing a list of models, it just printed help, so claude aborted and went looking for files instrad.

```
⏺ I'll help you find all dbt models that start with "gold". Let me search for these models in your dbt project.

⏺ dbt:list (MCP)(selector: "gold*")
  ⎿  Usage: dbt list [OPTIONS]   
     Try 'dbt list -h' for help.

     … +1 line (ctrl+r to expand)
```

DevonFulcher:
PR #189 should fix this issue. Please let us know if you have any more comments on this. I'll close this issue for now.

Issue #165, opened 2025-06-12
dbt-list can trigger endless loop of summarization of context in GitHub Copilot


hanslemm:
## Problem

In very large dbt project I noticed that the JSON content of `dbt list` might be too big, specially while using VSCode GitHub Copilot with Claude Sonnet 4 in agent mode. This command would trigger a summarization of the history (because it is too much information), and it would lose the context that it just fetched, stuck in an ever ending loop of `dbt list` retrievals, like:
- call `dbt-list` 
- too much content 
- summarize chat history (lose the result of the call) 
- call `dbt-list` (restarting the loop)

Therefore, it appears that the retrieved JSON should be more concise before passing on to the LLM (at least inside GitHub Copilot context), specially for large projects.

## Possible Solutions

Perhaps the parser should consider a limit of content retrieval, and after that limit, it should summarize before returning the result to the llm, perhaps based on number of models.

DevonFulcher:
Hey @hanslemm, https://github.com/dbt-labs/dbt-mcp/pull/189 should improve this issue. Please let us know if this doesn't work for you or if you have any other comments. I will close this issue for now.

PR #189, opened 2025-07-01
Decrease output size for list and improve node filtering

b-per:
Improvements related to #178 and #175 and maybe #165 as well

A few things to decrease the amount of data that `list` returns
- only select specific resource types (excluding tests by default should help a lot)
- allow searching nodes by parts of their name
- returning text and not JSON (we lose some info but it is not that useful)
- run with `--quiet` (even with `--quiet` we still list objects)

### DIFF
--- a/src/dbt_mcp/prompts/dbt_cli/args/resource_type.md
+++ b/src/dbt_mcp/prompts/dbt_cli/args/resource_type.md
@@ -1,0 +1,16 @@
+It is possible to list specific resource types we want to return.
+
+We can pick any from the following and need to return a list of all the resource types to select:
+- metric
+- semantic_model
+- saved_query
+- source
+- analysis
+- model
+- test
+- unit_test
+- exposure
+- snapshot
+- seed
+- default
+- all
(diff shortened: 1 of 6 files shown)


src/dbt_mcp/prompts/dbt_cli/args/resource_type.md
```diff
@@ -0,0 +1,16 @@
+It is possible to list specific resource types we want to return.
+
+We can pick any from the following and need to return a list of all the resource types to select:
```
DevonFulcher, line 3: Does this need to be a comma-separated list of resource types? An example may help.

b-per: Thinking about it, that is likely something that might be slightly different between dbt Core, Fusion and the dbt Cloud CLI ( I haven't tested it) as it relates to args parsing and we use different libraries for the different CLIs.

For dbt-core it expects the list space separated: `dbt list --resource-types model source`

But I think that especially because it might be slightly different for each CLI, we should take care of the formatting in our logic and not in the LLM answer we get.

DevonFulcher: Sounds good. In that case, we would need a way for the user to indicate which CLI they are using. For now, we are treating each as the same interface.

PR #194, opened 2025-07-02
Move Discover headers to config

DevonFulcher:
This should allow for greater flexibility in the headers passed to the Discovery API. This is particularly useful for our internal usage.

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -27,8 +27,8 @@ class SemanticLayerConfig:
 @dataclass
 class DiscoveryConfig:
     url: str
+    headers: dict[str, str]
     environment_id: int
-    token: str
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


src/dbt_mcp/config/config.py
```diff
@@ -169,8 +169,11 @@ def load_config() -> Config:
             url = f"https://metadata.{actual_host}/graphql"
         discovery_config = DiscoveryConfig(
             url=url,
+            headers={
+                "Authorization": f"Bearer {token}",
```
alan-andrade, line 173: I could be wrong here, but I think we might need to let the header name be passed in as well. Not too familiar with other use cases, but from the agent use case, we may have to pass in a different header name and value for authentication because we'll have a jwt already.  Tokens go on this authorization header, jwt go in the internal auth header.

PR #200, opened 2025-07-03
Pass empty list instead of None for order_by

### LINKED ISSUE #196: "'NoneType' object is not iterable" returned from query_metrics
When using the dbt-mcp server, I keep encountering this error when calling query_metrics:

`'NoneType' object is not iterable`

An example query is below (from GitHub Copilot, but the client doesn't matter)

```
{
  "metrics": [
    "time_in_hours"
  ],
  "group_by": [
    {
      "name": "user__full_name",
      "type": "dimension",
      "grain": null
    },
    {
      "name": "user__sub_department",
      "type": "dimension",
      "grain": null
    }
  ],
  "where": "{{ Dimension('user__sub_department') }} = 'Digital Solutions'",
  "limit": 100
}
```

If I run this same query from the dbt Cloud IDE I get the results that I expect. 

`dbt sl query --metrics time_in_hours --group-by user__full_name,user__sub_department --where "{{ Dimension('user__sub_department') }} = 'Digital Solutions'"`

The query is never compiled to SQL and sent to Snowflake (as it never appears in the query history for the warehouse). So it seems to be an issue in compiling the query. Debugging this project, I can see that the error is returned from this line:

https://github.com/dbt-labs/dbt-mcp/blob/dd6820c236478f20e95c555e662cbbdc7fdf04a8/src/dbt_mcp/semantic_layer/client.py#L242

Some queries return this error, others don't. I haven't found a pattern yet. I'd appreciate any help in debugging this.

WilliamDee:
## Context
Seems like `None` is not possible as it expects a list for `order_by` if passed in the SDK.

closes #196

### DIFF
--- a/src/dbt_mcp/semantic_layer/client.py
+++ b/src/dbt_mcp/semantic_layer/client.py
@@ -223,12 +223,12 @@ class SemanticLayerFetcher:
                 # Catching any exception within the session
                 # to ensure it is closed properly
                 try:
-                    parsed_order_by: list[OrderBySpec] | None = (
+                    parsed_order_by: list[OrderBySpec] = (
                         self.get_order_bys(
                             order_by=order_by, metrics=metrics, group_by=group_by
                         )
                         if order_by is not None
-                        else None
+                        else []
                     )
                     query_result = self.sl_client.query(
                         metrics=metrics,


DevonFulcher (dismissed):
🚀

DevonFulcher (approved):

Issue #203, opened 2025-07-09
Missing exception/error message in latest MCP


b-per:
In the latest MCP server version, when there are issues with tools we get a generic error:

<img width="823" height="652" alt="Image" src="https://github.com/user-attachments/assets/f0983a33-1ad4-4b43-8bba-8c2cb4b864ba" />

If we modify the tool to return a default value in case of exception, then we get a better error (like we used to have in previous versions)

```
    @dbt_mcp.tool(description=get_prompt("semantic_layer/list_metrics"))
    def list_metrics() -> list[MetricToolResponse] | str:
        try:
            return semantic_layer_fetcher.list_metrics()
        except Exception as e:
            return f"Error listing metrics: {e}"
```

<img width="798" height="415" alt="Image" src="https://github.com/user-attachments/assets/09615992-d6fb-41b9-9086-7846f010dc06" />

PR #204, opened 2025-07-09
Catch every tool error and surface as string

DevonFulcher:
This PR catches errors from tool calls and surfaces them as strings. This should fix https://github.com/dbt-labs/dbt-mcp/issues/203. I was able to reproduce the error and confirm that this solution fixes the problem. It is unfortunate that this is necessary, but it seems Claude specifically is having a hard time handling thrown exceptions with the latest version of the MCP SDK.


DevonFulcher:
I recommend hiding whitespace for this review:
image

b-per (approved):

Issue #182, opened 2025-06-25
Disable specific tools via configuration


DevonFulcher:
Certain users require specific controls over which tools are enabled. This would look like an env var like `DISABLE_TOOLS=get_dimensions,get_all_models,compile`

PR #220, opened 2025-07-18
Added ability to disable certain tools when registering the MCP server

WilliamDee:
## Context
Allow the flexibility to specify certain tools to disable instead of needing to register all tools. This is useful for when certain tools may not be needed.

### DIFF
--- a/.changes/unreleased/Under the Hood-20250717-213422.yaml
+++ b/.changes/unreleased/Under the Hood-20250717-213422.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Adding the ability to exclude certain tools when registering
+time: 2025-07-17T21:34:22.971538-04:00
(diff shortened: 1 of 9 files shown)


WilliamDee:
Maybe there's a better way to do this dynamically. It seems like Claude has the ability to dynamically disable certain tools 
Screenshot 2025-07-17 at 9 35 42 PM
I wonder what they do to achieve that 🤔

DevonFulcher (approved):
Nice! This PR is 99% the way to completing https://github.com/dbt-labs/dbt-mcp/issues/182. Do you mind adding an env var to make this configurable? Some clients, like Claude Desktop allow you to configure which tools are enabled, but this is client-side and isn't available for all clients, so this is valuable. Thank you!

DevonFulcher (approved):
Amazing!

PR #224, opened 2025-07-18
Create list of tool names

DevonFulcher:
This PR provides a list of tool names. This is helpful when considering which tools to enable, both internally & externally. I tested that the `test_tool_names_match_server_tools` errors properly when the enum isn't in sync. I also (accidentally) tested a nice error message when the `DISABLE_TOOLS` enum isn't valid:
```
ValueError: 'get_metrics' is not a valid ToolName
```
Finally, I ran `task client` and saw that the list of available tools has been filtered:
```
Available tools:
- get_dimensions(metrics)
- get_entities(metrics)
- query_metrics(metrics, group_by, order_by, where, limit)
- get_all_models()
- get_model_details(model_name, unique_id)
- get_model_parents(model_name, unique_id)
- get_model_children(model_name, unique_id)
- build(selector)
- compile()
- docs()
- list(selector, resource_type)
- parse()
- run(selector)
- test(selector)
- show(sql_query, limit)
- text_to_sql(text)
```


b-per:
This actually relates to a comment I left in #230

DevonFulcher:
> This actually relates to a comment I left in #230

Let's get your PR merged in first. Then, I can fix up this PR to be compatible.

src/dbt_mcp/config/config.py
```diff
@@ -87,8 +89,11 @@ def actual_prod_environment_id(self) -> int | None:
 
     @field_validator("disable_tools", mode="before")
     @classmethod
-    def parse_disable_tools(cls, value: str) -> list[str]:
-        return [tool.strip() for tool in value.split(",") if tool.strip()]
+    def parse_disable_tools(cls, env_var: str | list) -> list[ToolName]:
```
b-per, line 92: Why do we have a `str | list` as input here? As we have set `NoDecode` for `disable_tools` it will always be considered a `str` (or None maybe).

Because of `NoDecode` we can then set our own parsing

src/dbt_mcp/config/config.py
```diff
@@ -87,8 +89,11 @@ def actual_prod_environment_id(self) -> int | None:
...
+        if not env_var:
+            return []
+        assert isinstance(env_var, str)
+        return [ToolName(tool.strip()) for tool in env_var.split(",") if tool.strip()]
```
b-per, line 96: Isn't the goal of having a tool list being able to raise errors if people provide incorrect tools?

DevonFulcher: `ToolName(tool.strip()` does raise an error if the name isn't valid:
```
DISABLE_TOOLS
  Value error, 'foo' is not a valid ToolName [type=value_error, input_value='get_mart_models,list_metrics,execute_sql,foo', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
```

src/dbt_mcp/config/config.py
DevonFulcher: In my testing, if `DISABLE_TOOLS` isn't provided as an env var, then it is an empty list here rather than an empty string, as you would expect.

b-per: Thanks! There is a bug in my release then (i'll think of adding more tests about combinations of env vars)

In that case, could you fix the issue in your PR? I think that a solution a little bit cleaner would be having the value defaulting to None (instead of an empty string)
```
disable_tools: Annotated[list[str], NoDecode] = Field(None, alias="DISABLE_TOOLS")
```

and then having the type `str | None`
```
      def parse_disable_tools(cls, env_var: str | None) -> list[ToolName]:
        if value is None:
            return []
...
```

src/dbt_mcp/config/config.py
b-per: OK. If `DISABLE_TOOLS=foo,bar` though ; would this implementation just error on `foo` or on `bar` as well? Ideally we'd want to get the error for both at once.

DevonFulcher: Resolving for now so I can merge this. I will fix this in a follow up.

src/dbt_mcp/config/config.py
DevonFulcher: Resolving for now so I can merge this. I will fix this in a follow up.

PR #226, opened 2025-07-18
Add exclude_tools parameter to `register_dbt_cli_tools` function

jairus-m:
Closes #225 - Fixes `TypeError` when running `uvx dbt-mcp`.

Issue #225, opened 2025-07-18
`register_dbt_cli_tools` leading to a `TypeError` when running `uvx dbt-mcp` in latest `0.2.16`


jairus-m:
# Summary
I currently use the `dbt-mcp` server with Claude Code CLI (`1.0.53`) in a dev container. The startup command I run to get this working is `claude mcp add dbt-mcp -- uvx dbt-mcp`. As of, July 18, 2025, Claude Code CLI is not connecting to the dbt-mcp server:

# Details

Claude MCP status after running `/mcp` using the latest version:
<img width="589" height="163" alt="Image" src="https://github.com/user-attachments/assets/536a1c15-47de-4dd8-9d8b-4243571bc4af" />

However if I run `claude mcp add dbt-mcp==0.2.15 -- uvx dbt-mcp` and pin to a previous version, Claude Code CLI is able to properly connect to the MCP server. 
<img width="886" height="197" alt="Image" src="https://github.com/user-attachments/assets/ef0db633-3bc0-4ef0-8a22-f64988fdf6bb" />

It looks like this is related to the latest release [v0.2.16](https://github.com/dbt-labs/dbt-mcp/releases/tag/v0.2.16).

Here is the error when running `uvx dbt-mcp` (latest version):

```bash
[07/18/25 21:58:29] INFO     Registering semantic layer tools                                                                                                                                      server.py:93
                    INFO     Registering discovery tools                                                                                                                                           server.py:97
                    INFO     Registering dbt cli tools                                                                                                                                            server.py:101
Traceback (most recent call last):
  File "/root/.cache/uv/archive-v0/-uj7-4ImjCopD8fSMhRlM/bin/dbt-mcp", line 6, in <module>
    from dbt_mcp.main import main
  File "/root/.cache/uv/archive-v0/-uj7-4ImjCopD8fSMhRlM/lib/python3.12/site-packages/dbt_mcp/main.py", line 10, in <module>
    main()
  File "/root/.cache/uv/archive-v0/-uj7-4ImjCopD8fSMhRlM/lib/python3.12/site-packages/dbt_mcp/main.py", line 7, in main
    asyncio.run(create_dbt_mcp()).run()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/python/current/lib/python3.12/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
[...]

Issue #198, opened 2025-07-03
Add Tool Annotations


DevonFulcher:
MCP has a new feature for tool annotations: https://modelcontextprotocol.io/docs/concepts/tools#tool-annotations. It would be a good practice to add these to all tools.

DevonFulcher:
Implemented in #256

PR #229, opened 2025-07-20
Add integration test for server initialization

jairus-m:
# Summary
Add integration test for MCP server initialization to prevent regressions (like #226).. creates a basic check for initialization errors in the future. The goal here is to implement a low-cost but high-impact test!

# Details
- Added a module `tests/integration/initialization/test_initialization.py` with a single test
- Tests MCP server initialization by running `create_dbt_mcp` using the baseline `mock_config` from `tests/mocks/config.py`
  - Basic smoke test in asserts since any errors should be caught before
  - Also had to add a few `__init__.py` files to allow for Pytest to recognize the `tests/` module when importing `mock_config`
- Added decorator to filter noise from mulitple deprecation warnings from `dbts` when importing `create_dbt_mcp`

In general, this tests for any exceptions/errors that may cause the `dbt-mcp` sever to fail on startup. This doesn't cover runtime/operational or post-initializtion errors. Happy to get feedback, add to it, and/or make changes! Thank u 😄

### Testing
```bash
# Full integration suite
task test:integration

# Just this bad boy
pytest tests/integration/initialization/test_initialization.py
```

I also ran this test with the code from `dbt-mcp==0.2.16` (https://github.com/dbt-labs/dbt-mcp/issues/225) and this is how `test_initialization` catches the regression:

Screenshot 2025-07-20 at 7 36 23 AM


tests/integration/initialization/test_initialization.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+import asyncio
+from unittest.mock import patch
+
+from tests.mocks.config import mock_config
+
+
+@pytest.mark.filterwarnings("ignore::DeprecationWarning:dbtsl.*")
+def test_initialization():
+    from dbt_mcp.mcp.server import create_dbt_mcp
```
DevonFulcher: Can we move this line to the top of the file?

jairus-m: @DevonFulcher 

I absolutely can! However for my own learning, I did want to share and ask:

I put the import statement within the test itself in order to filter the 12 warnings that come up from deprecations when  the`dbtsl` module gets ran on import:

Screenshot 2025-07-21 at 7 35 56 AM

With that additional context, is it still fine to keep the import there? or is it best practice to maintain warnings in the test (and also keep the import at the top of the file)? 

..and then filter warnings explicitly through the CLI rather than by default ?
```python
pytest tests/integration/initialization/test_initialization.py -W ignore
```

DevonFulcher: I think we should leave the warnings present and not filter or mute them. I think they are real problems that we should address in the future. Filtering them just hides the problem.

jairus-m: Totally makes sense - will update

DevonFulcher (approved):
Looks great, thanks!

DevonFulcher (commented):
Oh, we should add a changelog this time!

jairus-m:
> Oh, we should add a changelog this time!

@DevonFulcher 
I turned off pre-commits yesterday but will run `ruff format` as well ! And then just to confirm, I do bump the changelog version to 0.2.18 ?  And will changie have any negative affects/issues if I update the v0.2.17 changelog description?:

```
## v0.2.18 - 2025-07-21
### Under the hood
* Add integration test for server initialization

## v0.2.17 - 2025-07-18
### Bug Fix
* Correct missing arguments in `register_dbt_cli_tools` causing a TypeError on startup
```

DevonFulcher:
@jairus-m You shouldn't update the changelog manually. You will need to install `changie` and run `changie new`. There are some instructions here: https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#changelog

DevonFulcher (approved):
Nice!

jairus-m:
@DevonFulcher Thank you for pointing me to the changie docs lol - never heard of it but looks like a neat utility!

DevonFulcher:
Hey @jairus-m, the PR looks good. I am going to merge it, but in the future, we will need signed commit for every commit in the PR. This is a policy that our security team has set up. I will bypass it this time, but if you plan to commit to dbt-mcp in the future, you will need to follow this guidance to set up signed commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits. Honestly, it is a good practice, even if you don't plan to contribute to dbt-mcp in the future.

jairus-m:
@DevonFulcher Appreciate it - my bad on that. Just started reading up to get it right. I'm currently looking at taking a stab at  #198 in the near future so will def set this up!

PR #230, opened 2025-07-21
Move env var loading to `pydantic_settings`

b-per:
Replace `dotenv` by `pydantic_settings` to get better config parsing out of the box. This should give us "free" parsing validation for our different types for our env vars

This feels relevant knowing that we have more and more settings being added

There are slight changes with the previous behavior though
- we used to check `true` for some values, I believe that is now truthy (e.g. 1 would work as well)
  - I can double check it
- `disable_tools` is a `list[str]`
  - I need to check what format it would accept. There is also logic to separate a string with `,`
  - I wanted to add some validation on that field that all tools are "valid" ones to make sure people don't add typos, but I couldn't manage to get the list of tools dynamically and didn't want us to have to define a tool and also hard code it in the validation


b-per:
With the latest commit, I have tested that the following works as expected:

```
DISABLE_TOOLS=list_metrics
or 
DISABLE_TOOLS=list_metrics,
or
DISABLE_TOOLS=list_metrics,get_dimensions
```


I also tested the "truthy" behavior change and it is following my assumption. Setting an env var for a bool param to 1 is matching it to true. I think that this is a good change and I don't expect that it will break people's setup.

uv.lock
DevonFulcher, line 1: This change indicates that we are on different versions of uv. I'm on version 0.6.16. This is the same as the `.tool-versions` file. Do you mind installing this version or upgrading the version in `.tool-versions`?

src/dbt_mcp/config/config.py
```diff
@@ -1,214 +1,237 @@
...
+    discovery_config: DiscoveryConfig | None = None
+    semantic_layer_config: SemanticLayerConfig | None = None
     disable_tools: list[str]
 
 
 def load_config() -> Config:
-    load_dotenv()
-
-    host = os.environ.get("DBT_HOST")
-    cursor_host = os.environ.get("DBT_MCP_HOST")
```
DevonFulcher, line 66: Did you mean to delete this one?

DevonFulcher: Oh I see you renamed it.

DevonFulcher (approved):
Looks great!

uv.lock
b-per: I updated the version in the file. With `uv` being so actively developed we should aim to get as close to the latest as possible.

PR #235, opened 2025-07-22
Improve DISABLE_TOOLS configuration

DevonFulcher:
This is an improvement on recent PRs for providing agent name types and the ability to disable certain tools. I tested:
- Starting the server without the `DISABLE_TOOLS` env var defined
- Setting multiple invalid tools in `DISABLE_TOOLS`
- Setting only valid tools in `DISABLE_TOOLS`

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -96,6 +96,8 @@ class DbtMcpSettings(BaseSettings):
         tool_names: list[ToolName] = []
         for tool_name in env_var.split(","):
             tool_name_stripped = tool_name.strip()
+            if tool_name_stripped == "":
+                continue
             try:
                 tool_names.append(ToolName(tool_name_stripped))
             except ValueError:


src/dbt_mcp/config/config.py
```diff
@@ -89,11 +89,23 @@ def actual_prod_environment_id(self) -> int | None:
...
+    def parse_disable_tools(cls, env_var: str | None) -> list[ToolName]:
         if not env_var:
             return []
-        assert isinstance(env_var, str)
-        return [ToolName(tool.strip()) for tool in env_var.split(",") if tool.strip()]
+        errors: list[str] = []
+        tool_names: list[ToolName] = []
+        for tool_name in env_var.split(","):
+            tool_name_stripped = tool_name.strip()
+            try:
```
b-per: We could skip empty strings, no?

```suggestion
            tool_name_stripped = tool_name.strip()
            if tool_name_stripped == "":
                continue
            try:
```

b-per (approved):
Left a quick comment. You can merge without it if you don't like it :-)

src/dbt_mcp/config/config.py
DevonFulcher: Good call that would help in the case of a trailing comma.

PR #249, opened 2025-07-24
Winkler/metadata experiment

matt-winkler:
Would like to figure out the best way to include the trust signals from Discovery API in as shown in the below. My MCP-Fu is not that good, so looking for feedback / advice on whether these should be all wrapped in the model details, or if breaking them into separate tools would make more sense.

### DIFF
--- a/src/dbt_mcp/discovery/client.py
+++ b/src/dbt_mcp/discovery/client.py
@@ -61,6 +61,17 @@ class GraphQLQueries:
                                         type
                                     }
                                 }
+                                executionInfo {
+                                    lastRunGeneratedAt
+                                    lastRunStatus
+                                    executeCompletedAt
+                                    executeStartedAt
+                                    executeEndedAt
+                                    executeDuration
+                                    executeError
+                                    executeErrorType
+                                    executeErrorDetails
+                                }
                             }
                         }
                     }
--- a/src/dbt_mcp/discovery/client.py
+++ b/src/dbt_mcp/discovery/client.py
@@ -54,23 +54,11 @@ class GraphQLQueries:
                                 description
                                 database
                                 schema
-                                catalog {
-                                    columns {
-                                        description
-                                        name
-                                        type
-                                    }
-                                }
                                 executionInfo {
                                     lastRunGeneratedAt
                                     lastRunStatus
                                     executeCompletedAt
                                     executeStartedAt
-                                    executeEndedAt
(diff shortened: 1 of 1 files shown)


matt-winkler:
In case this is useful context of what I'm trying to do, here was my prompt:

```
In order to assess health of the data behind dbt semantic layer metrics, you should evaluate the use all three of these data points from get model details: 
* the model level executionInfo
* test level executionInfo 
* the source freshness of model parents. 
** If the freshnessStatus is null, consider the model questionably health.
** If the freshnessRunGeneratedAt is null or more than 24 hours old, consider the model questionably healthy.
** If the freshnessStatus is "pass", consider the model healthy.
** If the freshnessStatus is "fail", consider the model unhealthy.
** If the freshnessStatus is "warn", consider the model questionably healthy.

For example, if the lastRunStatus is "success" consider the model healthy. 
Check the health of the underlying models when responding to requests to query semantic layer metrics, and return this information to the user.

Check the health of the underlying models when responding to requests to query semantic layer metrics, and return this information to the user.
```

src/dbt_mcp/discovery/client.py
```diff
@@ -54,13 +54,35 @@ class GraphQLQueries:
                                 description
                                 database
                                 schema
-                                catalog {
-                                    columns {
-                                        description
-                                        name
-                                        type
```
DevonFulcher: I think we should leave the columns piece. I'm not sure why you were getting nulls, but we should root cause that before removing.

src/dbt_mcp/prompts/semantic_layer/list_metrics.md
```diff
@@ -6,4 +6,8 @@ that can be used with other tools to answer the question.
 
 Examples:
 - "What are the top 5 products by revenue?"
-- "How many users did we have last month?"
\ No newline at end of file
+- "How many users did we have last month?"
+
+Synonyms:
+- Pay special attention to any synonyms which are defined in the meta properties returned from metrics
+- Use these synonyms to map questions from the user to the appropriate metrics
```
DevonFulcher, line 13: I like the idea of having synonyms in `meta` be meaningful, but we haven't documented anywhere that users should include synonyms in their metrics. I'll ask the SL team what we should do here. I recommend moving the trust signals in the Discovery API section into a separate PR because we can get that merged right away.

PR #254, opened 2025-07-25
Expand tool descriptions. Support selector on show and compile.

greg-mckeon:
hey folks!

this is mostly vibe coded, so pls feel free to critique heavily. this PR does the following:
1. updates the tool descriptions for the dbt cli. this has helped Cline better call specific tools.
2. adds selectors to the show and compile commands. this has helped cline rely more on dbt models vs generating its own inline code.

I've updated tests and added a changie - lmk what else you need!

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -75,8 +75,50 @@ def register_dbt_cli_tools(dbt_mcp: FastMCP, config: DbtCliConfig) -> None:
         return _run_dbt_command(["build"], selector, is_selectable=True)
 
     @dbt_mcp.tool(description=get_prompt("dbt_cli/compile"))
-    def compile() -> str:
-        return _run_dbt_command(["compile"])
+    def compile( 
+        sql_query: str | None = Field(
+            default=None, description=get_prompt("dbt_cli/args/sql_query")
+        ),
+        selector: str | None = Field(
+            default=None, description=get_prompt("dbt_cli/args/selectors")
+        ),
+    ) -> str:
+        """
+        Render executable SQL.
+
+        **One** of ``sql_query`` or ``selector`` may be passed. If both are
+        provided, an error string is returned.
+
+        An empty ``selector`` compiles the entire project.
+
+        Parameters
+        ----------
+        sql_query:
+            Raw SQL to compile via ``--inline``.
+        selector:
+            dbt node selector to compile via ``--select``.
+        """
+        # Handle Pydantic Field objects that might be passed instead of None
+        if hasattr(sql_query, 'default'):
+            sql_query = None
+        if hasattr(selector, 'default'):
+            selector = None
+            
+        if sql_query is not None and selector is not None:
+            return (
+                "You cannot provide *both* `sql_query` *and* `selector` "
+                "when calling the `compile` tool"
+            )
+
+        args: list[str] = ["compile"]
+        if sql_query is not None:
+            args.extend(["--inline", sql_query])
+            return _run_dbt_command(args)
+        elif selector is not None:
+            return _run_dbt_command(args, selector, is_selectable=True)
+        else:
+            # Neither provided - compile entire project
+            return _run_dbt_command(args)
 
     @dbt_mcp.tool(description=get_prompt("dbt_cli/docs"))
     def docs() -> str:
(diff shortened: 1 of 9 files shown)


src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -75,8 +75,50 @@ def build(
         return _run_dbt_command(["build"], selector, is_selectable=True)
 
     @dbt_mcp.tool(description=get_prompt("dbt_cli/compile"))
-    def compile() -> str:
-        return _run_dbt_command(["compile"])
+    def compile( 
+        sql_query: str | None = Field(
```
DevonFulcher, line 79: We should reconcile this with @reneruck-dbt's changes in our internal service. We should only have one way of compiling a SQL string. I'm not sure if that means two separate tools or a single one that accepts various parameters as you have here.

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -75,8 +75,50 @@ def build(
...
+        sql_query:
+            Raw SQL to compile via ``--inline``.
+        selector:
+            dbt node selector to compile via ``--select``.
+        """
+        # Handle Pydantic Field objects that might be passed instead of None
+        if hasattr(sql_query, 'default'):
+            sql_query = None
+        if hasattr(selector, 'default'):
+            selector = None
```
DevonFulcher, line 105: Is this necessary or vibe coded? The type annotations are `str | None`. So, I would think this wouldn't be necessary. Is the LLM passing a `Field` object instead? That is unexpected.

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -75,8 +75,50 @@ def build(
...
+
+        args: list[str] = ["compile"]
+        if sql_query is not None:
+            args.extend(["--inline", sql_query])
+            return _run_dbt_command(args)
+        elif selector is not None:
+            return _run_dbt_command(args, selector, is_selectable=True)
+        else:
+            # Neither provided - compile entire project
+            return _run_dbt_command(args)
```
DevonFulcher, line 121: I think you would still want `is_selectable=True` here. If the operation times out, the error message will suggest adding a selector.

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -122,25 +164,72 @@ def test(
...
+        # Inline SQL vs selector
+        if sql_query is not None:
+            args.extend(["--inline", sql_query])
+        else:  # selector is not None
+            args.extend(["--select", *str(selector).split(" ")])
+
+        # Handle limit logic
+        cli_limit: int | None = None
+        if sql_query and "limit" in sql_query.lower():
+            # If the user wrote their own LIMIT we tell dbt not to add another.
```
DevonFulcher, line 224: This comment isn't correct, exactly. I think: `If the LLM included LIMIT in the SQL we tell dbt not to add another.` is better.

src/dbt_mcp/prompts/dbt_cli/build.md
```diff
@@ -1,8 +1,74 @@
...
+
+Tasks are executed **in DAG order**, so dependencies are always honoured.
+
+---
+
+## Arguments  
+
+| Name | Required | Description |
+|------|----------|-------------|
+| `selector` | No (string) | Graph selector that limits what part of the DAG to build. Strongly recommended for large projects. See examples below for syntax. |
```
DevonFulcher, line 22: I think this is redundant of the prompt at `dbt_cli/args/selectors`, so we should delete it here and update that other prompt, if necessary.

src/dbt_mcp/prompts/dbt_cli/build.md
```diff
@@ -1,8 +1,74 @@
...
+
+---
+
+## Usage Guidelines  
+
+### ✅ DO
+* Use a **selector** whenever possible to avoid unnecessarily rebuilding the entire DAG.
+* Build **+model_name** when you want to regenerate a model _and_ its children (tests, snapshots, downstream marts, etc.).
+* Use `model_name+` sparingly; it can trigger an expensive full-DAG rebuild.
+* In CI pipelines, prefer `dbt build --select state:modified+` to build only items modified by the current commit.
```
DevonFulcher, line 34: dbt-mcp is very likely not running in CI, so we should remove this part.

src/dbt_mcp/prompts/dbt_cli/build.md
```diff
@@ -1,8 +1,74 @@
...
+
+### ❌ DON’T
+* Combine `--select` _and_ `--exclude` in ways that leave orphaned tests; they will still execute.
```
DevonFulcher, line 37: I don't think this is relevant for dbt-mcp.

src/dbt_mcp/prompts/dbt_cli/build.md
```diff
@@ -1,8 +1,74 @@
...
+* Expect `dbt build` to skip tests—**tests always run** for selected nodes.
```
DevonFulcher, line 38: Can you explain this bullet point?

src/dbt_mcp/prompts/dbt_cli/build.md
```diff
@@ -1,8 +1,74 @@
...
+
+### 3. Build a model _and_ its parents (useful for local debugging)
+```bash
+dbt build --select +my_model
+```
+
+### 4. CI / PR build: only modified resources
+```bash
+dbt build --select state:modified+
+```
```
DevonFulcher, line 62: ditto here.

src/dbt_mcp/prompts/dbt_cli/build.md
```diff
@@ -1,8 +1,74 @@
...
+```bash
+dbt build --select fqn:*marts*
+```
+
+---
+
+## Related Commands  
+* **`dbt run`** – Only materialise models (no tests / snapshots / seeds).  
+* **`dbt test`** – Run tests only.  
+* **`dbt compile`** – Generate SQL without executing it.
```
DevonFulcher, line 74: We probably don't need this section.

src/dbt_mcp/prompts/dbt_cli/build.md
DevonFulcher, line 1: This is much improved. Thank you!

src/dbt_mcp/prompts/dbt_cli/compile.md
DevonFulcher, line 1: I'm going to stop reviewing here at the moment. I think the rest of the prompts have similar issues as `build.md`. Let me know when you would like another review.

DevonFulcher (changes requested):
Looks great so far!

Issue #242, opened 2025-07-23
Give titles to each tool


DevonFulcher:
The MCP spec introduced a new feature to provide titles for tools. These are meant for human consumption as opposed to tool names, which are used by systems. Providing tool titles should be a simple change that makes our tools appear nicer in agent UIs.

jairus-m:
@DevonFulcher Based on https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations, will `ToolAnnotations` implementation in #198 close this issue as well? Im about to open a pr this weekend and realizing both issues may be related.

Also the original docs for tool annotations aren't available for some reason (https://modelcontextprotocol.io/docs/concepts/tools#tool-definition-structure) but based off the [Python MCP SDK repo](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py) for the `ToolAnnotations` Class, it looks like the `title` field may indeed be relevant for creating human-readable tool names:

```python
class ToolAnnotations(BaseModel):
    """
    Additional properties describing a Tool to clients.

    NOTE: all properties in ToolAnnotations are **hints**.
    They are not guaranteed to provide a faithful description of
    tool behavior (including descriptive properties like `title`).

    Clients should never make tool use decisions based on ToolAnnotations
    received from untrusted servers.
    """

    title: str | None = None
    """A human-readable title for the tool."""

    readOnlyHint: bool | None = None
    """
    If true, the tool does not modify its environment.
    Default: false
    """

    destructiveHint: bool | None = None
    """
    If true, the tool may perform destructive updates to its environment.
    If false, the tool performs only additive updates.
    (This property is meaningful only when `readOnlyHint == false`)
    Default: true
    """

    idempotentHint: bool | None = None
    """
    If true, calling the tool repeatedly with the same arguments
    will have no additional effect on the its environment.
    (This property is meaningful only when `readOnlyHint == false`)
    Default: false
    """

    openWorldHint: bool | None = None
    """
    If true, this tool may interact with an "open world" of external
    entities. If false, the tool's domain of interaction is closed.
    For example, the world of a web search tool is open, whereas that
    of a memory tool is not.
    Default: true
    """
    model_config = ConfigDict(extra="allow")
```

DevonFulcher:
Yup! I think you did it with that PR. Thanks again.

PR #255, opened 2025-07-26
Add `ToolAnnotations`

jairus-m:
# Summary
Closes issues #198 and #242 (?).

Introduces `ToolAnnotations` to all 17 registered tools within the dbt-mcp server..  this is essentially a way to provide "hints" [(i.e human-readable titles and operational hints)](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations) that allow agent UIs to display tools in a nice format while also giving users a heads up about the natue of a tool call: is it read only? destructive? idememptoent? does it interact with external systems?

Read more [here](https://blog.marcnuri.com/mcp-tool-annotations-introduction).

# Details
- Added a new module, `src/dbt_mcp/tools/annotations.py` with the following defaults:
    - `read_only_hint=True`
      - Most tools in the dbt-mcp are read-only. It is enabled for tools that only query data or metadata, so this defaults to True.
    - `destructive_hint=False`
      - Since most tools are read-only, tools default to False. This must be explicitly overridden for operations that modify data or state (like dbt run, dbt build. etc).
      - Should the default actually be the opposite such that we assume destructuve behaviour unless explicitly set otherwise?
    - `idempotent_hint=True`
      - Again, as most tool calls are read-only, this defaults to False.
    - `open_world_hint=True`
      - Since most tools interact with external systems like a dbt project or a database, this defaults to True.
- More details on what annotations were applied to what tools:
    - Destructive Tools (?):
        - dbt build, dbt run, dbt test
    - Read-Only Tools:
        - dbt compile (?), dbt docs, dbt list, dbt parse, dbt show
        - list_metrics, get_dimensions, get_entities, query_metrics
        - get_mart_models, get_all_models, get_model_details, get_model_parents, get_model_children
- Again, each tool is also given a human-readable title to look nicer for agent UIs and users

I think a discussion can be had to vet if the applied annotations make sense here!

# Tests
I didn't add any tests for these changes, but I did update `tests/integration/initialization/test_initialization.py`. A recent change introduced in PR #224 made it so that you need to pass in the `dbt_mcp.config.Config` class as an argument to `create_dbt_mcp()` . Added the `mock_config` object  as a param to keep parity and fix the failing initialization test.

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20250726-053207.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20250726-053207.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Add ToolAnnotations
+time: 2025-07-26T05:32:07.606678-07:00
--- a/.changes/unreleased/Under the Hood-20250726-053152.yaml
+++ b/.changes/unreleased/Under the Hood-20250726-053152.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Fix initialization integration test
+time: 2025-07-26T05:31:52.185494-07:00


jairus-m:
Closed due to crazy n messed up branch from my fork and unverified commits

PR #256, opened 2025-07-26
Add `ToolAnnotations`

jairus-m:
# Summary
Closes issues #198 and #242 (?).

Introduces `ToolAnnotations` to all 17 registered tools within the dbt-mcp server..  this is essentially a way to provide "hints" [(i.e human-readable titles and operational hints)](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations) that allow agent UIs to display tools in a nice format while also giving users a heads up about the natue of a tool call: is it read only? destructive? idememptoent? does it interact with external systems?

Read more [here](https://blog.marcnuri.com/mcp-tool-annotations-introduction).

# Details
- Added a new module, `src/dbt_mcp/tools/annotations.py` with the following defaults:
    - `read_only_hint=True`
      - Most tools in the dbt-mcp are read-only. It is enabled for tools that only query data or metadata, so this defaults to True.
    - `destructive_hint=False`
      - Since most tools are read-only, tools default to False. This must be explicitly overridden for operations that modify data or state (like dbt run, dbt build. etc).
      - Should the default actually be the opposite such that we assume destructuve behaviour unless explicitly set otherwise?
    - `idempotent_hint=True`
      - Again, as most tool calls are read-only, this defaults to False.
    - `open_world_hint=True`
      - Since most tools interact with external systems like a dbt project or a database, this defaults to True.
- More details on what annotations were applied to what tools:
    - Destructive Tools (?):
        - dbt build, dbt run, dbt test
    - Read-Only Tools:
        - dbt compile (?), dbt docs, dbt list, dbt parse, dbt show
        - list_metrics, get_dimensions, get_entities, query_metrics
        - get_mart_models, get_all_models, get_model_details, get_model_parents, get_model_children
- Again, each tool is also given a human-readable title to look nicer for agent UIs and users

I think a discussion can be had to vet if the applied annotations make sense here!

# Tests
I didn't add any tests for these changes, but I did update `tests/integration/initialization/test_initialization.py`. A recent change introduced in PR #224 made it so that you need to pass in the `dbt_mcp.config.Config` class as an argument to `create_dbt_mcp()` . Added the `mock_config` object  as a param to keep parity and fix the failing initialization test.

### DIFF
--- a/src/dbt_mcp/discovery/tools.py
+++ b/src/dbt_mcp/discovery/tools.py
@@ -8,7 +8,11 @@ from dbt_mcp.discovery.client import MetadataAPIClient, ModelsFetcher
 from dbt_mcp.prompts.prompts import get_prompt
 from dbt_mcp.tools.definitions import ToolDefinition
 from dbt_mcp.tools.register import register_tools
+<<<<<<< Updated upstream
 from dbt_mcp.tools.tool_names import ToolName
+=======
+from dbt_mcp.tools.annotations import create_tool_annotations
+>>>>>>> Stashed changes
 
 logger = logging.getLogger(__name__)
 
@@ -63,22 +67,27 @@ def create_discovery_tool_definitions(config: DiscoveryConfig) -> list[ToolDefin
         ToolDefinition(
             description=get_prompt("discovery/get_mart_models"),
             fn=get_mart_models,
+            annotations=create_tool_annotations(title="Get Mart Models"),
         ),
         ToolDefinition(
             description=get_prompt("discovery/get_all_models"),
             fn=get_all_models,
+            annotations=create_tool_annotations(title="Get All Models"),
         ),
         ToolDefinition(
             description=get_prompt("discovery/get_model_details"),
             fn=get_model_details,
+            annotations=create_tool_annotations(title="Get Model Details"),
         ),
         ToolDefinition(
             description=get_prompt("discovery/get_model_parents"),
             fn=get_model_parents,
+            annotations=create_tool_annotations(title="Get Model Parents"),
         ),
         ToolDefinition(
             description=get_prompt("discovery/get_model_children"),
             fn=get_model_children,
+            annotations=create_tool_annotations(title="Get Model Children"),
         ),
     ]
 
--- a/src/dbt_mcp/tools/annotations.py
+++ b/src/dbt_mcp/tools/annotations.py
@@ -1,0 +1,28 @@
+from mcp.types import ToolAnnotations
+
+
+def create_tool_annotations(
+    title: str | None = None,
+    read_only_hint: bool = True,
+    destructive_hint: bool = False,
+    idempotent_hint: bool = True,
+    open_world_hint: bool = True,
+) -> ToolAnnotations:
+    """
+    Create tool annotations. Defaults to read-only, non-destructive,
+    idempotent, and open-world hints. Forced to explicitly set hints
+    to destructive and non-idempotent.
+    Args:
+        - title: Human-readable title for the tool
+        - read_only_hint: Whether the tool only reads data
+        - destructive_hint: Whether the tool makes destructive changes
+        - idempotent_hint: Whether repeated calls have the same effect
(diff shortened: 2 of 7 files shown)


src/dbt_mcp/tools/annotations.py
```diff
@@ -0,0 +1,28 @@
+from mcp.types import ToolAnnotations
+
+
+def create_tool_annotations(
+    title: str | None = None,
+    read_only_hint: bool = True,
+    destructive_hint: bool = False,
+    idempotent_hint: bool = True,
+    open_world_hint: bool = True,
```
DevonFulcher: This PR is great, thank you! I think we may want to update this part, though. I think it would be better if we defaulted to the safest options:

```
    read_only_hint: bool = False,
    destructive_hint: bool = True,
    idempotent_hint: bool = False,
```

It may be a bit duplicative, but I think it is better to be explicit.

Additionally, I'm not sure exactly how to interpret `open_world`. The MCP team has moved the documentation on this, and I can no longer find it. Do you know where we can find more documentation on that? I was under the impression that none of our tools are open_world because they all operate within the confines of dbt, but now I'm questioning that assumption.

jairus-m: @DevonFulcher 

Thanks! I was struggling cause I couldn't find the official docs that I first read through last week 😿 . I only really had these as reference:
- https://blog.marcnuri.com/mcp-tool-annotations-introduction
- [Python MCP SDK Repo](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py) and looking at the Pydantic model/docs

I am agreeable to going the safe route and being explicit! That makes sense to me. 

As for `open_world_hint`, my understanding is that any tool that interacts outside of your local filesystem is considered true? so if it interacts with external systems (through some sort of network call) where the state can change unpredictably and independently of the tool call itself, then i think it would be open world.

For the discovery/semantic layer tools that hit dbt Cloud or a tool that queries Snowflake, those are external because the state of dbt Cloud / Snowflake can change due to external/dynamic factors (and require external network calls). And even though dbt CLI commands are arguably acting on and affected by the state of your local file system, they are still making external calls to dbt Cloud which in my head, is `open_world_hint = true`.

But if open world means, "within the confines of a dbt project and its related systems" then I can see these being false. My assumptions were based more so on the bounds of the tool (local machine vs external systems).

jairus-m: @DevonFulcher 

Made some updates in the last two commits, but still very much open to define and reset my priors on the `open_world_hint` param. I added the original documentation from the MCP `ToolAnnotations` pydantic model to the docstring and made every use of `create_tool_annotations` explicitly call out the params (except for open world).

Should `open_world_hint` also be explicit here?

DevonFulcher: Ya, I think that makes sense regarding `open_world`. It is better to be cautious and consider these tools to impact the open world. We can always change that once we learn more. Changing a tool from not `open_world` to `open_world` seems less of a breaking change than the other way around. I'm taking another look now.

src/dbt_mcp/discovery/tools.py
```diff
@@ -63,22 +64,47 @@ def get_model_children(
...
+                title="Get Model Parents",
+                read_only_hint=True,
+                destructive_hint=False,
+                idempotent_hint=True,
+            ),
         ),
         ToolDefinition(
             description=get_prompt("discovery/get_model_children"),
             fn=get_model_children,
+            annotations=create_tool_annotations(title="Get Model Children"),
```
DevonFulcher: I think you forgot to update this one.

DevonFulcher (commented):
Overall looks good! I broadly agree with the hint values. I think all there is to do is fix get_model_children

DevonFulcher:
I tried running this with inspector, but it looks like it doesn't support annotations. Regardless, nothing broke, so I feel good about this PR.

src/dbt_mcp/discovery/tools.py
jairus-m: Ah. Good eye!!

PR #263, opened 2025-07-31
Tool policies

DevonFulcher:
dbt Labs' policies dictate that we do not feed row-level data to LLMs in our systems. There is no such restriction for typical users of dbt-mcp because they have their own terms with LLM providers. However, when we use dbt-mcp internally, we need to ensure that we don't send tool results with row-level data to LLMs. This PR applies policies to each tool to ensure we are always compliant.

I had to change the get_model_details tool to return raw SQL instead of compiled SQL in order for it to be compliant with metadata only.


jairus-m:
I'm actually very intrigued by this PR as it's very relevant to a lot of the internal conversations I've been having with one of my clients about using dbt with MCP and Claude. 

The specific thing that I was curious by, is dbt's internal policy that row-level data cannot be fed into LLM systems. For the  `ToolPolicy`  class you created, is that currently the first pass at creating metadata and/or declarations for a future state of being able to configure these policies to control the behavior of what is allowed to be returned to the LLM? 

I ask as we have a similar policy but specific to PII fields and folks flagged `dbt show` as a possible tool call that would breach this policy. It would be cool to have the option to control the results of these tool calls to be safe and compliant. 

Am i understanding this? Or are policies already being enforced through this PR?

DevonFulcher:
> The specific thing that I was curious by, is dbt's internal policy that row-level data cannot be fed into LLM systems.

Yes, that is right. That policy applies to our internal copilot product, not usage of dbt-mcp with external clients like Claude.

> For the ToolPolicy class you created, is that currently the first pass at creating metadata and/or declarations for a future state of being able to configure these policies to control the behavior of what is allowed to be returned to the LLM?
> ...
> Or are policies already being enforced through this PR?

This PR is just declaring the policies; it doesn't enforce them. In the backend, we will enforce these policies for upcoming products that we build.

> I ask as we have a similar policy but specific to PII fields and folks flagged dbt show as a possible tool call that would breach this policy. It would be cool to have the option to control the results of these tool calls to be safe and compliant.

This is interesting because we are always interested in enhancing our security and compliance posture for dbt-mcp. I'm not sure the best way to make tools like `dbt show` compliant with your situation. What we are likely going to do internally is redact the output of `result_set` tools before feeding them into an LLM. I'm not sure if this will work well or if it is broadly applicable enough to include in dbt-mcp rather than just internally.

jairus-m:
@DevonFulcher Ahh thanks for the context and helping to clarify my understanding. This makes sense. As for the last piece, we are also having similar talks about different paths to addressing this.. like somehow limiting the LLM agent's access role to non-pii data only, redacting the output of `result_set`(like what you shared), or having a another layer to ["sanitize"](https://konghq.com/blog/enterprise/building-pii-sanitization-for-llms-and-agentic-ai) the output from the result before being shared with the LLM.

..and yeah, I don't think this particular situation is going to be applicable enough for any specific implementation in the dbt-mcp itself. But lots to think about on this topic regardless and appreciate the discussion!

alan-andrade (approved):
✅

PR #270, opened 2025-08-05
Add --full-refresh support for incremental builds and runs

### LINKED ISSUE #239: Support for --full-refresh argument for incremental models (and potentially all CLI args)
Building on #107 and #153, there are missing CLI args that my workflow would need to be successful.  I use a lot of incremental models, but currently I cannot run a full refresh.  The work around is to use something like `dbt show --in-line 'DROP TABLE {{ ref(some_model) }}'` and then `dbt run` but this leave a lot to be desired.

It would be great to add `--full-refresh` to `dbt run` and `dbt build`.

Thinking about the utility of the `dbt-cli` part of this, I'm curious why all the CLI args are not supported?  This seems like it could be bootstrapped from dbt itself.  I'm probably missing something here though.  Thanks!

jairus-m:
# Summary
This PR closes #239 and adds the `--full-refresh` flag to the `build` and `run` commands to support complete rebuilds of incremental models.

# Details
`src/dbt_mcp/dbt_cli/tools.py`:
- Added `is_full_refresh: bool` argument to the internal `_run_dbt_command` function which appends the `--full-refresh` flag to the command list if true
- Added `is_full_refresh` param to `build` and `run` tools using a Pydantic `Field` object with a description
         loaded from a new prompt file

`src/dbt_mcp/prompts/dbt_cli/args/full_refresh.md`:
- Added a new prompt file to provide a clear description of the `--full-refresh flag` for future LLM agents

# Tests 
`tests/unit/dbt_cli/test_tools.py`
- Added a new parameterized unit test (`test_full_refresh_flag_added_to_command)` to explicitly verify that the `--full-refresh` flag is correctly applied to both the `run` and `build` commands when `is_full_refresh=True`
   - This new test follows the existing mock/fixture pattern to isolate the command building logic without executing an actual subprocess/dbt command

### MCP Inspector
I was able to get the MCP inspector running and was able to set the `--full-refresh` option in the `build` and `run` tools alongside the `selector` param and successfully ran both tools with full refresh!

Build Tool in UI:

Screenshot 2025-08-05 at 1 17 41 PM

GitHub Copilot Pointed to dev dbt-mcp:

Asking copilot to test both `build` and `run` with `--full-refresh`:
Screenshot 2025-08-05 at 10 15 47 AM

### DIFF
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -77,7 +77,7 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
         selector: str | None = Field(
             default=None, description=get_prompt("dbt_cli/args/selectors")
         ),
-        is_full_refresh: bool = Field(
+        is_full_refresh: bool | None = Field(
             default=False,
             description=get_prompt("dbt_cli/args/full_refresh"),
         ),
@@ -115,7 +115,7 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
         selector: str | None = Field(
             default=None, description=get_prompt("dbt_cli/args/selectors")
         ),
-        is_full_refresh: bool = Field(
+        is_full_refresh: bool | None = Field(
             default=False,
             description=get_prompt("dbt_cli/args/full_refresh"),
         ),


jairus-m:
Close due to
- Using main branch on my fork
- Having a messed up diff that included changes to the `CHANGLELOG`
- did not have pre-commits on

Instead, re-opening PR with a clean slate (and also fixing mypy type error check)

PR #279, opened 2025-08-11
LangGraph create_react_agent example

DevonFulcher:
An example conversational agent built with LangGraph.


examples/langgraph_agent/main.py
```diff
@@ -0,0 +1,67 @@
...
+            "dbt": {
+                "url": url,
+                "headers": headers,
+                "transport": "streamable_http",
+            }
+        }
+    )
+    tools = await client.get_tools()
+    agent = create_react_agent(
+        model="anthropic:claude-3-7-sonnet-latest",
```
b-per, line 50: How do we authenticate to query the model?

If this is via an env var, can we mention it in the README?

b-per (approved):
OK with it but left a comment on the authentication to Anthropic

examples/langgraph_agent/main.py
reneruck-dbt: that's either explicitly via parameter or each AI vender SDK is looking for an env variable with the API key.

DevonFulcher: Yes, it is an env var. Sorry, I forgot to mention that. I'll add it to the README.

PR #281, opened 2025-08-12
Add  for non-time group bys so that it doesn't fail validations

### LINKED ISSUE #260: "'grain' is a required property" error
I often get an Input validation error from fast mcp:

`fastmcp.exceptions.ToolError: Input validation error: 'grain' is a required property`.

I narrowed this down to a fast mcp validation checking the `group_by` parameters for metrics queries, and not validating when the `grain` property is not explicitly present in the `GroupByParam`s.

In the [semantic layer SDK](https://github.com/dbt-labs/semantic-layer-sdk-python/blob/main/dbtsl/api/shared/query_params.py#L15), `grain` in `GroupByParam` is typed as optional, however, I think because there isn't a default specified the fast mcp validation is failing when a query is submitted without the `grain` property.

 I noticed that in `query_metrics.md` in this repo the few shot examples don't align with this expectation from the semantic layer SDK.

For example,

``` bash
<example>
Question: "What's our customer satisfaction score by region?"
    Thinking step-by-step:
    - First, I need to check if we have a customer satisfaction metric
    - Using list_metrics(), I find we don't have a direct "customer_satisfaction" metric
    - I should check what metrics we do have that might be related
    - I see we have "net_promoter_score" and "customer_retention_rate"
    - I should inform the user that we don't have a direct customer satisfaction metric
    - I can suggest using NPS as a proxy for customer satisfaction
    Response to user:
    "I don't have a direct customer satisfaction metric, but I can show you Net Promoter Score (NPS) by region, which is often used as a proxy for customer satisfaction. Would you like to see that instead?"
    If user agrees, then:
    Parameters:
    metrics=["net_promoter_score"]
    group_by=[{"name": "region", "type": "dimension"}]
    order_by=[{"name": "net_promoter_score", "desc": true}]
    limit=10
</example>
```

does not have `"grain": null` in the `group_by` spec.

When I added `"grain": null` to the `group_by` specs in `query_metrics.md` I no longer get the validation error and the LLM correctly constructs metrics queries.

WilliamDee:
closes #260

### DIFF
--- a/.changes/unreleased/Bug Fix-20250813-105549.yaml
+++ b/.changes/unreleased/Bug Fix-20250813-105549.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Fix the prompt to ensure grain is passed even for non-time group by"
+time: 2025-08-13T10:55:49.072094-04:00


DevonFulcher (approved):
🚀

PR #282, opened 2025-08-12
Add list saved queries tool

### LINKED ISSUE #264: Support for saved queries in Semantic Layer tools
[Saved queries](https://docs.getdbt.com/docs/build/saved-queries) are a feature of the Semantic Layer that would be good to expose via an MCP tool. They are provided through the Python SDK: https://github.com/dbt-labs/semantic-layer-sdk-python.

WilliamDee:
closes #264 

Adding a list saved query tool that is usually used in conjunction with list metrics to find ways to make the right query


WilliamDee:
> Do we also need to add a new tool or update the existing tools to query/compile by saved query name? For instance, [here](https://github.com/dbt-labs/semantic-layer-sdk-python/blob/main/dbtsl/client/sync.pyi#L84)?

I have a separate PR for that, happy to squash into this one too

PR #284, opened 2025-08-13
Add default --limit to show tool

### LINKED ISSUE #272: dbt show should have a default limit
I saw cursor crash because the LLM was trying to execute dbt show but didn't realize it was going to be loading thousands/millions of rows. In dbt Cloud - we default to a limit (something like limit 100). I think we should do the same to save users from accidentally trying to load entire tables. They should be able to override it if they explicitly say otherwise.

jairus-m:
# Summary
Closes #272.

# Details 
- Set explicit value in Pydantic `Field` default param and remove `None` type
- Extend prompt with guiding words of advice

# Tests
I don't have an OpenAI key to run `task:client` but was able to successfully use the MCP Inspector and test in Claude/Copilot/Cursor:

Claude - on asking about the default values for show:
Screenshot 2025-08-13 at 9 48 51 AM

GitHub Copilot and Cursor had the same results and limited the data:
Screenshot 2025-08-13 at 10 13 58 AM

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20250813-070837.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20250813-070837.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Add default --limit to show tool
+time: 2025-08-13T07:08:37.699907-07:00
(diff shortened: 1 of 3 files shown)


DevonFulcher (approved):
Awesome!

PR #285, opened 2025-08-13
Define toolsets

DevonFulcher:
Just like `tools/tool_names.py` and `tools/policy.py`, it is helpful to statically define toolsets for them to be referenced.

### DIFF
--- a/.changes/unreleased/Under the Hood-20250813-160036.yaml
+++ b/.changes/unreleased/Under the Hood-20250813-160036.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Define toolsets
+time: 2025-08-13T16:00:36.907595-05:00


tests/unit/tools/test_toolsets.py
```diff
@@ -0,0 +1,29 @@
+from dbt_mcp.config.config import load_config
+from dbt_mcp.mcp.server import create_dbt_mcp
+from dbt_mcp.tools.toolsets import toolsets
+from tests.env_vars import default_env_vars_context
+
+
+async def test_toolsets_match_server_tools():
+    """Test that the ToolPolicy enum matches the tools registered in the server."""
```
alan-andrade: Is this comment outdated or `ToolPolicy` is part of the internal details?

tests/unit/tools/test_toolsets.py
```diff
@@ -0,0 +1,29 @@
...
+    sql_tool_names = {"text_to_sql", "execute_sql"}
+
+    with default_env_vars_context():
+        config = load_config()
+        dbt_mcp = await create_dbt_mcp(config)
+
+        # Get all tools from the server
+        server_tools = await dbt_mcp.list_tools()
+        # Manually adding SQL tools here because the server doesn't get them
```
alan-andrade, line 17: Why it doesn't get them?

alan-andrade (approved):
Few nit comments :shipit:

tests/unit/tools/test_toolsets.py
DevonFulcher: Oh, sorry! Ya, copy-paste error. I'll fix it.

tests/unit/tools/test_toolsets.py
DevonFulcher: In order to get the list of SQL tools, the test would need to make an authenticated network request to list them. Since this is just a unit test, we aren't enabling that.

PR #288, opened 2025-08-15
Winkler/trust signals

matt-winkler:
Adds model health analysis tool

### DIFF
--- a/src/dbt_mcp/discovery/client.py
+++ b/src/dbt_mcp/discovery/client.py
@@ -37,6 +37,69 @@ class GraphQLQueries:
         }
     """)
 
+    GET_MODEL_HEALTH = textwrap.dedent("""
+        query GetModelDetails(
+            $environmentId: BigInt!,
+            $modelsFilter: ModelAppliedFilter
+            $first: Int,
+        ) {
+            environment(id: $environmentId) {
+                applied {
+                    models(filter: $modelsFilter, first: $first) {
+                        edges {
+                            node {
+                                name
+                                uniqueId
+                                rawCode
+                                description
+                                database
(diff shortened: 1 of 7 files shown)


src/dbt_mcp/discovery/client.py
```diff
@@ -317,3 +396,23 @@ def fetch_model_children(
         if not edges:
             return []
         return edges[0]["node"]["children"]
+
+    def fetch_model_health(
+        self, model_name: str, unique_id: str | None = None
```
DevonFulcher, line 401: This matches the pattern of `fetch_model_children` & `fetch_model_parents`, but we should fix these three to make `model_name` optional, like `fetch_model_details`. I can fix all three in a follow-up, though.

src/dbt_mcp/prompts/semantic_layer/query_metrics.md
```diff
@@ -29,6 +29,14 @@ For queries that may return large amounts of data, it's recommended
...
 1. First make a query with a small limit to verify the results are what you expect
 2. Then make a follow-up query without a limit (or with a larger limit) to get the full dataset
+
+IMPORTANT:
+
+Do the below if the GET_MODEL_HEALTH tool is enabled.
+
+When responding to user requests to pull metrics data, check the health of the dbt models that
+are the parents of the dbt semantic models. Use the instructions from the "ASSESSING MODEL HEALTH"
+section of the get_model_health() prompts to do this.
```
DevonFulcher, line 39: It would be better if we could elicit this interaction from the agent without such a heavy-handed approach, but, practically speaking, if this produces better results, we should stick with it.

PR #291, opened 2025-08-19
Add dbt Admin tools to interact with the dbt platform

b-per:
Add tools to allow people to monitor day to day tasks via the MCP server, calling the dbt Admin API.
This can be enhanced in the future based on feedback.

The main concern with those tools is that returning the raw JSON for all tools fills in the context

Currently add the following tools:
- list_jobs
- get_job
- trigger_job_run
- list_runs
- get_run
- cancel_run
- retry_run
- list_run_artifacts
- get_run_artifact

This can help answer question like:
- tell me if my jobs ran successfully?
- which run returned some errors lately?
- for the run xyz, what was the issue?
- trigger the daily job right now

## TODO when merged
- update docs with the new tools
- improve the logic to allow selecting some envs and or project or not
- create follow-up issues for improving some tools (like the artifacts one)

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20250811-204758.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20250811-204758.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Add support for --full-refresh flag
+time: 2025-08-11T20:47:58.841232-07:00
--- a/.changes/unreleased/Under the Hood-20250819-194052.yaml
+++ b/.changes/unreleased/Under the Hood-20250819-194052.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Update example with OpenAI to show tool calls
+time: 2025-08-19T19:40:52.220516+02:00
--- a/examples/openai_agent/main_streamable.py
+++ b/examples/openai_agent/main_streamable.py
@@ -6,9 +6,44 @@ import os
 from agents import Agent, Runner, trace
 from agents.mcp import create_static_tool_filter
 from agents.mcp.server import MCPServerStreamableHttp
+from agents.stream_events import RawResponsesStreamEvent, RunItemStreamEvent
+from openai.types.responses import ResponseCompletedEvent, ResponseOutputMessage
 
 
-async def main():
+def print_tool_call(tool_name, params, color="yellow", show_params=True):
+    # Define color codes for different colors
+    # we could use a library like colorama but this avoids adding a dependency
+    color_codes = {
+        "grey": "\033[37m",
+        "yellow": "\033[93m",
+    }
+    color_code_reset = "\033[0m"
+
+    color_code = color_codes.get(color, color_codes["yellow"])
+    msg = f"Calling the tool {tool_name}"
+    if show_params:
+        msg += f" with params {params}"
+    print(f"{color_code}# {msg}{color_code_reset}")
+
+
+def handle_event_printing(event, show_tools_calls=True):
+    if type(event) is RunItemStreamEvent and show_tools_calls:
+        if event.name == "tool_called":
+            print_tool_call(
+                event.item.raw_item.name,
+                event.item.raw_item.arguments,
+                color="grey",
+                show_params=True,
+            )
+
+    if type(event) is RawResponsesStreamEvent:
+        if type(event.data) is ResponseCompletedEvent:
+            for output in event.data.response.output:
+                if type(output) is ResponseOutputMessage:
+                    print(output.content[0].text)
+
+
+async def main(inspect_events_tools_calls=False):
     prod_environment_id = os.environ.get("DBT_PROD_ENV_ID", os.getenv("DBT_ENV_ID"))
     token = os.environ.get("DBT_TOKEN")
     host = os.environ.get("DBT_HOST", "cloud.getdbt.com")
@@ -30,13 +65,15 @@ async def main():
                 "get_dimensions",
                 "get_entities",
                 "query_metrics",
+                "get_metrics_compiled_sql",
             ],
         ),
     ) as server:
         agent = Agent(
             name="Assistant",
-            instructions="Use the tools to answer the user's questions",
+            instructions="Use the tools to answer the user's questions. Do not invent data or sample data.",
             mcp_servers=[server],
+            model="gpt-5",
         )
         with trace(workflow_name="Conversation"):
             conversation = []
@@ -45,12 +82,19 @@ async def main():
                 if result:
                     conversation = result.to_input_list()
                 conversation.append({"role": "user", "content": input("User > ")})
-                result = await Runner.run(agent, conversation)
-                print(result.final_output)
+
+                if inspect_events_tools_calls:
+                    async for event in Runner.run_streamed(
+                        agent, conversation
+                    ).stream_events():
+                        handle_event_printing(event, show_tools_calls=True)
+                else:
+                    result = await Runner.run(agent, conversation)
+                    print(result.final_output)
 
 
 if __name__ == "__main__":
     try:
-        asyncio.run(main())
+        asyncio.run(main(inspect_events_tools_calls=True))
     except KeyboardInterrupt:
         print("\nExiting.")
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -19,6 +19,7 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
         selector: str | None = None,
         resource_type: list[str] | None = None,
         is_selectable: bool = False,
+        is_full_refresh: bool | None = False,
     ) -> str:
         try:
             # Commands that should always be quiet to reduce output verbosity
@@ -31,6 +32,9 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
                 "test",
                 "list",
             ]
+
+            if is_full_refresh is True:
+                command.append("--full-refresh")
 
             if selector:
                 selector_params = str(selector).split(" ")
@@ -74,8 +78,13 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
         selector: str | None = Field(
             default=None, description=get_prompt("dbt_cli/args/selectors")
         ),
+        is_full_refresh: bool | None = Field(
+            default=None, description=get_prompt("dbt_cli/args/full_refresh")
+        ),
     ) -> str:
-        return _run_dbt_command(["build"], selector, is_selectable=True)
+        return _run_dbt_command(
+            ["build"], selector, is_selectable=True, is_full_refresh=is_full_refresh
+        )
 
     def compile() -> str:
         return _run_dbt_command(["compile"])
@@ -106,8 +115,13 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
         selector: str | None = Field(
             default=None, description=get_prompt("dbt_cli/args/selectors")
         ),
+        is_full_refresh: bool | None = Field(
+            default=None, description=get_prompt("dbt_cli/args/full_refresh")
+        ),
     ) -> str:
-        return _run_dbt_command(["run"], selector, is_selectable=True)
+        return _run_dbt_command(
+            ["run"], selector, is_selectable=True, is_full_refresh=is_full_refresh
+        )
 
     def test(
         selector: str | None = Field(
--- a/src/dbt_mcp/prompts/dbt_cli/args/full_refresh.md
+++ b/src/dbt_mcp/prompts/dbt_cli/args/full_refresh.md
@@ -1,0 +1,1 @@
+If true, dbt will force a complete rebuild of incremental models (built from scratch) rather than processing new or modifed data.
--- a/tests/unit/dbt_cli/test_tools.py
+++ b/tests/unit/dbt_cli/test_tools.py
@@ -241,3 +241,26 @@ def test_list_command_timeout_handling(monkeypatch: MonkeyPatch, mock_fastmcp):
     result = list_tool(selector="my_model", resource_type=["model"])
     assert "Timeout: dbt command took too long to complete" in result
     assert "Try using a specific selector to narrow down the results" in result
+
+
+@pytest.mark.parametrize("command_name", ["run", "build"])
+def test_full_refresh_flag_added_to_command(
+    monkeypatch: MonkeyPatch, mock_process, mock_fastmcp, command_name
+):
+    mock_calls = []
+
+    def mock_popen(args, **kwargs):
+        mock_calls.append(args)
+        return mock_process
+
+    monkeypatch.setattr("subprocess.Popen", mock_popen)
+
+    fastmcp, tools = mock_fastmcp
+    register_dbt_cli_tools(fastmcp, mock_dbt_cli_config)
+    tool = tools[command_name]
+
+    tool(is_full_refresh=True)
+
+    assert mock_calls
+    args_list = mock_calls[0]
+    assert "--full-refresh" in args_list


src/dbt_mcp/config/config.py
```diff
@@ -41,10 +41,18 @@ class DbtCliConfig(BaseModel):
...
     multicell_account_prefix: str | None = None
     host: str
-    user_id: int
-    dev_environment_id: int
-    prod_environment_id: int
+    user_id: int | None = None
+    dev_environment_id: int | None = None
+    prod_environment_id: int | None = None
     token: str
+    account_id: int | None = None
```
DevonFulcher: Is this change necessary? It seems unrelated.

b-per: I can remove it.

src/dbt_mcp/config/config.py
```diff
@@ -96,6 +105,10 @@ def actual_disable_sql(self) -> bool:
             return self.disable_remote
         return True
 
+    @property
+    def actual_disable_admin_api(self) -> bool:
+        return self.disable_admin_api if self.disable_admin_api is not None else False
```
DevonFulcher: I think this is only necessary when there are aliases for a value. For instance, we don't have a `actual_disable_discovery_api`.

DevonFulcher: We would just need to change the default like so:
```
disable_admin_api: bool | None = Field(False, alias="DISABLE_ADMIN_API")
```

src/dbt_mcp/config/config.py
```diff
@@ -125,6 +138,8 @@ class Config(BaseModel):
     dbt_cli_config: DbtCliConfig | None = None
     discovery_config: DiscoveryConfig | None = None
     semantic_layer_config: SemanticLayerConfig | None = None
+    admin_api_config: AdminApiConfig | None = None
+    disable_admin_api: bool
```
DevonFulcher: We should follow the existing pattern. Disabling a toolset is dependent on the presence of its config, so we don't need a separate `disable_admin_api` field here.

src/dbt_mcp/config/config.py
```diff
@@ -143,18 +158,19 @@ def load_config() -> Config:
         not settings.disable_semantic_layer
         or not settings.disable_discovery
         or not settings.actual_disable_sql
+        or not settings.actual_disable_admin_api
     ):
         if not settings.actual_host:
             errors.append(
-                "DBT_HOST environment variable is required when semantic layer, discovery, or SQL tools are enabled."
+                "DBT_HOST environment variable is required when semantic layer, discovery, remote or admin API tools are enabled."
```
DevonFulcher: The toolset was renamed to SQL tools
```suggestion
                "DBT_HOST environment variable is required when semantic layer, discovery, SQL, or admin API tools are enabled."
```

src/dbt_mcp/config/config.py
```diff
@@ -172,6 +188,9 @@ def load_config() -> Config:
             errors.append(
                 "DBT_USER_ID environment variable is required when SQL tools are enabled."
             )
+    if not settings.actual_disable_admin_api:
+        # Admin API tools need basic auth but don't require user_id for all operations
+        pass
```
DevonFulcher: We can delete this, right? It seems like dead code.

src/dbt_mcp/config/config.py
```diff
@@ -187,22 +206,49 @@ def load_config() -> Config:
...
+        # Type guard: requirements check ensures these are not None
+        token = settings.dbt_token
+        host = settings.actual_host
+        account_id = settings.dbt_account_id
+        if token is not None and host is not None and account_id is not None:
+            admin_api_config = AdminApiConfig(
+                host=host,
+                token=token,
+                account_id=account_id,
+            )
```
DevonFulcher: I think it is better to follow the existing patterns/conventions here.

src/dbt_mcp/dbt_admin/client.py
```diff
@@ -0,0 +1,260 @@
...
+                if job.get("schedule")
+                else None,
+                "next_run": job.get("next_run"),
+            }
+            for job in data
+        ]
+
+        return filtered_data
+
+    def get_job(self, account_id: int, job_id: int) -> Dict[str, Any]:
```
DevonFulcher: We may want to rename this to `get_job_details` to be similar to `get_model_details`

src/dbt_mcp/dbt_admin/client.py
```diff
@@ -0,0 +1,260 @@
...
+        self, account_id: int, run_id: int, debug: bool = False
+    ) -> Dict[str, Any]:
+        """Get details for a specific run."""
+
+        # we add this for individual runs but not all of them
+        incl = "?include_related=['run_steps']"
+        if debug:
+            incl = "?include_related=['run_steps','debug_logs']"
+        result = self._make_request(
+            "GET", f"/api/v2/accounts/{account_id}/runs/{run_id}/{incl}"
```
DevonFulcher, line 185:
```suggestion
            "GET", f"/api/v2/accounts/{account_id}/runs/{run_id}{incl}"
```

src/dbt_mcp/dbt_admin/client.py
```diff
@@ -0,0 +1,260 @@
...
+        data = result.get("data", {})
+
+        # we remove the truncated debug logs, they are not very relevant
+        # if needed, in debug mode, we get the full debug logs which will be more relevant
+        for step in data.get("run_steps", []):
+            step.pop("truncated_debug_logs", None)
+
+        return data
+
+    def cancel_run(self, account_id: int, run_id: int) -> Dict[str, Any]:
```
DevonFulcher: In what scenarios would an agent cancel a run? Just curious

src/dbt_mcp/dbt_admin/client.py
```diff
@@ -0,0 +1,260 @@
...
+            headers=get_artifact_header,
+            params=params,
+        )
+        response.raise_for_status()
+
+        # Return raw content for artifacts
+        if response.headers.get("content-type", "").startswith("application/json"):
+            return response.json()
+        else:
+            return response.text
```
DevonFulcher: Do we need to parse the JSON or can we just leave it as a string?

src/dbt_mcp/dbt_admin/client.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def get_admin_api_client(config: AdminApiConfig) -> DbtAdminAPIClient:
```
DevonFulcher: Do we need this? I think we can just use the constructor.

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -0,0 +1,266 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def create_admin_api_tool_definitions(
+    admin_client: Any, admin_api_config: Any
+) -> list[ToolDefinition]:
+    def list_jobs(
+        project_id: Optional[int] = None,
+        environment_id: Optional[int] = None,
```
DevonFulcher: These identifiers should come from environment variables for now. Agents won't know which identifiers to use, so it is best to use the ones that are explicitly configured in the environment settings.

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -0,0 +1,266 @@
...
+        except Exception as e:
+            logger.error(f"Error getting job {job_id}: {e}")
+            return str(e)
+
+    def trigger_job_run(
+        job_id: int,
+        cause: str,
+        git_branch: Optional[str] = None,
+        git_sha: Optional[str] = None,
+        schema_override: Optional[str] = None,
```
DevonFulcher: These parameters are interesting. I'm unsure if the agent will know how or when to use them.

src/dbt_mcp/dbt_admin/client.py
DevonFulcher: Okay, this is actually answered here https://github.com/dbt-labs/dbt-mcp/pull/291/files#diff-e7ac296e574f0bbe8b5e5060697b44cb078169fb779e42a546448a845db66211R23-R27

src/dbt_mcp/prompts/admin_api/cancel_run.md
```diff
@@ -0,0 +1,49 @@
...
+## Important Notes
+
+- Once cancelled, a run cannot be resumed
+- Partial work may have been completed before cancellation
+- Artifacts from cancelled runs may not be available
+- Use the retry functionality if you need to re-run after cancellation
+
+## Example Usage
+
+```python
```
DevonFulcher: I think the examples should be either JSON or pseudocode. LLMs don't write Python to call MCP servers.

src/dbt_mcp/prompts/admin_api/get_run.md
```diff
@@ -0,0 +1,60 @@
...
+- Performance metrics and timing
+
+## Run Statuses
+
+- **1**: Queued - Run is waiting to start
+- **2**: Starting - Run is initializing
+- **3**: Running - Run is executing
+- **10**: Success - Run completed successfully
+- **20**: Error - Run failed with errors
+- **30**: Cancelled - Run was cancelled
```
DevonFulcher: We may want to do this number to status-name mapping in code rather than in the prompt, since it is deterministic and relies less on LLMs interpreting the numbers correctly.

DevonFulcher (commented):
Amazing work!

src/dbt_mcp/dbt_admin/client.py
b-per: Yes. Or it could be useful if when querying the status of job runs someone realizes that something is stuck or taking much longer than expected.

src/dbt_mcp/dbt_admin/client.py
b-per: Doing it also for job_run then

src/dbt_mcp/dbt_admin/client.py
b-per: Let's get the raw JSON for now. As I mentioned in the README, we should revise the artifacts tool at some point though.

Normally, artifacts contain 
- JSON files like `run_results.json` and `manifest.json` but they can be huge and will feel the context in an instant. It might be useful for people to get those downloaded, but not ingested by the LLM. Not sure if this is possible
- SQL files for the SQL that was compiled and run in the data warehouse for given models. This can be useful, but right now I am filtering these from `list_artifacts` because otherwise the answer we get is to big and fills the context

Long story short. We can ingest as text for now and we can revise in the future to improve the experience.

src/dbt_mcp/dbt_admin/client.py
b-per: We actually need the `/` otherwise we get redirected to it

src/dbt_mcp/config/config.py
b-per: I think I got it back to what it was 😄

src/dbt_mcp/dbt_admin/tools.py
b-per: I agree with this and had comments in the TODO about how we might want to scope things to specific projects/envs.

Today, most of the remote tools are pointing to the Prod env of a given project.

But when interacting with jobs and job runs, there might be a few different use cases
1. people are only interested in prod jobs of 1 project
2. people are interested in jobs in different envs for 1 project (maybe a prod env and a CI env)
3. some other people might be interested with jobs/runs across all projects they have access to (this would be the first tool in the MCP server that would be cross-project in this case)

For now, I am keen to reuse the `prod_env_id` if it is set as an env var and think about use cases 2 and 3 later

src/dbt_mcp/dbt_admin/tools.py
b-per: It actually managed pretty well! I think we can revisit this based on feeback.
image

src/dbt_mcp/prompts/admin_api/get_run.md
b-per: In this use case it doesn't actually matter because we don't send this info and just retrieve it + the JSON payload also has the "humanized" version of those statuses. I will just remove this section from the md file.

But it makes sense to update it in the `list_jobs_runs.md` where here we can use those as parameters. I will update them.

```
- **status** (optional, integer): Filter runs by status. 1=Queued, 2=Starting, 3=Running, 10=Success, 20=Error, 30=Cancelled
```

src/dbt_mcp/config/config.py
```diff
@@ -143,18 +154,19 @@ def load_config() -> Config:
...
     ):
         if not settings.actual_host:
             errors.append(
-                "DBT_HOST environment variable is required when semantic layer, discovery, or SQL tools are enabled."
+                "DBT_HOST environment variable is required when semantic layer, discovery, SQL or admin API tools are enabled."
             )
         if not settings.actual_prod_environment_id:
             errors.append(
-                "DBT_PROD_ENV_ID environment variable is required when semantic layer, discovery, or SQL tools are enabled."
+                "DBT_PROD_ENV_ID environment variable is required when semantic layer, discovery, remote or admin API tools are enabled."
```
DevonFulcher:
```suggestion
                "DBT_PROD_ENV_ID environment variable is required when semantic layer, discovery, SQL, or admin API tools are enabled."
```

src/dbt_mcp/config/config.py
```diff
@@ -143,18 +154,19 @@ def load_config() -> Config:
...
             )
         if not settings.dbt_token:
             errors.append(
-                "DBT_TOKEN environment variable is required when semantic layer, discovery, or SQL tools are enabled."
+                "DBT_TOKEN environment variable is required when semantic layer, discovery, remote or admin API tools are enabled."
```
DevonFulcher:
```suggestion
                "DBT_TOKEN environment variable is required when semantic layer, discovery, SQL, or admin API tools are enabled."
```

src/dbt_mcp/config/config.py
```diff
@@ -41,10 +41,18 @@ class DbtCliConfig(BaseModel):
 class SqlConfig(BaseModel):
     multicell_account_prefix: str | None = None
     host: str
-    user_id: int
-    dev_environment_id: int
-    prod_environment_id: int
+    user_id: int | None = None
+    dev_environment_id: int | None = None
+    prod_environment_id: int | None = None
```
DevonFulcher: I'm not sure if this got lost in the last review, but I think these shouldn't change, right?

src/dbt_mcp/config/config.py
b-per: Oh, I forgot some!

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -0,0 +1,279 @@
...
+            return admin_client.trigger_job_run(
+                admin_api_config.account_id, job_id, cause, **kwargs
+            )
+        except Exception as e:
+            logger.error(f"Error triggering job {job_id}: {e}")
+            return str(e)
+
+    def list_jobs_runs(
+        job_id: Optional[int] = None,
+        status: Optional[str] = None,
```
DevonFulcher: We may want to make this an enum for more determinism. If we do, FastMCP should provide a better JSON schema to the agent.

src/dbt_mcp/config/config.py
b-per: There were comments about some stuff that got changed around `SqlConfig` but not this in particular. Thanks for spotting. Reversing this now.

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -0,0 +1,279 @@
...
+                destructive_hint=False,
+                idempotent_hint=True,
+            ),
+        ),
+    ]
+
+
+def register_admin_api_tools(
+    dbt_mcp: FastMCP,
+    config: Config,
```
DevonFulcher: Non-blocking: we may want to pass in the AdminConfig instead of Config here for consistency with other toolsets.

b-per: Handled it 👍

src/dbt_mcp/dbt_admin/tools.py
b-per: Done

PR #294, opened 2025-08-21
Using streamable http for SQL tools

DevonFulcher:
SQL tools were originally implemented before streamable http was a viable option and before we had a remote MCP server. Now that we have the remote dbt MCP server, we can use that to make the SQL tools available. This should make our code simpler since remote tools are only defined in one place. I tested this with `task client`.


alan-andrade (approved):
Nice!! Thanks Devon

PR #297, opened 2025-08-22
handle AdminAPI client host and url differently

reneruck-dbt:
Assemble the URL for the admin API client at config load time and pass along a finished url instead of the host


DevonFulcher (approved):
Looks great! Please test it out manually before merging while we don't have automated integration tests yet. There is some guidance [here](https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#testing). Personally, I typically use `task client` to test.

PR #299, opened 2025-08-25
Remove redundant and outdated documentation

DevonFulcher:
We moved our documentation to our docs site [here](https://docs.getdbt.com/docs/dbt-ai/about-mcp). So, this documentation is largely redundant, and some of it is outdated. This makes the dbt MCP readme more similar to the dbt-core readme [here](https://github.com/dbt-labs/dbt-core).


b-per:
Thanks! What about a small section to point people to the `examples` folder?

PR #301, opened 2025-08-27
Add support for `--vars` flag

### LINKED ISSUE #153: Add support for cli args
In issue https://github.com/dbt-labs/dbt-mcp/issues/107 a selector argument was requested and implemented.

In same fashion, it would be beneficial to have support for CLI vars. After iterating multiple times, Cursor couldn't get it working with current implementation.

Tried to run MCP with parameters:
```
{
  "selector": "--select <my_model> --vars <my_var>
}
```

And tried multiple iterations for the variable value:
- 'variable_name: value'
- "variable_name: value"
- variable_name: value
- "{\"variable_name\": \"value\"}"
- "variable_name: 'value'"

Either I'm missing one iteration that works but looks like the MCP doesn't support passing CLI variables in the command

For reference, matching dbt CLI command I'm running would be
```bash
dbt build -s my_model_name --vars 'my_variable: value'
```

jairus-m:
# Summary

Closes #153

# Details
- Added `vars` arg to `build`, `run`, `test` 
- Standardize command list modifications that add multiple elements to use `extend()` instead of list concatenation in `_run_dbt_command`

# Testing
Added two unit tests:
- `test_vars_flag_added_to_command`
- `test_vars_not_added_when_none`

### MCP Inspector
Looking at `build` tool:
Screenshot 2025-08-27 at 7 24 49 AM

### Cursor
Asking about the availability of `--vars`:
Screenshot 2025-08-27 at 7 18 44 AM

### Claude
Single Param:
Screenshot 2025-08-27 at 7 42 24 AM

Multi Param:
Screenshot 2025-08-27 at 7 56 13 AM

### DIFF
--- a/src/dbt_mcp/prompts/dbt_cli/args/vars.md
+++ b/src/dbt_mcp/prompts/dbt_cli/args/vars.md
@@ -1,6 +1,8 @@
 Variables can be passed to dbt commands using the vars parameter. Variables can be accessed in dbt code using `{{ var('variable_name') }}`.
 
+Note: The vars parameter must be passed as a simple STRING with no special characters (i.e "\", "\n", etc). Do not pass in a dictionary object. 
+
 Supported formats:
-- Single variable (curly brackets optional): `variable_name: value`
-- Multiple variables (curly brackets needed): `{"key1": "value1", "key2": "value2"}`
-- Mixed types: `{"string_var": "hello", "number_var": 42, "boolean_var": true}`
+- Single variable (curly brackets optional): `"variable_name: value"`
+- Multiple variables (curly brackets needed): `"{"key1": "value1", "key2": "value2"}"`
+- Mixed types: `"{"string_var": "hello", "number_var": 42, "boolean_var": true}"`


src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -36,12 +37,15 @@ def _run_dbt_command(
...
             if selector:
                 selector_params = str(selector).split(" ")
-                command = command + ["--select"] + selector_params
+                command.extend(["--select"] + selector_params)
 
             if isinstance(resource_type, Iterable):
-                command = command + ["--resource-type"] + resource_type
+                command.extend(["--resource-type"] + resource_type)
 
             full_command = command.copy()
```
jairus-m, line 50: As long as the main command is first in `_run_dbt_command` when built, the order of the flags and args after doesn't matter right?
- Is it worth re-ordering them just for the sake of making the command list order "clean" and aligned with how they're usually typed in the CLI?

For example, to order something like:
1. Main command
2. --select + arg
3. --resource-type + arg
5. --full-refresh
6. --vars + arg

DevonFulcher: Yes, I would think that any order should work. However, if it isn't broke, I wouldn't fix it. I don't think it makes a significant difference to code readability, and I don't think it affects how dbt operates or what text we feed to the LLM.

jairus-m: Totally. Makes sense to me!

PR #305, opened 2025-08-29
feat: error handling as per mcp spec

AjmeraParth132:
### Description:

This PR introduces standardized error handling in accordance with the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/server/tools?utm_source=chatgpt.com#error-handling)
 for DBT tools.

## Summary:

Implemented a lightweight wrapper for formatting exceptions as MCP-compliant error responses.

The wrapper mirrors `dbt_mcp._mcp_server._make_error_result` from FastMCP's internal implementation but avoids using the internal method directly to keep things high-level and decoupled.

This approach allows tool developers to return structured MCP errors without needing to pass around the entire dbt_mcp object within register_tools().

## Why this approach:

- Keeps error handling consistent across tools.
- Avoids tight coupling with internal/private FastMCP methods.
- Provides a simple and reusable way to return compliant error responses across multiple tools.
- This implementation has been tested across several internal tool servers and is working as expected.

Let us know if you’d like any changes or refinements!

### DIFF
--- a/.changes/unreleased/Bug Fix-20250829-223935.yaml
+++ b/.changes/unreleased/Bug Fix-20250829-223935.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: error handling as per mcp spec
+time: 2025-08-29T22:39:35.567305+05:30
--- a/src/dbt_mcp/dbt_admin/tools.py
+++ b/src/dbt_mcp/dbt_admin/tools.py
@@ -13,6 +13,7 @@ from dbt_mcp.tools.annotations import create_tool_annotations
 from dbt_mcp.tools.definitions import ToolDefinition
 from dbt_mcp.tools.register import register_tools
 from dbt_mcp.tools.tool_names import ToolName
+from dbt_mcp.tools.error_handling import make_error_result
 
 logger = logging.getLogger(__name__)
 
@@ -63,7 +64,7 @@ def create_admin_api_tool_definitions(
             logger.error(
                 f"Error listing jobs for account {admin_api_config.account_id}: {e}"
             )
-            return str(e)
+            return make_error_result(str(e))
 
     def get_job_details(job_id: int) -> dict[str, Any] | str:
         """Get details for a specific job."""
@@ -71,7 +72,7 @@ def create_admin_api_tool_definitions(
             return admin_client.get_job_details(admin_api_config.account_id, job_id)
         except Exception as e:
             logger.error(f"Error getting job {job_id}: {e}")
-            return str(e)
+            return make_error_result(str(e))
 
     def trigger_job_run(
         job_id: int,
@@ -94,7 +95,7 @@ def create_admin_api_tool_definitions(
             )
         except Exception as e:
             logger.error(f"Error triggering job {job_id}: {e}")
-            return str(e)
+            return make_error_result(str(e))
 
     def list_jobs_runs(
         job_id: int | None = None,
@@ -122,7 +123,7 @@ def create_admin_api_tool_definitions(
             logger.error(
                 f"Error listing runs for account {admin_api_config.account_id}: {e}"
             )
-            return str(e)
+            return make_error_result(str(e))
 
     def get_job_run_details(
         run_id: int,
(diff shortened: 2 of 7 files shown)


DevonFulcher (changes requested):
I’m a bit hesitant about this change. In the examples in the SDK: https://github.com/modelcontextprotocol/python-sdk, they always either return the raw results or raise an error. They don’t return a ServerResult. I would be curious to dig into the root cause of why we had to wrap everything in strings in the first place. Maybe that was a bug in FastMCP or Claude Desktop that has since been fixed? Here is the PR where I implemented that: https://github.com/dbt-labs/dbt-mcp/pull/204. It was a bit of a quick fix that we haven't prioritized going back and investigating.

AjmeraParth132:
Thanks for your comments! I dived upon it further and below are my findings, let me know how you think to proceed further.

### Why we introduced this wrapper

Fundamentally, the **MCP server already has a built-in way to catch errors** and transform them into MCP-compliant error responses using the server’s internal `_make_error_result`. See [`_make_error_result` inside `call_tool`’s decorator](https://github.com/modelcontextprotocol/python-sdk/blob/47d35f0b3ca2e59ddc26ddc0f2a816ac3dc8df9b/src/mcp/server/lowlevel/server.py#L502).

According to the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling), MCP understands errors in **two formats**:

1. **Protocol-level error** : JSON with an `error` field
2. **Tool-level error** : JSON with `isError: true`

---

### What DBT currently does

Right now, DBT catches errors and **returns them as plain strings** instead of raising:

```python
def list_metrics() -> list[MetricToolResponse] | str:
    try:
        return semantic_layer_fetcher.list_metrics()
    except Exception as e:
        return str(e)
```

This technically doesn’t raise a proper MCP error, so the client doesn't know an error has been raised

---

### Possible alternatives

There are two ways forward:

**1. Manually mimic the MCP error format (what this PR does)**
This mirrors what `_make_error_result` does internally, but without relying on private methods:

```python
def list_metrics() -> list[MetricToolResponse] | str | ServerResult:
    try:
        return semantic_layer_fetcher.list_metrics()
    except Exception as e:
        return make_error_result(str(e))
```

**2. Let the server wrapper handle error propagation**
Just let them raise naturally and allow the server to wrap them, rather than we handling them:

```python
def list_metrics() -> list[MetricToolResponse] | str:
    return semantic_layer_fetcher.list_metrics()
```

---

Both approaches work (1) ensures consistency, while (2) relies on the low level mcp server’s native error handling. In both cases, errors are well caught within our SDK’s error mechanism. Which do you think we should align on for DBT Labs?

DevonFulcher:
Thanks for looking into that, @AjmeraParth132! The second option is definitely preferable. This is what we initially had before [this](https://github.com/dbt-labs/dbt-mcp/pull/204) quick fix. However, before that PR, we saw some errors come up in Claude Desktop, specifically [here](https://github.com/dbt-labs/dbt-mcp/issues/203). It would be great to figure out why those errors were occurring so that we can just let the errors bubble up and let FastMCP handle them.

AjmeraParth132:
Hey @DevonFulcher,  

We ran a few simulations to compare the current **DBT error handling** with **FastMCP’s built-in protocol error handling**.  
From a user perspective, both approaches felt almost identical. FastMCP’s protocol handling seems quite mature now.

### Comparison

| Tool        | Current DBT Error Handling                                                                                   | Relying on FastMCP                                                                                 |
|-------------|----------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
| `docs`      | docs_w_handling | docs_wo_handling |
| `listmetrics` | listmetrics_w_handling | listmetrics_wo_handling |

---

**Next Steps**  
Do we want to test any other specific tools before deciding,  
or are we good to move forward with **protocol-level error handling** as the default?

DevonFulcher:
> From a user perspective, both approaches felt almost identical. FastMCP’s protocol handling seems quite mature now.
> Do we want to test any other specific tools before deciding,
> or are we good to move forward with protocol-level error handling as the default?

Great! As long as we test thoroughly, this change would be very beneficial. Thank you! I will also do a bit of smoke testing once this PR is ready.

DevonFulcher:
This is great! I tested it with Claude Desktop. We just need to fix CI, and it should be good to merge. Thank you!

AjmeraParth132:
Done! updated the `test_tool_handle_exceptions` CI test with checking it now in pytest raises.

PR #307, opened 2025-09-02
Extract tool config into FastMCP Context

wiggzz:
Currently creation of tools is coupled with environment configuration and set up. This presents challenges when we want to register tools before we know details about environment configuration (e.g. if the environment is _per request_).

This is one proposal on how to address this.


DevonFulcher:
We applied this method in a different PR, so this one is no longer relevant.

PR #308, opened 2025-09-04
Add exmples with Pydantic AI

b-per:
This example also prints the tools being called
Pydantic AI work with many LLMs (OpenAI, Anthropic, local ones etc...) by just changing the env var and model name


b-per:
We should stop running `mypy` on `examples`, it doesn't make sense.

DevonFulcher (approved):
Awesome! We are growing a nice collection.

PR #311, opened 2025-09-05
Oauth initial implementation

DevonFulcher:
This PR is the initial implementation of oauth login for the MCP server running locally. There are a few main components of this PR:

1. The initial oauth flow with dbt Platform login and redirect to localhost.
2. Hosting of static HTML, CSS, and JS assets with a FastAPI server.
3. APIs to retrieve dbt Project information provided by that same FastAPI server.
4. A frontend React UI to select and view dbt Projects.

These components work in concert. For example:

* When a user selects a project, that information is stored in a local yaml file `mcp.json` to be automatically reused for the next time.
* When the browser window is closed, an API request is sent to the FastAPI server to shut it down since it is no longer needed.
* After the login process is finished, the MCP server is configured based on the user's selection and continues executing as it normally would.

I tested this OAuth flow end-to-end, as well as tested the previous env var-based configuration, to ensure I didn't break anything. This feature is behind an undocumented env var while it is still being developed.


reneruck-dbt:
This is awesome and I have only a single gripe with it.
>4. A frontend React UI to select and view dbt Projects.

I'm really wondering why it has to be React.
Does some html, CSS and plain old javascript not do for a single page with some dropdowns?

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -0,0 +1,299 @@
...
+            (a for a in accounts if a.id == selected_project_request.account_id), None
+        )
+        if account is None:
+            raise ValueError(f"Account {selected_project_request.account_id} not found")
+        environments = _get_all_environments_for_project(
+            dbt_platform_url=dbt_platform_url,
+            account_id=selected_project_request.account_id,
+            project_id=selected_project_request.project_id,
+            headers=headers,
+            limit=100,
```
reneruck-dbt, line 261: Are we only every making 100 environments available to select?
I know 100 is a lot but I wouldn't be surprised if someone out there has more than 100 environments.

src/dbt_mcp/config/config.py
```diff
@@ -192,7 +259,66 @@ def load_config() -> Config:
...
+
+
+def load_config() -> Config:
+    # Load settings from environment variables using pydantic_settings
+    settings = DbtMcpSettings()  # type: ignore[call-arg]
+    dbt_platform_errors = validate_dbt_platform_settings(settings)
+    # Oauth is exerimental but secure, so you shouldn't use it,
+    # but there are no security concerns if you do.
+    enable_oauth = os.environ.get("ENABLE_EXPERIMENAL_SECURE_OAUTH") == "true"
+    if enable_oauth and dbt_platform_errors:
```
reneruck-dbt, line 274: What does it mean when the `dbt_platform_errors` list is not empty?
Why do we start an OAuth flow then?

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -0,0 +1,299 @@
...
+    return projects
+
+
+def _get_all_environments_for_project(
+    *,
+    dbt_platform_url: str,
+    account_id: int,
+    project_id: int,
+    headers: dict[str, str],
+    limit: int = 100,
```
reneruck-dbt: Nitpick:
`limit` initially confused me. 
`page_size` might make more sense in this context.

src/dbt_mcp/oauth/dbt_platform.py
```diff
@@ -0,0 +1,49 @@
...
+class SelectedProjectRequest(BaseModel):
+    account_id: int
+    project_id: int
+
+
+class DbtPlatformContext(BaseModel):
+    user_id: int
+    host_prefix: str | None = None
+    dev_environment: DbtPlatformEnvironment | None = None
+    prod_environment: DbtPlatformEnvironment | None = None
```
reneruck-dbt, line 49:
```suggestion
    prod_environment: DbtPlatformEnvironment | None = None
    
    def __add__(self, other: "DbtPlatformContext") -> Self:
        return DbtPlatformContext(
            dev_environment=other.dev_environment or self.dev_environment,
            prod_environment=other.prod_environment or self.prod_environment,
            user_id=other.user_id,
            host_prefix=other.host_prefix or self.host_prefix,
        )
      
```

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -0,0 +1,299 @@
...
+        existing_dbt_platform_context = _read_dbt_platform_context()
+        new_dbt_platform_context = DbtPlatformContext(
+            dev_environment=new_dbt_platform_context.dev_environment
+            or existing_dbt_platform_context.dev_environment,
+            prod_environment=new_dbt_platform_context.prod_environment
+            or existing_dbt_platform_context.prod_environment,
+            user_id=new_dbt_platform_context.user_id,
+            host_prefix=new_dbt_platform_context.host_prefix
+            or existing_dbt_platform_context.host_prefix,
+        )
```
reneruck-dbt:
```suggestion
        new_dbt_platform_context = existing_dbt_platform_context + new_dbt_platform_context
```

reneruck-dbt: Just an idea

src/dbt_mcp/oauth/login.py
```diff
@@ -0,0 +1,100 @@
...
+    dbt_platform_url: str,
+    port: int,
+    client_id: str,
+    config_location: Path,
+) -> LoginResult:
+    """Start OAuth login flow with PKCE using authlib and return
+    the decoded access token
+    """
+    # OAuth2 configuration
+    redirect_uri = f"http://localhost:{port}"
```
alan-andrade, line 36: Does dbt uses an allow list of redirection urls?

src/dbt_mcp/config/config.py
DevonFulcher: This means that the user's environment variables were missing some information. So, this allows the user to still configure the MCP server with environment variables and skip the whole oauth flow. If the environment variables aren't configured properly, then the OAuth flow starts so that the server can be configured without environment variables.

src/dbt_mcp/oauth/login.py
DevonFulcher: I'll answer this in Slack!

b-per:
I just tried running it locally.
- Is the backend code only on US MT at the moment? I got some error on EMEA MT
- `task client` failed until I removed the file `ui/pnpm-workspace.yaml` , and then it installed the relevant node packages

src/dbt_mcp/oauth/dbt_platform.py
DevonFulcher: Good idea, although I preferred not to overload `+` in this case.

src/dbt_mcp/oauth/fastapi_app.py
DevonFulcher: Fixed.

aiguofer:
I was looking mainly out of curiosity for the oauth flow but noticed something... it likely doesn't matter since it's running locally and won't have much concurrent traffic, but:

You really shouldn't use `async` endpoints unless you're actually using async libraries. Using `async` endpoints causes FastAPI to use the event loop to handle concurrent requests. Any blocking code will block the entire event loop, making any concurrent requests slower and harder to debug. If you just use normal `def` for your endpoints, FastAPI will instead use threads to execute concurrent requests, which for a simple scenario like this one is probably better.

reneruck-dbt:
til!
Thanks @aiguofer!

PR #321, opened 2025-09-10
Add Google ADK + local dbt-core example

Muizzkolapo:
Adds an example configuration for using Google ADK with local dbt-core installation. This provides an alternative for users who don't have dbt Cloud credentials but want to work
  with Google BigQuery datasets.

### DIFF
--- a/examples/google_adk_agent/main.py
+++ b/examples/google_adk_agent/main.py
@@ -18,9 +18,9 @@ async def main():
         print("Missing GOOGLE_GENAI_API_KEY environment variable.")
         print("Get your API key from: https://aistudio.google.com/apikey")
         return
-    
+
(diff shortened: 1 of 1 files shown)


examples/google_adk_dbt_core_agent/README.md
DevonFulcher, line 1: I think providing instructions for a local-only setup without dbt Cloud is great! However, I don't think there is anything in the code that is strictly applicable to dbt Core. I think we can change the language of this README (and the name of this directory) such that this example is more broadly applicable to either dbt Cloud or non-dbt Cloud users.

Muizzkolapo: That makes sense actually! I have changed the folder name to a more general one.

PR #322, opened 2025-09-11
Implement the `search` parameter for SL metadata

WilliamDee:
## Context

Adding a search parameter, so that the output can be filtered down for large projects.

### DIFF
--- a/.changes/unreleased/Under the Hood-20250911-132021.yaml
+++ b/.changes/unreleased/Under the Hood-20250911-132021.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Add search string to SL metadata queries
+time: 2025-09-11T13:20:21.201421-04:00


src/dbt_mcp/prompts/semantic_layer/get_dimensions.md
```diff
@@ -7,4 +7,5 @@ that describe or categorize data.
 
 <parameters>
 metrics: List of metric names
+search: Optional string used to filter dimensions by name using partial matches (only use when absolutely necessary as some dimensions might be missed due to specific naming styles)
```
DevonFulcher, line 10: Can you elaborate on "absolutely necessary"? I'm not sure myself in which situations this parameter should be used, so I imagine an agent would struggle too.

Additionally, I think this statement is okay for now, but as we improve our search capabilities, relevant resources should be pulled more robustly with semantics, and specific naming styles shouldn't matter.

src/dbt_mcp/prompts/semantic_layer/list_metrics.md
```diff
@@ -6,4 +6,8 @@ that can be used with other tools to answer the question.
 
 Examples:
 - "What are the top 5 products by revenue?"
-- "How many users did we have last month?"
\ No newline at end of file
+- "How many users did we have last month?"
+
+<parameters>
+search: Optional string used to filter metrics by name using partial matches
```
DevonFulcher, line 12: Ditto here. When should the agent use this parameter?

src/dbt_mcp/semantic_layer/client.py
```diff
@@ -78,21 +78,26 @@ def list_metrics(self) -> list[MetricToolResponse]:
...
-                    "variables": {"metrics": [{"name": m} for m in metrics]},
+                    "variables": {
+                        "metrics": [{"name": m} for m in metrics],
+                        "search": search,
+                    },
                 },
             )
             dimensions = []
-            for d in dimensions_result["data"]["dimensions"]:
+            for d in dimensions_result["data"]["dimensionsPaginated"]["items"]:
```
DevonFulcher, line 100: Does the search parameter ensure that only the first page is returned, or do we actually need to loop through the pages?

src/dbt_mcp/prompts/semantic_layer/get_dimensions.md
WilliamDee: The only thing I don't like about this search parameter is that there could very likely be cases where it misses certain things due to the name not matching exactly to certain things. Ie., "Whats the annual revenue?" - the LLM might end up running it with `search="revenue"` and in that case it'll miss metrics with names like `arr`. I prompted that it should be used when absolutely necessary to have it not use it that often until we have better searching/filtering capabilities via semantics search.

src/dbt_mcp/semantic_layer/client.py
WilliamDee: for this iteration, we aren't using the paginated fields, so it'll by default return all.

PR #326, opened 2025-09-13
 Add dbt-codegen MCP toolset 

Muizzkolapo:
This PR adds support for dbt-codegen package operations through the MCP protocol, enabling AI assistants to automatically generate dbt boilerplate code.

  What we are adding
  ***4 new MCP tools for dbt-codegen operations:***

    - generate_source: Creates source YAML definitions from database schemas
    - generate_model_yaml: Generates documentation for existing dbt models
    - generate_base_model: Creates staging SQL models from sources
    - generate_model_import_ctes: Generates CTE imports for model dependencies

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -46,6 +46,13 @@ class DbtCliConfig(BaseModel):
     dbt_cli_timeout: int
     binary_type: BinaryType
 
+class DbtCodegenConfig(BaseModel):
+    project_dir: str
+    dbt_path: str
+    dbt_cli_timeout: int
+    binary_type: BinaryType
+
+
 
 class SqlConfig(BaseModel):
     host_prefix: str | None = None
@@ -93,6 +100,7 @@ class DbtMcpSettings(BaseSettings):
 
     # Disable tool settings
     disable_dbt_cli: bool = Field(False, alias="DISABLE_DBT_CLI")
+    disable_dbt_codegen: bool = Field(False, alias="DISABLE_DBT_CODEGEN")
     disable_semantic_layer: bool = Field(False, alias="DISABLE_SEMANTIC_LAYER")
     disable_discovery: bool = Field(False, alias="DISABLE_DISCOVERY")
     disable_remote: bool | None = Field(None, alias="DISABLE_REMOTE")
@@ -156,6 +164,7 @@ class Config(BaseModel):
     tracking_config: TrackingConfig
     sql_config: SqlConfig | None = None
     dbt_cli_config: DbtCliConfig | None = None
+    dbt_codegen_config: DbtCodegenConfig | None = None
     discovery_config: DiscoveryConfig | None = None
     semantic_layer_config: SemanticLayerConfig | None = None
     admin_api_config: AdminApiConfig | None = None
@@ -380,6 +389,16 @@ def create_config(settings: DbtMcpSettings) -> Config:
             binary_type=binary_type,
         )
 
+    dbt_codegen_config = None
+    if not settings.disable_dbt_codegen and settings.dbt_project_dir and settings.dbt_path:
+        binary_type = detect_binary_type(settings.dbt_path)
+        dbt_codegen_config = DbtCodegenConfig(
+            project_dir=settings.dbt_project_dir,
+            dbt_path=settings.dbt_path,
+            dbt_cli_timeout=settings.dbt_cli_timeout,
+            binary_type=binary_type,
+        )
+        
     discovery_config = None
     if (
         not settings.disable_discovery
@@ -451,6 +470,7 @@ def create_config(settings: DbtMcpSettings) -> Config:
         ),
         sql_config=sql_config,
         dbt_cli_config=dbt_cli_config,
+        dbt_codegen_config=dbt_codegen_config,
         discovery_config=discovery_config,
         semantic_layer_config=semantic_layer_config,
         admin_api_config=admin_api_config,
--- a/src/dbt_mcp/dbt_codegen/__init__.py
+++ b/src/dbt_mcp/dbt_codegen/__init__.py
@@ -1,0 +1,1 @@
+"""dbt-codegen MCP tools for automated dbt code generation."""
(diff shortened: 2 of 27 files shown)


b-per:
Thanks @Muizzkolapo . I'll spend some time reviewing it in the next few days.

Muizzkolapo:
> Thanks @Muizzkolapo . I'll spend some time reviewing it in the next few days.

no worries, I tried to break the pr down into bits but it just kept getting bigger and bigger.

b-per:
Quick update! I am checking it at the moment but there is quite a bit of work to resolve conflicts with main first.

Muizzkolapo:
> Quick update! I am checking it at the moment but there is quite a bit of work to resolve conflicts with main first.

do you need me to rebase with main?

b-per:
> do you need me to rebase with main?

I think I managed to do it.

b-per:
I have been trying it on my side and here is some ideas/feedback that I'm keen to discuss and be challenged on.

### Overall comment

For the tools to work, `codegen` needs to be installed. This makes me think that we should maybe deactivate the codegen tools by default and make it opt-in instead of opt-out. Otherwise, people trying to run those tools will get errors.

### Individual tools comments

We also start to have (too) many tools in the dbt MCP server, which
1. makes it more complicated for the LLM to pick which one it should choose
2. fills the context just with MCP metadata/tools info

This makes me think. Are all the tools of codegen really required?

- `base_model_creation` + `create_base_models`
  - we could let the LLM create the files after a `generate_base_model` without relying on extra tools, no? 
- `generate_model_import_ctes`: Generates CTE imports for model dependencies
  - this, to me, feels like something an LLM could do easily without tools, no?
- `generate_source`: Creates source YAML definitions from database schemas
  - i'm ok with this one, it adds value that the LLM might struggle with by itself
- `generate_model_yaml`: Generates documentation for existing dbt models
  - i'm ok with this one, it adds value that the LLM might struggle with by itself
- `generate_base_model`: Creates staging SQL models from sources
  - i'm ok with this one, but I think we should rename it to generate_staging_model. Yes, the tool name will differ from the macro but this is the taxonomy most people use today for the model on top of the source


tldr - my suggestion:
- toolset off by default
- remove `base_model_creation`, `create_base_models` and `generate_model_import_ctes`
- rename `generate_base_model` to `generate_staging_model`
 
What do people think about this?

Muizzkolapo:
> I have been trying it on my side and here is some ideas/feedback that I'm keen to discuss and be challenged on.
> 
> ### Overall comment
> For the tools to work, `codegen` needs to be installed. This makes me think that we should maybe deactivate the codegen tools by default and make it opt-in instead of opt-out. Otherwise, people trying to run those tools will get errors.
> 
> ### Individual tools comments
> We also start to have (too) many tools in the dbt MCP server, which
> 
> 1. makes it more complicated for the LLM to pick which one it should choose
> 2. fills the context just with MCP metadata/tools info
> 
> This makes me think. Are all the tools of codegen really required?
> 
> * `base_model_creation` + `create_base_models`
>   
>   * we could let the LLM create the files after a `generate_base_model` without relying on extra tools, no?
> * `generate_model_import_ctes`: Generates CTE imports for model dependencies
>   
>   * this, to me, feels like something an LLM could do easily without tools, no?
> * `generate_source`: Creates source YAML definitions from database schemas
>   
>   * i'm ok with this one, it adds value that the LLM might struggle with by itself
> * `generate_model_yaml`: Generates documentation for existing dbt models
>   
>   * i'm ok with this one, it adds value that the LLM might struggle with by itself
> * `generate_base_model`: Creates staging SQL models from sources
>   
>   * i'm ok with this one, but I think we should rename it to generate_staging_model. Yes, the tool name will differ from the macro but this is the taxonomy most people use today for the model on top of the source
> 
> tldr - my suggestion:
> 
> * toolset off by default
> * remove `base_model_creation`, `create_base_models` and `generate_model_import_ctes`
> * rename `generate_base_model` to `generate_staging_model`
> 
> What do people think about this?

This actually makes sense, I did not even consider the choice overload for the LLM. I think we can strip it down to the essentials that would add value. I am fine doing this if others are in agreement then.

Muizzkolapo:
> I have been trying it on my side and here is some ideas/feedback that I'm keen to discuss and be challenged on.
> 
> ### Overall comment
> For the tools to work, `codegen` needs to be installed. This makes me think that we should maybe deactivate the codegen tools by default and make it opt-in instead of opt-out. Otherwise, people trying to run those tools will get errors.
> 
> ### Individual tools comments
> We also start to have (too) many tools in the dbt MCP server, which
> 
> 1. makes it more complicated for the LLM to pick which one it should choose
> 2. fills the context just with MCP metadata/tools info
> 
> This makes me think. Are all the tools of codegen really required?
> 
> * `base_model_creation` + `create_base_models`
>   
>   * we could let the LLM create the files after a `generate_base_model` without relying on extra tools, no?
> * `generate_model_import_ctes`: Generates CTE imports for model dependencies
>   
>   * this, to me, feels like something an LLM could do easily without tools, no?
> * `generate_source`: Creates source YAML definitions from database schemas
>   
>   * i'm ok with this one, it adds value that the LLM might struggle with by itself
> * `generate_model_yaml`: Generates documentation for existing dbt models
>   
>   * i'm ok with this one, it adds value that the LLM might struggle with by itself
> * `generate_base_model`: Creates staging SQL models from sources
>   
>   * i'm ok with this one, but I think we should rename it to generate_staging_model. Yes, the tool name will differ from the macro but this is the taxonomy most people use today for the model on top of the source
> 
> tldr - my suggestion:
> 
> * toolset off by default
> * remove `base_model_creation`, `create_base_models` and `generate_model_import_ctes`
> * rename `generate_base_model` to `generate_staging_model`
> 
> What do people think about this?

This actually makes sense, I did not even consider the choice overload for the LLM. I think we can strip it down to the essentials that would add value. I am fine doing this if others are in agreement then.

b-per:
Thanks @Muizzkolapo . Feel free to do the updates, of if you are short on time we can also take it forward on our side

Muizzkolapo:
> Thanks @Muizzkolapo . Feel free to do the updates, of if you are short on time we can also take it forward on our side

I had started over the weekend just gonna make a few tweaks and push

b-per (approved):
Thanks @Muizzkolapo 
I will raise a PR to update the docs

PR #331, opened 2025-09-16
Improve dbt platform context mcp.yml parsing

DevonFulcher:
I received some errors pertaining to a corrupted mcp.yml file. I'm not sure how the file became invalid, but this update should make this code more robust. I tested with `task client`.


alan-andrade (approved):
✅

PR #332, opened 2025-09-16
Fix partially initialized dbt platform context

DevonFulcher:
This PR fixes a bug when the user would exit the login flow before selecting a project. This caused a partially initialized dbt platform context to be used, which would cause downstream errors.


alan-andrade (approved):
✅  Nice, thanks!

PR #335, opened 2025-09-18
Add `get_job_run_error` to Admin API tools

### LINKED ISSUE #310: Add `get_job_run_error` to the Admin API tools
The ability to retrieve dbt Cloud artifacts/details via `get_job_run_details` and `get_job_run_artifacts` is great! However, for projects with _many_ nodes, the context that is read in by these tools often exceed the LLM agent's token limit. 

In addition, a good majority of the use of these artifacts (in my experience) have to do with understanding errors that caused a run job to fail. 

My hope is to implement and add a tool `get_job_run_error` that will retrieve relevant errors without having to read-in the rest of the details/artifacts into the agent's context window.

May be worth a discussion!

Here is an example of hitting the token limit:

Image

jairus-m:
## Summary
Adds a specific tool, `get_job_run_error`, to the Admin API tools. The overall goal is to be able to use the MCP in order to return the specific error(s) that failed a job run in dbt Cloud with the minimal context needed for an LLM to diagnose and fix the problem. Currently, getting errors via `get_job_run_details` only returns a failed run step and `get_job_run_artifacts` leads to too much context and exceeds the token limit of agents like Claude, Copilot, etc.

## What Changed

Changes for the core additions can be seen in the following file tree:
```bash
src/dbt_mcp/dbt_admin/
├── client.py                    # Existing DbtAdminAPICLient
├── constatnts.py                # Enums and step string
├── tools.py                     # Existing MCP tool definitions (add `get_job_run_error`) 
└── run_results_errors/          # New module
    ├── __init__.py  
    ├── config.py                # Pydantic schemas to validate artifact input / dict output
    └── parser.py                # ErrorFetcher (fetch errors to create output for LLM)
```

The majority of the parsing logic lives in the`ErrorFetcher` class. Specifically, it parses the `run_results.json` artifact. The flow of its class methods are as follows:

Main Entry Point
  - `analyze_run_errors()`: Main and only public method that orchestrates the error fetching/parsing

Core Processing Steps
- `_find_all_failed_steps()`: Goes through "run_steps" to locate all steps with error status (code 20)
- `_get_failure_details()`: Extracts detailed failure info from each failed step
- `_fetch_run_results_artifact()`: Makes API call to retrieve the correct `run_results.json` from the failed step (using step index)
- `_parse_run_results()`: Handles JSON parsing and validation of the incoming artifact content
- `_extract_errors_from_results()`: Filters through artifact to find items with "error" or "fail" status
- `_build_error_response()`: Constructs the final structured response

Utility Methods
- `_create_error_result()`: Creates standardized error objects
- `_handle_artifact_error()`: Handles cases where artifacts are unavailable or invalid with special logic for source freshness errors

## Why

- Using the Admin API MCP tools (and others) can exceeded token limits
  - In this case, getting artifacts and trying to parse them 
- Work with client led to discussions for a desire to have an error fetching tool to automate the diagnosis and fix of production failures (as their dbt project has operational components)

## Related Issues

Closes #310

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required
    - [PR](https://github.com/dbt-labs/docs.getdbt.com/pull/7965)

## Additional Notes

---
### Testing
- Tested with Adhoc script across multiple failure scenarios to see the final output (in the comments below)
- Tested with MCP inspector across the same failure scenarios
- TO DO: Test with LLM


jairus-m:
Adhoc script to test the functionality of the `ErrorFetcher` class:

```python
"""
Quick prototyping to sanity check error parsing logic.
- Lives in src/dbt_mcp/dbt_admin/run_results_errors/test_script_parser.py
- Test with a real dbt Cloud job run ID
- Set RUN_ID = <#########>

Run: `uv run src/dbt_mcp/dbt_admin/run_results_errors/test_script_parser.py`
"""

import asyncio
import json
import logging
import os

from dotenv import load_dotenv

from dbt_mcp.dbt_admin.run_results_errors.parser import ErrorFetcher
from dbt_mcp.dbt_admin.client import DbtAdminAPIClient
from dbt_mcp.config.config_providers import AdminApiConfig, ConfigProvider
from dbt_mcp.config.headers import AdminApiHeadersProvider
from dbt_mcp.oauth.token_provider import StaticTokenProvider

load_dotenv()

DBT_CLOUD_URL = os.environ["DBT_CLOUD_URL"]
DBT_TOKEN = os.environ["DBT_TOKEN"]
DBT_ACCOUNT_ID = int(os.environ["DBT_ACCOUNT_ID"])

# Redacted my saved test scenarios (i.e real prod failures)
# <job_run_id> Prod, No Failures
# <job_run_id> Prod, Source Freshness
# <job_run_id> Prod, Test Failure
# <job_run_id> Staging, Model Failure
# <job_run_id> Staging, Multiple Test Failures
# <job_run_id> Prod, Cancelled Run with No Steps Triggered
# <job_run_id> Prod, Cancelled Run with Steps Triggered
# <job_run_id> Staging, Unit Test Failure 
# <job_run_id> Prod, Snapshot Failure
# <job_run_id> Prod, Data Contract Failure
# <job_run_id> Prod, Dupe detected in Snapshot DML
# <job_run_id>  Staging, Git credential error (No run_results.json, but should return logs)
RUN_ID = <enter_your_job_run_id> 


class StaticConfigProvider(ConfigProvider[AdminApiConfig]):
    """Simple static config provider for testing."""

    def __init__(self, config: AdminApiConfig):
        self._config = config

    async def get_config(self) -> AdminApiConfig:
        return self._config


logging.basicConfig(level=logging.ERROR)


async def main():
    token_provider = StaticTokenProvider(token=DBT_TOKEN)
    headers_provider = AdminApiHeadersProvider(token_provider=token_provider)
    config = AdminApiConfig(
        url=DBT_CLOUD_URL, headers_provider=headers_provider, account_id=DBT_ACCOUNT_ID
    )

    config_provider = StaticConfigProvider(config)
    client = DbtAdminAPIClient(config_provider)

    run_details = await client.get_job_run_details(DBT_ACCOUNT_ID, RUN_ID)

    error_analyzer = ErrorFetcher(RUN_ID, run_details, client, config)
    result = await error_analyzer.analyze_run_errors()
    print(json.dumps(result, indent=2))


if __name__ == "__main__":
    asyncio.run(main())

```

jairus-m:
Light MCP validation via the MCP inspector:

### Test Failure
Screenshot 2025-09-17 at 6 00 33 PM


### Model Failure
Screenshot 2025-09-17 at 6 01 36 PM

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
jairus-m, line 1: ## `ErrorFetcher`
The majority of the parsing logic lives in the `ErrorFetcher` class. The flow of its methods are as follows: 

### Main Entry Point 
- `analyze_run_errors()`: Main and only public method that orchestrates the error fetching/parsing
    - Step 1: Find Failed Step
      - `_find_failed_step()`: Specifically goes through the "run_steps" to locate the first one with error status (code 20)
    - Step 2: Get Failure Details
      - `_get_failure_details()`: Extracts detailed failure info from error run_step (like step index, name, finished_at)
    - Step 3a: Fetch Artifact
      - `_fetch_run_results_artifact()`: Makes API call to retrieve the correct run_results.json from the failed step (using step index)
    - Step 3b: Parse Artifact (IF available)
      - `_parse_run_results()`: Handles JSON parsing and validation of the incoming artifact content
      - `_extract_errors_from_results()`: Filters through artifact to find items with "error" or "fail" status
      -` _build_error_response()`: Constructs the final structured response

### Utility Methods:
- `_create_error_result()`: Creates standardized error objects (uses Pydantic schemas)

### Error Handling:
- `_handle_artifact_error()`: Handles cases where artifacts are unavailable or invalid with "special" logic for source freshness errors (i.e spitting out a basic message about source freshness causing the error)

src/dbt_mcp/dbt_admin/run_results_errors/config.py
```diff
@@ -0,0 +1,63 @@
+from typing import Any, Dict, List, Optional
```
b-per: We use Python 3.9+ style across the code base, e.g. , we don't import `Dict` but we use `dict` directly. (we should add a linting rule about it in CI....)

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
```diff
@@ -0,0 +1,207 @@
...
+
+        except ValidationError as e:
+            logger.error(f"Schema validation failed for run {self.run_id}: {e}")
+            return self._create_error_result(f"Validation failed: {str(e)}")
+        except Exception as e:
+            logger.error(f"Error analyzing run {self.run_id}: {e}")
+            return self._create_error_result(str(e))
+
+    def _find_failed_step(self, run_details: Dict) -> Optional[Dict[str, Any]]:
+        """Find the first failed step in the run."""
```
b-per: I think that this might need to be tweaked. I saw cases where the source freshness failed but it wasn't set to fail the whole run (e.g. the source freshness checkbox was checked, but not added as a step). 

In that case a `dbt build` step failed after, but the logic in the code didn't look for the `build` step `run_results.json` and just mentioned that the failure was due to source freshness.

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
```diff
@@ -0,0 +1,207 @@
...
+        unique_id: Optional[str] = None,
+        relation_name: Optional[str] = None,
+        target: Optional[str] = None,
+        step_name: Optional[str] = None,
+        finished_at: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Create a standardized error results using MultiErrorResultSchema."""
+        error = ErrorResultSchema(
+            unique_id=unique_id,
+            relation_name=relation_name,
```
b-per, line 206: From a quick check, it seems that if it is a test that failed the `relation_name` might not be set

b-per (changes requested):
This looks nearly there. I added a couple of comments.

Let me know if I help but we should check a few different scenarios:
- what if the source freshness failed but was not supposed to break the job and another build/run/test step failed after
- what if the run was cancelled?
- does it work in case of failure due to models errors? tests errors? unit tests errors?

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
```diff
@@ -0,0 +1,207 @@
...
+    RunDetailsSchema,
+    RunResultsArtifactSchema,
+    ErrorResultSchema,
+    MultiErrorResultSchema,
+)
+
+logger = logging.getLogger(__name__)
+
+# Should re-use the JobRunStatus enum / STATUS_MAP in dbt_admin/tools.py somehow ?
+# Will lead to circular import so keeping separate for now
```
DevonFulcher: Can the circular import be fixed by moving that code to a separate file?

jairus-m:
@b-per @DevonFulcher 

Thanks for the feedback and review this far, just got back and will address most of these today.

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
jairus-m: @DevonFulcher Yes, most def! I'll add a `dbt_admin/constants.py` file with:           
- `JobRunStatus` enum (from `dbt_admin/tools.py`)
- `STATUS_MAP` dict (from `dbt_admin/tools.py`)
- `SOURCE_FRESHNESS_STEP_NAME` constant (from `dbt_admin/run_results_error/sparser.py`)

Happy to modify this if necessary

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
jairus-m: @b-per I believe the `relation_name` for a failed test is always returned as a `null` in the response. 

For model errors it does populate (`DB.SCHEMA.OBJECT`) but for tests it does not. In my test example from a production test failure, the failed test in `run_results.json` looks like:

```json
  {
      "status": "fail",
      "timing": [
        {
          "name": "compile",
          "started_at": "2025-09-10T16:31:28.938747Z",
          "completed_at": "2025-09-10T16:31:28.963736Z"
        },
        {
          "name": "execute",
          "started_at": "2025-09-10T16:31:28.965013Z",
          "completed_at": "2025-09-10T16:33:29.772792Z"
        }
      ],
      "thread_id": "Thread-233 (worker)",
      "execution_time": 120.93445444107056,
      "adapter_response": {
        "_message": "SUCCESS 1",
        "code": "SUCCESS",
        "rows_affected": 1,
        "query_id": "01bef5ff-0615-9847-006c-45010b8fe16b"
      },
      "message": "Got 30 results, configured to fail if != 0",
      "failures": 30,
      "unique_id": "test.dbt_account.unique_<table_name>_<column_name>.c1fa8884bf",
      "compiled": true,
      "compiled_code": "\n    \n    \n\nselect\n    column_name as unique_field,\n    count(*) as n_records\n\nfrom db.schema.table_name\nwhere column_name is not null\ngroup by column_name\nhaving count(*) > 1\n\n\n",
      "relation_name": null,
      "batch_results": null
    },
```

It may be an option to parse the `compiled_code` value to extract the relation name, but for tests, maybe it's fine to use the metadata from the `unique_id` which does contain info on the node itself, the name of the test, and the column that's being checked? 

Also, in general, it may be helpful to return the `compiled_code` in the final output of `get_job_run_error` as well? What do you think?

jairus-m:
> Let me know if I help but we should check a few different scenarios:
> 
> * what if the source freshness failed but was not supposed to break the job and another build/run/test step failed after
> * what if the run was cancelled?
> * does it work in case of failure due to models errors? tests errors? unit tests errors?

So far these are the scenarios I have verified that have valid output from `get_job_run_error`(and will eventually document):
1. No failures
2. Source freshness that fails the entire run
3. Model error
4. Data test failure
5. Multiple data test failures
6. Cancelled run with no run steps triggered or executed
7. Cancelled run with run steps triggered (and a few executed)

The last big scenarios to check are:
1. Source freshness failure (not set to fail job) but build continues without a further downstream run step error
2. Source freshness failure (not set to fail job) but build continues with a downstream run step error
3. Unit test error

...any others to consider?

So based on this, the remaining things to do before opening the PR:
- Surface `compiled_code` in final tool call output (???)
- Implement some logic to handle the different "source freshness" error scenarios 
- Test the remaining cases (i.e 2 source freshness and 1 unit test scenario)
- Discuss best practice unit test approach for `get_job_run_error` and implement

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
b-per: Maybe we could:
- set `relation_name` to "Not a relation" (or something similar) when `relation_name` is None? Returning a null/None just feels odd
- and yes, add the `compiled_code` as it might be useful for models and tests

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
```diff
@@ -0,0 +1,231 @@
...
+        """
+        Parse the run data and return simplified failure details with validation.
+
+        Returns:
+            Structured error information
+        """
+        try:
+            run_details = RunDetailsSchema.model_validate(self.run_details)
+
+            failed_step = self._find_failed_step(run_details.model_dump())
```
DevonFulcher: This converts run_details into a Pydantic object and then immediately converts it back into a dict. Ideally, `_find_failed_step` would accept the Pydantic object, not a dict. Otherwise, it makes sense for it to just accept the `self.run_details` dict instead of converting the Pydantic object back into a dict. Let me know if I'm missing something here.

src/dbt_mcp/dbt_admin/run_results_errors/parser.py
```diff
@@ -0,0 +1,231 @@
...
+            return MultiErrorResultSchema.model_validate(result).model_dump()
+
+        except ValidationError as e:
+            logger.error(f"Schema validation failed for run {self.run_id}: {e}")
+            return self._create_error_result(f"Validation failed: {str(e)}")
+        except Exception as e:
+            logger.error(f"Error analyzing run {self.run_id}: {e}")
+            return self._create_error_result(str(e))
+
+    def _find_failed_step(self, run_details: dict) -> Optional[dict[str, Any]]:
```
DevonFulcher: Even for short-lived variables, I prefer to use dataclasses or Pydantic models to contain the data. Would you mind changing this to return a dataclass or Pydantic model and updating the downstream functions?

jairus-m:
> I think that this might need to be tweaked. I saw cases where the source freshness failed but it wasn't set to fail the whole run (e.g. the source freshness checkbox was checked, but not added as a step).
> 
> In that case a `dbt build` step failed after, but the logic in the code didn't look for the `build` step `run_results.json` and just mentioned that the failure was due to source freshness.

Based on current projects I'm on, I was biased to assume that a failed run job is caused by a single run step error. However, this obviously isn't true with the above comment/scenario. To address this, I changed the final output to be able to capture all run step failures instead of just the first. This should cover that case (and others that end up resulting in mulitple run step errors). 

In addition, I've been able to test and verify additional scenarios to the ones I mentioned [here](https://github.com/dbt-labs/dbt-mcp/pull/335#issuecomment-3330991787):
- Unit Test Failure
- Snapshot Failure
- Data Contract Failure
- Error in Snapshot DML

I do believe the majority of cases that lead to a failed run step are captured in the `run_results.json` and should be properly captured by `ErrorFetcher` (except specifics of "source freshness" errors since we don't parse `sources.json`).

So given this, the last things to do are:
- Test the multiple run step error scenario that @b-per mentioned above
  - I currently don't have a convenient project to test this out with, so would appreciate any help here if yall have examples to test on hand!
- Add unit tests (will try and discuss general testing approach with yall next week either async or informally in a call, time-willing)
- Update documentation in dbt docs?
  - Need to understand what I need to do to add `get_job_run_error` to the official docs site

b-per:
I just did a quick test. A run failed because the credentials are wrong. And in that case, it means that no `run_results.json` was created.

The tool came back with a payload including the message `"run_results.json not available: not available"`.

I think that if there is no `run_results.json`, we could get the logs for the step and return those (truncated if they are too long).

I created [this branch](https://github.com/dbt-labs/dbt-mcp/tree/bper-get-job-run-error) with an implementation that you can copy/paste or just get some inspiration from if you want.

There is also an issue with Fusion right now where `run_results.json` is not being created when the step fails, but this is something that will get fixed soon.

jairus-m:
> I think that if there is no run_results.json, we could get the logs for the step and return those (truncated if they are too 
> long).

Done! With the changes in https://github.com/dbt-labs/dbt-mcp/pull/335/commits/6e3c26ceb28b11c983c19ee3259135e7816f3ede , the behavior is that anytime there is an attempt to retrieve the `run_results.json` in a failed run step that does not produce an artifact (such as source freshness, credential error, etc that triggers the `_handle_artifact_error` method), to return the truncated logs. Otherwise, the `truncated_logs` key will be null. I did test returning the full logs, however, for large runs, that would also overload the context. 

In addition, I tested all the scenarios listed in the [adhoc test script](https://github.com/dbt-labs/dbt-mcp/pull/335#issuecomment-3305028892) I shared above so far and have updated existing unit tests. 

Lastly, I will open this PR and again, note the outstanding "TO DOs" for myself:
1. ~~Add `get_job_run_error` to documentation:~~
    - ~~[dbt Docs](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/docs/dbt-ai/about-mcp.md)~~
    - ~~[Architecture diagram](https://github.com/dbt-labs/dbt-mcp/blob/main/docs/diagram.d2)~~
2. Confirm test scenario for source freshness error (does not fail job) but subsequent run step error for build or test failure (i.e return multiple failed run steps in output)
3. Add net-new unit tests for
    - ~~`get_job_run_error` MCP tool at minimum~~
    - ~~`ErrorFetcher` class (?)~~

b-per:
Thanks @jairus-m . I looked at the code, tested it on some projects and made a few updates:
- we were not retrieving logs when checking errors, this was a small typo
- we should never retrieve the debug logs, even the truncated ones, they are way too big
- I added some logic to truncate the "normal" logs to the last 50 lines in case there is no `run_results.json`. It worked to get the LLM returning valid info when the DW creds are not valid or when we can't connect to the repo

jairus-m:
@b-per Awesome! Thanks for taking a final look and all the feedback over the last several weeks on this. 

As for:

> we were not retrieving logs when checking errors, this was a small typo

This was intentional as I was thinking that the combination of the "message", "unique_id", and "relation_name" were enough context but always returning the last 50 lines of the logs should be good.

Given the changes/updates, I do think I need to update the [prompt](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/prompts/admin_api/get_job_run_error.md) to reflect the behavior of:
1. Always returning logs
2. Full log is truncated to the last 50 lines

jairus-m:
Thank you @DevonFulcher and @b-per for the multiple rounds of review/feedback over the last several weeks. I know yall are so busy (especially with Coalesce coming up) so extra appreciate the time n effort. 🙌

PR #337, opened 2025-09-18
Add prereleases to the Create release PR action

### LINKED ISSUE #325: Add support for pre-release version
Currently, our release pipeline only allows patch, minor, and major releases. As we prepare for dbt MCP 1.0, we will need the ability to publish pre-release versions such that we can more easily & strictly follow semantic versioning.

b-per:
Closes #325 

Allow setting pre-releases when triggering the release action

The design is the following:
- this flow allows adding alpha.x/beta.x/rc.x pre-releases
- when pre-releases are created, they are added to the changelog but the changie files are also saved to a specific folder
- when the final release is done, changie collects all the pre-release files for this version, merge them all and remove the original prereleases from the changelog


b-per:
Updated info in CONTRIBUTING.md

Issue #251, opened 2025-07-25
Remove color from CLI tool output


DevonFulcher:
Currently, we are rendering output from CLI tools with color escape characters. For instance, here is a snippet from a the `show` tool output:
```
\x1b[0m00:37:13  Running with dbt=1.9.4\n\x1b[0m00:37:13  Registered adapter: snowflake=1.9.2\n\x1b[0m00:37:13
```
LLMs seem to be able to ignore these characters, but it would be better if we were able to remove them entirely, like this:
```
00:36:45  Running with dbt=1.9.4\n00:36:46  Registered adapter: snowflake=1.9.2\n00:36:46
```
This is possible with dbt-core with the `--no-use-colors` argument. However, this is different in Cloud CLI. It is called `--no-color`. So, we would have to detect the dbt executable type or make it configurable to make this work. I'm not sure how to turn off color in Fusion.

DevonFulcher:
Another, simpler option crossed my mind. Perhaps we shouldn't change the CLI parameters; instead, we could remove terminal escape characters using regex.

PR #343, opened 2025-09-22
Add env var to force no color flag

b-per:
When a person used the MCP server with the dbt Cloud CLI, for some reason the code didn't recognize that the executable was the dbt Cloud CLI and added the flag `--no-use-colors` instead of `--no-color` which raised issues.

I will dig further on why we didn't find the correct CLI type, but for now, the env var `DBT_COLOR_DISABLE_FLAG` could be used by anyone blocked by this behavior.

I don't think we need to document the env var `DBT_COLOR_DISABLE_FLAG` for now as it might be a temp workaround until I can understand the issue with recognizing the CLI type.

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20250922-151943.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20250922-151943.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Allow setting the "no color flag" when we can't detect the CLI type
+time: 2025-09-22T15:19:43.245974+02:00
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -60,6 +60,7 @@ class DbtCliConfig:
     dbt_path: str
     dbt_cli_timeout: int
     binary_type: BinaryType
+    dbt_no_color_flag: str | None = None
 
 
 @dataclass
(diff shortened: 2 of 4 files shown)


DevonFulcher:
@b-per I thought of another option that I mentioned [here](https://github.com/dbt-labs/dbt-mcp/issues/251#issuecomment-3259218849). It wouldn't rely on the type of executable being used.

b-per:
> @b-per I thought of another option that I mentioned [here](https://github.com/dbt-labs/dbt-mcp/issues/251#issuecomment-3259218849). It wouldn't rely on the type of executable being used.

I really really want to avoid regexes if possible 🤣 
I feel like this could open a whole new can of worms (what if people print logs with wrongly formatted colors as part of their code?)

DevonFulcher:
> I really really want to avoid regexes if possible 🤣 I feel like this could open a whole new can of worms (what if people print logs with wrongly formatted colors as part of their code?)

Ya, I hear you. The specific color or escape character shouldn't matter as long as we strip all of them. It is my understanding that escape characters are unique, such that a situation where they are printed unintentionally would be rare and would result in poorly formatted text in the terminal.

b-per:
Discussing internally to see if it is better to provide an env var to overwrite the CLI type when someone wants to provide one that is not the same as the one we auto-identify

b-per:
Closing for now as we have a new way of identifying what type a CLI is

PR #348, opened 2025-09-22
Create ConfigProvider ABC

DevonFulcher:
## Summary

This PR adds the `ConfigProvider` abstract base class, which all config providers inherit from. This allows us to create custom config providers for our internal usage.


bclifton (approved):
Thanks @DevonFulcher!

PR #368, opened 2025-09-30
Fix .user.yml error with Fusion

DevonFulcher:
## Summary

This fixes an issue when Fusion generates a `.user.yml` file in `<uuid>` format rather than the `id: <uuid>` format that dbt Core uses. I think this issue is being addressed in Fusion, but this should fix it in the meantime. I reproed the issue with `task client` and ensured this fix can handle the alternate format.

## Checklist

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -119,7 +119,12 @@ def load_config() -> Config:
     user_dir = get_dbt_profiles_path(settings.dbt_profiles_dir)
     user_yaml = try_read_yaml(user_dir / ".user.yml")
     if user_yaml:
-        local_user_id = user_yaml.get("id")
+        try:
+            local_user_id = user_yaml.get("id")
+        except Exception:
+            # dbt Fusion may have a different format for
+            # the .user.yml file which is handled here
+            local_user_id = str(user_yaml)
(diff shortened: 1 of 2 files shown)


src/dbt_mcp/config/config.py
```diff
@@ -119,7 +119,12 @@ def load_config() -> Config:
     user_dir = get_dbt_profiles_path(settings.dbt_profiles_dir)
     user_yaml = try_read_yaml(user_dir / ".user.yml")
     if user_yaml:
-        local_user_id = user_yaml.get("id")
+        try:
+            local_user_id = user_yaml.get("id")
+        except Exception:
+            # dbt Fusion may have a different format for
+            # the .user.yml file which is handled here
+            local_user_id = str(user_yaml)
```
reneruck-dbt, line 127: Do we want to check if it is a UUID before we use the file content as is?

DevonFulcher: dbt Core doesn't seem to do any validation, so I don't think we should either: https://dbt-labs.slack.com/archives/C044WU55X7Y/p1758830036930989?thread_ts=1758821476.003199&cid=C044WU55X7Y (internal to dbt Labs)

PR #373, opened 2025-09-30
automatically disable cloud tools that don't have proper setup.  

VDFaller:
## Summary
When the user doesn't give env vars for cloud, the tools for cloud still get registered.  This ends up causing problems when the client tries to use a tool that it definitely won't have access to.  

## What Changed
Automatically disabling

* Semantic Layer
* Admin Api
* Discovery

If the value for their respective environment variables is None.  

Still respects explicit setting of the variables.  

## Why
More tools isn't always better.  The client is getting confused.

## Checklist

## Additional Notes
I just used the three absolutely required values. But I can break it out better if that's needed.


src/dbt_mcp/config/settings.py
```diff
@@ -153,6 +173,12 @@ def parse_disable_tools(cls, env_var: str | None) -> list[ToolName]:
             raise ValueError("\n".join(errors))
         return tool_names
 
+    def _is_valid_dbt_cloud(self) -> bool:
+        return (
+            self.dbt_host is not None
+            and self.actual_prod_environment_id is not None
+            and self.dbt_token is not None
+        )
```
DevonFulcher: I wonder if we can consolidate this with the `validate_dbt_platform_settings` function.

DevonFulcher:
Thank you for raising this! The original intent of the configuration system was to error out if there was a misconfiguration. For instance, if a user didn't disable the dbt Platform toolsets, but also didn't set `DBT_HOST`, this would be a misconfiguration that would error out rather than disable the dbt Platform tools.

I think that UX has caused some issues, though, and users often struggle to find the errors in their MCP client to resolve the issue. Perhaps disabling misconfigured tools rather than erroring out is a better approach, as you have done in this PR.

I do have some overall comments.

As an alternative approach, would adding an env var like `DISABLE_DBT_PLATFORM` abate your concerns without changing how the config system works too much?

If we do decide to go with your original approach, we should make sure this applies to non-dbt Platform tools as well. For instance, if the user intends to use dbt CLI tools, but forgets to set `DBT_PROJECT_DIR`, then I would expect the dbt CLI tools to be disabled, rather than erroring out.

VDFaller:
@DevonFulcher so all the other tools already behave this way.  I was thinking that maybe I should consolidate everything to use the same system because it's a little disparate right now.  

I'd argue that erroring out because the environment is setup wrong is often a cause of frustration and feels like a barrier to entry.  Unless it had really good error messages.  In VSCode, at least, the mcp server errors feel a little hidden.  

I think `DISABLE_DBT_PLATFORM` isn't bad, but I still think that the less a user has to do the better in terms of UX.  

Ideally, I'd disable the tools and warn the user that tools are being disabled with good error messages how to remedy.  I haven't played with warnings inside of MCP stuff, but I'll try that and see how it feels today.

DevonFulcher:
@VDFaller sounds great, thank you!

VDFaller:
@DevonFulcher, <del>I think this is what I want reviewed. I didn't change how cli stuff works because I'm just going to make a separate issue for improving Misconfiguration messaging in general.  I'll try to put it in sometime later this week. </del>

*Edit* or if you just want me to hold off on this and do a more complete reworking of the misconfig stuff, I can do that.

DevonFulcher:
Hey @VDFaller, I just realized that this may conflict with my recent changes to enable oauth. Let's hold off on this for another week until the oauth option is released.

PR #381, opened 2025-10-02
Improved logging for development

DevonFulcher:
## Summary

This PR adds the ability to configure logging to a file. This can make it easier to debug certain clients that don't readily display MCP logs. I tested this in Claude Desktop.

## Checklist

### DIFF
--- a/src/dbt_mcp/telemetry/logging.py
+++ b/src/dbt_mcp/telemetry/logging.py
@@ -1,0 +1,46 @@
+from __future__ import annotations
+
+import logging
+import os
+from pathlib import Path
+
+FILE_LOGGING_ENV_VAR = "FILE_LOGGING"
+LOG_FILENAME = "dbt-mcp.log"
+
+
+def _find_repo_root() -> Path:
+    module_path = Path(__file__).resolve().parent
+    home = Path.home().resolve()
+    for candidate in [module_path, *module_path.parents]:
+        if (
+            (candidate / ".git").exists()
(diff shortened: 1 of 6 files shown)


src/dbt_mcp/telemetry/logging.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+import logging
+import os
+from pathlib import Path
+
+FILE_LOGGING_ENV_VAR = "FILE_LOGGING"
```
b-per: Is `FILE_LOGGING` a good name or should we make it something like `DBT_MCP_SERVER_FILE_LOGGING` so that 
- it is clear that it is for the dbt MCP server
- it doesn't interfere with people potentially already having `FILE_LOGGING` set

src/dbt_mcp/telemetry/logging.py
```diff
@@ -0,0 +1,46 @@
...
+            (candidate / ".git").exists()
+            or (candidate / "pyproject.toml").exists()
+            or candidate == home
+        ):
+            return candidate
+    return module_path
+
+
+def configure_file_logging() -> None:
+    if os.environ.get(FILE_LOGGING_ENV_VAR, "").lower() != "true":
```
b-per: Should we use `pydantic-settings` like we do for the config?

Here, we would need the value to be `True`/`true`/`TRUE` but with pydantic-settings we would also allow `t`, `on`, `1` etc...

PR #382, opened 2025-10-03
Include examples in prompt for entities/dimension metadata calls

WilliamDee:
## Summary

There were cases where the LLM passed an entire metric JSON object instead of just the metric name causing the tool to trip up. This PR adds a few examples to try to ensure the LLM just uses strictly names.

## Checklist


src/dbt_mcp/prompts/semantic_layer/get_dimensions.md
```diff
@@ -5,6 +5,26 @@ Dimensions are the attributes, features, or characteristics
...
+
+<example>
+Question: "Are there any time-related dimensions for my sales metrics?"
+Thinking step-by-step:
+   - The user is interested in time dimensions specifically
+   - I should use the search parameter to filter for dimensions with "time" in the name
+   - This will narrow down the results to just time-related dimensions
+Parameters:
+    metrics=["total_sales", "average_order_value"]
+    search="time"
```
DevonFulcher, line 29: Does `search` only work with one word? It may work better if this were possible: `search="time sales revenue"`.

WilliamDee: yea it's a partial string match and only works with one word can think of it as `*<search>*`

src/dbt_mcp/prompts/semantic_layer/get_entities.md
```diff
@@ -7,6 +7,26 @@ specific entities, such as customer churn or
 annual recurring revenue modeling.
 </instructions>
 
+<examples>
+<example>
+Question: "I want to analyze revenue - what entities are available?"
+Parameters:
+    metrics=["revenue"]
```
DevonFulcher, line 18: It might help if we add some language around knowing that `revenue` is a valid metric. If the agent doesn't know that, then it needs to call `list_metrics` first.

DevonFulcher: We can also add some thinking steps related to knowing that the user is asking for all of the entities, i.e. the search parameter should not be used.

DevonFulcher (approved):
Looks great!

PR #389, opened 2025-10-07
Write .user.yml if it does not exist

DevonFulcher:
## Summary

This PR generates .user.yml if it doesn't exist. This allows us to correlate our usage data with dbt Core & dbt Fusion. I tested this by running dbt MCP and seeing it generate this file. Here is what it looks like:

```
id: 0b9b5fe0-b7e2-42d6-827a-4aaff463e262
```

## Checklist


src/dbt_mcp/tracking/tracking.py
```diff
@@ -73,18 +76,25 @@ def _get_disabled_toolsets(self, settings: DbtMcpSettings) -> list[Toolset]:
...
                 except Exception:
                     # dbt Fusion may have a different format for
                     # the .user.yml file which is handled here
                     self._local_user_id = str(user_yaml)
+            else:
+                self._local_user_id = str(uuid.uuid4())
+                with suppress(Exception):
+                    Path(user_yaml_path).write_text(
+                        yaml.dump({"id": self._local_user_id})
+                    )
```
reneruck-dbt, line 97: In case `.user.yml` exists but can't be parsed as yaml, we override it with a new uuid?
What is writing that file originally and what effect does that have on it?

DevonFulcher: It should always be parseable as yaml. The format is `id: <uuid>`. There was an issue with the format being `<uuid>`, but I think that is being addressed. dbt Core & dbt Fusion can also read & write this file. This is where we discussed this: https://dbt-labs.slack.com/archives/C044WU55X7Y/p1758821476003199 (internal to dbt Labs)

gshank (approved):

src/dbt_mcp/tracking/tracking.py
reneruck-dbt: sounds good.

PR #394, opened 2025-10-08
Avoid double counting in usage tracking proxied tools

DevonFulcher:
## Summary

We are now counting remote tool calls on our backend, so we shouldn't double-count them here as well.

## Checklist

### DIFF
--- a/src/dbt_mcp/tools/toolsets.py
+++ b/src/dbt_mcp/tools/toolsets.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from typing import Literal
 
 from dbt_mcp.tools.tool_names import ToolName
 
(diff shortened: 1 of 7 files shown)


src/dbt_mcp/tools/toolsets.py
```diff
@@ -12,6 +13,13 @@ class Toolset(Enum):
     DBT_CODEGEN = "dbt_codegen"
 
 
+proxied_tools: set[Literal[ToolName.TEXT_TO_SQL, ToolName.EXECUTE_SQL]] = set(
+    [
+        ToolName.TEXT_TO_SQL,
+        ToolName.EXECUTE_SQL,
```
reneruck-dbt, line 19: We probably want to add the lsp/fusion tools here as well.

DevonFulcher: Yes, when we start proxying the remote Fusion tools, we should add them here.

PR #397, opened 2025-10-09
LSP Tools Integration - Part 1: Connection and Communication Layer

reneruck-dbt:
## Summary
This is the first PR in a series (lsp-tools-1 → lsp-tools-4) implementing dbt Fusion LSP integration into the MCP server.

Implements the core LSP connection management and JSON-RPC communication layer:
- LspConnection class (src/dbt_mcp/lsp/lsp_connection.py:687): Manages LSP process lifecycle, handles JSON-RPC message formatting, and implements
  request/response/notification patterns per the Language Server Protocol specification
- Event handling: Support for dbt-specific LSP events (compileComplete, diagnostics, progress updates)
- Comprehensive test coverage:
    - Unit tests with mocked subprocess communication
    - Integration tests against real sockets

## What Changed

- Handles stdio-based LSP communication with proper message framing (Content-Length headers)
- Async request/response correlation using unique message IDs
- Graceful process lifecycle management with cleanup
- Socket-based availability checking for LSP server readiness

## Related Issues

- lsp-tools-2: LSP client implementation
- lsp-tools-3: LSP tool definitions and configuration
- lsp-tools-4: Wire LSP tools into MCP server

## Checklist

### DIFF
--- a/src/dbt_mcp/lsp/lsp_connection.py
+++ b/src/dbt_mcp/lsp/lsp_connection.py
@@ -491,7 +491,9 @@ class LSPConnection:
             else:
                 # it's an unknown request, we respond with an empty result
                 logger.debug(f"LSP request {message.to_dict()}")
-                self._send_message(JsonRpcMessage(id=message.id, result=None))
+                self._send_message(
+                    JsonRpcMessage(id=message.id, result=None), none_values=True
+                )
 
         if message.method is None:
             return
@@ -604,10 +606,10 @@ class LSPConnection:
 
         return future
 
-    def _send_message(self, message: JsonRpcMessage) -> None:
+    def _send_message(self, message: JsonRpcMessage, none_values: bool = False) -> None:
         """Send a message to the LSP server."""
         # Serialize message
-        content = json.dumps(message.to_dict())
+        content = json.dumps(message.to_dict(none_values=none_values))
         content_bytes = content.encode("utf-8")
 
         # Create LSP message with headers


src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,687 @@
...
+        logger.debug(f"Starting LSP server: {' '.join(cmd)}")
+        if self.cwd:
+            logger.debug(f"Working directory: {self.cwd}")
+
+        # Start the process with stdout/stderr capture
+        self.process = await asyncio.create_subprocess_exec(
+            str(self.binary_path),
+            "--socket",
+            f"{self.port}",
+            "--project-dir",
```
alan-andrade, line 213: should we use the `cmd` built right before this?

src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,687 @@
...
+                        future.set_exception,
+                        RuntimeError(f"LSP error: {message.error}"),
+                    )
+                else:
+                    future_loop.call_soon_threadsafe(future.set_result, message.result)
+                return
+            else:
+                # it's an unknown request, we respond with an empty result
+                logger.debug(f"LSP request {message.to_dict()}")
+                self._send_message(jsonrpc({"id": message.id, "result": None}))
```
alan-andrade: Should this be using `JsonRpcMessage` instead?

reneruck-dbt: JsonRpcMessage explicitly does not serialize fields with None values but we need to return `result` as response to an LSP request.
We can combine jsonrpc and JsonRpcMessage, I'll think about how.

src/dbt_mcp/lsp/lsp_connection.py
reneruck-dbt: good catch, they overlap.

src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,676 @@
...
+        if self.method is not None or none_values:
+            msg["method"] = self.method
+        if self.params is not None or none_values:
+            msg["params"] = self.params
+        if self.result is not None or none_values:
+            msg["result"] = self.result
+        if self.error is not None or none_values:
+            msg["error"] = self.error
+
+        return msg
```
DevonFulcher: There is a built in way of doing this with dataclasses
```
from dataclasses import asdict
```

src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,676 @@
...
+    def from_dict(cls, data: dict[str, Any]) -> "JsonRpcMessage":
+        """Create message from dictionary."""
+        return cls(
+            jsonrpc=data.get("jsonrpc", "2.0"),
+            id=data.get("id"),
+            method=data.get("method"),
+            params=data.get("params"),
+            result=data.get("result"),
+            error=data.get("error"),
+        )
```
DevonFulcher: Likewise this is built-in:
```
JsonRpcMessage(**mydict)
```

src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,676 @@
...
+    shutting_down: bool = False
+    capabilities: dict[str, Any] = field(default_factory=dict)
+    next_request_id: int = 1
+    pending_requests: dict[int | str, asyncio.Future] = field(default_factory=dict)
+    pending_notifications: dict[LspEventName, list[asyncio.Future]] = field(
+        default_factory=dict
+    )
+    compiled: bool = False
+    # start at 20 to avoid collisions between ids of requests we are waiting for and the lsp server requests from us
+    request_id_counter: Iterator[int] = itertools.count(20)
```
DevonFulcher: I'm still not understanding this. Do you mind elaborating a bit more about the magic number `20`?

src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,676 @@
...
+    - Error handling and cleanup
+    """
+
+    def __init__(
+        self,
+        binary_path: str,
+        cwd: str,
+        args: Sequence[str] | None = None,
+        connection_timeout: float = 10,
+        default_request_timeout: float = 30,
```
DevonFulcher: `DEFAULT_DBT_CLI_TIMEOUT` is set to 60. Maybe this should also be set to 60. Some dbt interactions can take some time.

Non-blocking: perhaps we should have a configurable global tool timeout rather than just for the CLI tools.

src/dbt_mcp/lsp/lsp_connection.py
```diff
@@ -0,0 +1,676 @@
...
+            str(self.binary_path),
+            "--socket",
+            f"{self.port}",
+            "--project-dir",
+            self.cwd,
+            *self.args,
+        ]
+
+        logger.debug(f"Starting LSP server: {' '.join(cmd)}")
+        self.process = await asyncio.create_subprocess_exec(*cmd)
```
DevonFulcher, line 219: Do we want to merge stdout and stderr like we are doing [here](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/dbt_cli/tools.py#L70-L72)? Although, I'm not sure if these subprocesses serve the same purpose.

sandbox.ipynb
DevonFulcher, line 1: Can you explain this file?

reneruck-dbt: oh that slipped in by accident, will remove it

src/dbt_mcp/lsp/lsp_connection.py
reneruck-dbt: lsp logs are pretty noisy, not sure if want that.

src/dbt_mcp/lsp/lsp_connection.py
reneruck-dbt: the configurable timeout makes sense

src/dbt_mcp/lsp/lsp_connection.py
reneruck-dbt: both the lsp client and server keep a counter for requests they are sending.
This allows them to match responses to request since lsp works off two socket connections, one for sending and one for receiving.
Say we send a request with id 4 and and remember in `pending_requests` a response with id 4 should resolve the attached future.
Now at the same time the lsp sends us a request with id 4.
In that case we mistakenly assume the message with id 4 just received is the response to the request we just sent.
We do resolve the future and move on. The lsp however is waiting for a response to their request 4 and will not accept further requests until that one has been resolved. (I'm not sure if that is due to how the fusion lsp is implemented or part of the protocol)

One challenge is to disambiguate responses from request, given that `result` can be None.
In aca we do blindly send back a response with result None request or not for every message we do not explicitly handle.
20 is an arbitrary number.
I'm sure there is an elegant way to solve this.

src/dbt_mcp/lsp/lsp_connection.py
DevonFulcher: Got it. If it isn't helpful for an agent, then we shouldn't include it.

src/dbt_mcp/lsp/lsp_connection.py
reneruck-dbt: you are right, forrest for the trees 🌲 🌴 🌳 😆

b-per:
I tried to run the MCP server locally from this branch but I don't get the new tools showing up.

This is my `.env` file:
```
DISABLE_SEMANTIC_LAYER=True
DISABLE_DISCOVERY=True
DISABLE_ADMIN_API=True
DISABLE_SQL=True
DISABLE_DBT_CODEGEN=False

DISABLE_LSP=False

DBT_PROJECT_DIR=<mypath>
DBT_PATH=<mypath>
```

I get the CLI and codegen tools showing up but not the LSP ones

b-per:
I am confused now. I ran it with `uv` on top of a local git repo and it worked. 
But what didn't seem to work at first was to run it with `uvx` and pointing to the git branch.

I will check more

PR #399, opened 2025-10-09
LSP Tools Integration - Part 3: MCP Tool Definitions and Configuration

reneruck-dbt:
## Summary
This is the third PR in the LSP integration series (lsp-tools-1 → lsp-tools-4), defining the MCP tools that expose LSP functionality to AI agents.

## What Changed
Defines the MCP tool layer and configuration for LSP integration:

  - MCP Tool Registration (src/dbt_mcp/lsp/tools.py:182):
    - get_column_lineage tool: Traces column-level lineage from source to target
    - Connection lifecycle management with module-level singleton pattern
    - Automatic LSP client initialization and error handling
    - Decorator pattern to inject LSP client into tool functions while hiding it from MCP signatures
  - Configuration Integration (src/dbt_mcp/config/config.py):
    - New LspConfig dataclass for LSP settings
    - Auto-detection of LSP availability based on project directory
    - Optional LSP path override via settings
  - AI-Friendly Documentation (src/dbt_mcp/prompts/lsp/):
    - Tool descriptions explaining when and how to use column lineage
    - Parameter descriptions for model IDs and column names
    - Helps AI agents understand the tool's purpose and usage patterns

  **How it builds on previous PRs**

  - Uses LSPConnection (part 1) and LSPClient (part 2) under the hood
  - Wraps async LSP operations in MCP-compatible tool functions
  - Manages connection lifecycle transparently for tool callers
 
## Related Issues
- lsp-tools-4: Wire LSP tools into the main MCP server and enable in production

## Checklist

### DIFF
--- a/src/dbt_mcp/lsp/tools.py
+++ b/src/dbt_mcp/lsp/tools.py
@@ -100,10 +100,18 @@ async def list_lsp_tools(config: LspConfig) -> list[ToolDefinition]:
 
         return wrapper
 
-    for tool_definition in LSP_TOOL_DEFINITIONS:
-        tool_definition.fn = call_with_lsp_client(tool_definition.fn)
-
-    return LSP_TOOL_DEFINITIONS
+    return [
+        ToolDefinition(
+            fn=call_with_lsp_client(get_column_lineage),
+            description=get_prompt("lsp/get_column_lineage"),
+            annotations=create_tool_annotations(
+                title="get_column_lineage",
+                read_only_hint=False,
+                destructive_hint=False,
+                idempotent_hint=True,
+            ),
+        ),
+    ]
 
 
 async def get_column_lineage(
@@ -166,17 +174,3 @@ async def cleanup_lsp_connection() -> None:
             logger.error(f"Error cleaning up LSP connection: {e}")
         finally:
             _lsp_connection = None
-
-
-LSP_TOOL_DEFINITIONS = [
-    ToolDefinition(
-        fn=get_column_lineage,
-        description=get_prompt("lsp/get_column_lineage"),
-        annotations=create_tool_annotations(
-            title="get_column_lineage",
-            read_only_hint=False,
-            destructive_hint=False,
-            idempotent_hint=True,
-        ),
-    ),
-]


src/dbt_mcp/prompts/lsp/get_column_lineage.md
DevonFulcher, line 1: Nice prompt!

src/dbt_mcp/lsp/tools.py
```diff
@@ -0,0 +1,182 @@
...
+async def list_lsp_tools(config: LspConfig) -> list[ToolDefinition]:
+    """Register dbt Fusion tools with the MCP server.
+
+    Args:
+        config: LSP configuration containing LSP settings
+
+    Returns:
+        List of tool definitions for LSP tools
+    """
+    global _lsp_connection
```
DevonFulcher, line 47: I think this will be problematic when we try hosting these tools in our infra.

src/dbt_mcp/lsp/tools.py
```diff
@@ -0,0 +1,182 @@
...
+                if param.name != "lsp_client"
+            ]
+        )
+
+        return wrapper
+
+    for tool_definition in LSP_TOOL_DEFINITIONS:
+        tool_definition.fn = call_with_lsp_client(tool_definition.fn)
+
+    return LSP_TOOL_DEFINITIONS
```
DevonFulcher: This seems a bit atypical. Should we just return the tool definitions list with `fn` rather than have a global variable and update `fn`?

src/dbt_mcp/lsp/tools.py
```diff
@@ -0,0 +1,182 @@
...
+            return await func(lsp_client, *args, **kwargs)
+
+        # remove the lsp_client argument from the signature
+        wrapper.__signature__ = inspect.signature(func).replace(  # type: ignore
+            parameters=[
+                param
+                for param in inspect.signature(func).parameters.values()
+                if param.name != "lsp_client"
+            ]
+        )
```
DevonFulcher, line 99: This is also a bit atypical. If `get_column_lineage` were defined within `call_with_lsp_client` then it could access `lsp_client` without passing it in as a parameter.

DevonFulcher: We could follow the same pattern as [this](https://github.com/dbt-labs/dbt-mcp/blob/7db3b24b1030420f92df7c70ddff8da335cbc485/src/dbt_mcp/semantic_layer/tools.py#L32) function where `semantic_layer_fetcher` is provided to the tools, but isn't passed in as a parameter. I think your approach here is fine, but it is just inconsistent with how other tool definitions are created. Let me know if there is a good reason for it to be different.

src/dbt_mcp/lsp/tools.py
reneruck-dbt: I'm having a lspClient provider in the works to abstract that away, besides there will need to be more abstraction required before we use it internally.

src/dbt_mcp/lsp/tools.py
reneruck-dbt: you are right that might be cleaner

src/dbt_mcp/lsp/tools.py
reneruck-dbt: The catch with the semantic layer tools is they are not independent from create_sl_tool_definitions and are harder to unit test individually.
The reason we are stripping out the lspClient from the signature is not only so the agent doesn't see but also for MCP itself.
Under the hood it's doing some pydantic magic that serializes the function signature to json and chokes on the non primitive type LspClient.

PR #403, opened 2025-10-13
Implement get_all_sources  Tool

### LINKED ISSUE #336: Add Support for Source Table Metadata in MCP
**Description:**

A PoC customer has requested the ability to retrieve source table metadata through the Metadata Communication Protocol (MCP). While this information is currently accessible through the Discovery API, it appears MCP does not currently expose it.

The use case involves teams wanting to inspect source metadata alongside models. At the moment, the `get_all_models` method does not include source tables, which makes it harder to get a complete picture of the data graph via MCP.

As discussed internaly, a new MCP method such as `get_all_sources` could be introduced to align with the Discovery API structure and avoid overloading get_all_models.

**Proposed enhancement:**
	•	Add a get_all_sources method to MCP that retrieves metadata for all source tables.
	•	Structure the response similarly to the existing Discovery API source table metadata format.

Muizzkolapo:
## Summary

This PR is in collaboration with @jairus-m . The changes adds support for Source Table Metadata in the dbt mcp. 

Filter discovery:

Using the dbt Cloud GraphQL playground to introspect the actual schema:
image

### Available Filter Options
From the schema introspection, we found 8 verified filter fields:
- `database` - Filter by database name
- `freshnessChecked` - Boolean for freshness check status
- `freshnessStatus` - Enum: "pass", "warn", "error"
- `identifier` - Filter by table identifier
- `schema` - Filter by schema name
- `sourceNames` - Array of source names (corrected from `sourceName`)
- `tags` - Array of tags
- `uniqueIds` - Array of unique identifiers

Rather than implementing all 8 fields, we chose the most commonly used ones:
- sourceNames - Most common use case (filter by source)
- uniqueIds - For precise targeting

image

### Core Implementation
- **Added GraphQL Query**: Created `GET_SOURCES` query in `GraphQLQueries` class to fetch source table metadata with filtering and pagination support
- **New SourceFilter Type**: Defined `SourceFilter` TypedDict with 2 filter fields (`sourceNames`, `uniqueIds`)
- **SourcesFetcher Class**: Implemented new fetcher class with pagination support, following the same pattern as existing `ModelsFetcher` and `ExposuresFetcher`
- **Tool Integration**: Added `get_all_sources` tool to discovery tools with proper annotations and documentation
- **Enhanced Lineage**: Updated `GET_MODEL_LINEAGE_QUERY` to include `sourceName` and `uniqueId` fields for source nodes

## Why

Based on the details here #336 https://github.com/dbt-labs/dbt-mcp/issues/336

This enhancement addresses the gap by providing:

1. **Complete Data Graph Visibility**: Teams can now get a full picture of their data infrastructure including upstream sources
2. **MCP-Discovery API Alignment**: Maintains consistency with existing Discovery API source metadata format
3. **Source-Specific Workflows**: Enables dedicated source inspection and monitoring workflows without cluttering model queries

## Related Issues

Closes #336

## Checklist
- [ ] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required https://github.com/dbt-labs/docs.getdbt.com/pull/8036-- Mention it here

### DIFF
--- a/.changes/unreleased/Bug Fix-20251028-143835.yaml
+++ b/.changes/unreleased/Bug Fix-20251028-143835.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Minor update to the instruction for LSP tool
+time: 2025-10-28T14:38:35.729371+01:00
--- a/src/dbt_mcp/prompts/lsp/args/model_id.md
+++ b/src/dbt_mcp/prompts/lsp/args/model_id.md
@@ -1,1 +1,3 @@
-The fully qualified dbt model identifier (e.g., "model.my_project.my_model").
+The fully qualified dbt model identifier (e.g., "model.my_project_name.my_model").
+
+Do not use just the model name, always use the model full identifier. If you don't have the full identifier, the project name is the field `name:` in the `dbt_project.yml` file at the root of the dbt project.


src/dbt_mcp/discovery/tools.py
```diff
@@ -61,6 +62,9 @@ async def get_exposure_details(
     ) -> list[dict]:
         return await exposures_fetcher.fetch_exposure_details(exposure_name, unique_ids)
 
+    async def get_all_sources(source_filter: dict | None = None) -> list[dict]:
+        return await sources_fetcher.fetch_sources(source_filter)
+
```
jairus-m: I see that there's currently 6 possible "keys" that can be passed into the `source_filter` dict param. 

One thing I'm thinking about, is that this may be too vague for an LLM. It may be worth considering to have named params instead with stronger typing?

i.e doing something like this:

```python
  async def get_all_sources(
      source_names: list[str] | None = None,
      unique_ids: list[str] | None = None,
      database: str | None = None,
      schema: str | None = None,
      freshness_status: str | None = None,
      tags: list[str] | None = None,
  ) -> list[dict]:
``` 

and then building out a `source_filter` dict within `get_all_sources` to then pass to the LLM.

The biggest reason why i think this may be worth  considering , is mainly for better "discoverability" for the LLM and it will help make it clear for it how to use this tool!

tests/unit/discovery/test_sources_fetcher.py
```diff
@@ -0,0 +1,476 @@
...
+    mock_config = Mock()
+    mock_config.environment_id = 123
+
+    # Make get_config async
+    async def mock_get_config():
+        return mock_config
+
+    mock_config_provider.get_config = mock_get_config
+    mock_client.config_provider = mock_config_provider
+    return mock_client
```
jairus-m: This fixture also exists in `tests/unit/discovery/test_exposures_fetcher.py` - can it be moved to something like `tests/unit/discovery/conftest.py` for reuse between the two modules?

tests/unit/discovery/test_sources_fetcher.py
```diff
@@ -0,0 +1,476 @@
...
+    assert result[0]["name"] == "customers"
+    assert result[0]["sourceName"] == "raw_data"
+    assert result[0]["resourceType"] == "source"
+    assert result[0]["freshness"]["freshnessStatus"] == "pass"
+    assert result[1]["name"] == "orders"
+    assert result[1]["freshness"]["freshnessStatus"] == "warn"
+
+
+async def test_fetch_sources_with_filter(sources_fetcher, mock_api_client):
+    mock_response = {
```
jairus-m, line 113: This test specifically tests the mechanics of the `SourceFetcher` 's `fetch_source()` method right? 

Would this make some downstream tests a bit redundant? 
- test_fetch_sources_with_unique_ids_filter
- test_fetch_sources_with_tags_filter
- test_fetch_sources_with_combined_filters
- test_fetch_sources_with_database_filter
- test_fetch_sources_with_schema_filter

They all kinda look to test the same general mechanism/code path of taking in `source_filter` dict param, then passing that to `fetch_sources` method, and then validating the return result. 

While it does feel like these tests would all fail for similar reasons, maybe it is worth keeping them? It just feels like there's a lot of extra lines/tests that may not add the same amount of value.

jairus-m (commented):
Great work!! 🔥

I've only given this a preliminary look and left some comments. I haven't tested with the MCP inspector / or with an LLM yet (will do so soon). I also I didn't look too heavily at the integration testing side ,but may try and find time to better understand.

tests/unit/discovery/test_sources_fetcher.py
Muizzkolapo: I think I will review some of the tests they were from an initial approach I took when implementing, so I agree some might be redundant now. 

Thats right, they test the same mechanism, I have used @pytest.mark.parametrize  consolidate them to reduce the noise while we can still maintain the same test coverage.

src/dbt_mcp/discovery/tools.py
Muizzkolapo: Good catch, thats a better idea tbf. I have come across scenarios from other mcp tools  where it just kept calling the tool with the wrong type till it timed out, this would prolly help gate the tool better.

tests/unit/discovery/test_sources_fetcher.py
Muizzkolapo: This is a good idea Jairus, will move it as well and see what people think.

jairus-m:
@Muizzkolapo thanks for addressing the comments! I'm gonna cut out some time tomorrow to test this out with the MCP inspector and an LLM.

EDIT: Have you also been able to test and run this tool successfully on your end?

src/dbt_mcp/discovery/tools.py
```diff
@@ -61,6 +62,18 @@ async def get_exposure_details(
...
+        unique_ids: list[str] | None = None,
+        database: str | None = None,
+        schema: str | None = None,
+        freshness_status: str | None = None,
+        tags: list[str] | None = None,
+    ) -> list[dict]:
+        return await sources_fetcher.fetch_sources(
+            source_names, unique_ids, database, schema, freshness_status, tags
+        )
+
```
jairus-m, line 75: As I'm starting to test in the MCP inspector, as a user, it's almost a bit overwhelming for what filters to use (and not to mention the combinatorial increase in options available to me and LLMs). 

My thought is - maybe you can keep those other options within the `SourcesFetcher` class , but for the tool itself, to just keep:

- `source_names`
- `unique_ids`

I do understand that the original issue was asking for the ability to get ALL source object metadata, but for bigger projects, that may overload context (?). So I do think filtering is a good call here, but it may be worth augmenting the approach to be more similar to the other Discovery API tools such that you have `get_sources` with the two params mentioned above and if nothing is supplied, return all sources as originally outlined. 

Let me know what you think!

Muizzkolapo: That’s a valid concern, actually. A good design for better UX might be to try out elicitation and prompt the user for more filters if they have them or a form of progressive disclosure. We can keep it simple for now, but I’m thinking about an approach that considers both the context overload by providing more filters and requires less input from users. I’ll let you know what I come up with and we can write a new issue for that.

Muizzkolapo: I have marked into draft for now till we are satisfied with the approach.

Muizzkolapo:
> @Muizzkolapo thanks for addressing the comments! I'm gonna cut out some time tomorrow to test this out with the MCP inspector and an LLM.
> 
> EDIT: Have you also been able to test and run this tool successfully on your end?

yeah ran via google ADK and mcp inspector.

jairus-m:
Hey @Muizzkolapo - got to testing your branch in-depth and specifically on the UX and calling lineage between sources/models. I have a suite of changes that improve the specific workflows/prompts that get to the heart of the original issue being raised.

It may be easier if I just commit the changes but do recognize that you may not want me to do that. Would you be okay if I added some work with documentation to point out the suggestions? 

You can always revert if you don't find the changes acceptable!

Muizzkolapo:
> Hey @Muizzkolapo - got to testing your branch in-depth and specifically on the UX and calling lineage between sources/models. I have a suite of changes that improve the specific workflows/prompts that get to the heart of the original issue being raised.
> 
> It may be easier if I just commit the changes but do recognize that you may not want me to do that. Would you be okay if I added some work with documentation to point out the suggestions?
> 
> You can always revert if you don't find the changes acceptable!

Nice! that’s  really cool! Go for it and I can test it out.

jairus-m:
@Muizzkolapo I was about to push my changes but realized I don't have write access to your forked repo/branch. Let me know if you can [add me as a collaborator](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/repository-access-and-collaboration/inviting-collaborators-to-a-personal-repository) - if not, totally cool! I can figure it out on Sunday.

jairus-m:
About to push some commits but here is the gist! The changes can generally be summarized in 3 parts:

Alignment to discovery API and top-level alignment with dbt Platform source view
 - `source_names` and `unique_ids` filter params aligns with existing discovery tool set
   - you did this already but just want to re-state!
 - `source_names` filter mirrors the top-level grouping/filtering of the dbt Platform catalog when you look at the sources
   - in my head, this makes it easier for users to generally start filtering by if `get_all_sources` overloads context while still being able to target specific sources nodes via `unique_ids`
- Included identifier, database, and schema fields to the returned metadata of `get_all_sources`
  - metadata matches what other discovery tools return and provides additional/relevant context for source nodes

Lineage between models and their source parents
- Currently, connecting a model to its source parent and then getting that additional source metadata is not obviously possible through the MCP tools (?) as stated by the issue. 
  - when you call `get_model_parents` and it includes a source node, there's no way to use the existing metadata data with `get_all_sources` for further inspection
- To address this, we can can add `sourceName `and `uniqueId` as part of the returned metadata when you call `get_model_parents` for a node that has a source parent. 
  - since `source_name` and `unique_id` , that can be used to further link the parent lineage and get further source metadata through `get_all_sources`

Housekeeping :D
- Added `get_all_sources` to tool policy and toolsets modules
- Updated prompts
- Ran ruff format
- Merged main

Muizzkolapo:
> About to push some commits but here is the gist! The changes can generally be summarized in 3 parts:
> 
> Alignment to discovery API and top-level alignment with dbt Platform source view
> 
> * `source_names` and `unique_ids` filter params aligns with existing discovery tool set
>   
>   * you did this already but just want to re-state!
> * `source_names` filter mirrors the top-level grouping/filtering of the dbt Platform catalog when you look at the sources
>   
>   * in my head, this makes it easier for users to generally start filtering by if `get_all_sources` overloads context while still being able to target specific sources nodes via `unique_ids`
> * Included identifier, database, and schema fields to the returned metadata of `get_all_sources`
>   
>   * metadata matches what other discovery tools return and provides additional/relevant context for source nodes
> 
> Lineage between models and their source parents
> 
> * Currently, connecting a model to its source parent and then getting that additional source metadata is not obviously possible through the MCP tools (?) as stated by the issue.
>   
>   * when you call `get_model_parents` and it includes a source node, there's no way to use the existing metadata data with `get_all_sources` for further inspection
> * To address this, we can can add `sourceName `and `uniqueId` as part of the returned metadata when you call `get_model_parents` for a node that has a source parent.
>   
>   * since `source_name` and `unique_id` , that can be used to further link the parent lineage and get further source metadata through `get_all_sources`
> 
> Housekeeping :D
> 
> * Added `get_all_sources` to tool policy and toolsets modules
> * Updated prompts
> * Ran ruff format
> * Merged main

Nice one @jairus-m  - I am just gonna run some tests as well using inspector and one of the agent sdks and will revert with my thoughts!

Muizzkolapo:
@jairus-m  Looking good just completed testing with mcp inspector, I am going to use the google SDK example to test it later.

src/dbt_mcp/discovery/client.py
```diff
@@ -575,3 +619,65 @@ async def fetch_exposure_details(
...
+        if unique_ids is not None:
+            source_filter["uniqueIds"] = unique_ids
+
+        has_next_page = True
+        after_cursor: str = ""
+        all_edges: list[dict] = []
+
+        while (
+            has_next_page and len(all_edges) < MAX_NUM_MODELS
+        ):  # Reuse MAX_NUM_MODELS limit
```
DevonFulcher: Non-blocking: We could probably just rename this variable, but that is okay. Thanks for the comment.

src/dbt_mcp/discovery/client.py
```diff
@@ -575,3 +619,65 @@ async def fetch_exposure_details(
...
+    async def fetch_sources(
+        self,
+        source_names: list[str] | None = None,
+        unique_ids: list[str] | None = None,
+    ) -> list[dict]:
+        source_filter: SourceFilter = {}
+        if source_names is not None:
+            source_filter["sourceNames"] = source_names
+        if unique_ids is not None:
+            source_filter["uniqueIds"] = unique_ids
```
DevonFulcher, line 658: I think we want this to be implemented like `_get_model_filters`, such that if both `source_names` and `unique_ids` are None, we raise an error.

DevonFulcher: Oh, actually never mind. I see that these should be optional.

src/dbt_mcp/discovery/client.py
jairus-m: @Muizzkolapo how do you feel about using `MAX_NODE_QUERY_LIMIT` ?

happy to make that change but let me know if you have thoughts or want to make that update!

DevonFulcher: That sounds good! Feel free to make that update if you have the time. Thank you!

Muizzkolapo: Sounds good to me! Thanks @jairus-m

jairus-m: pushed! thanks yall.

DevonFulcher:
I should add that I have tested this myself, and it works as expected.

PR #414, opened 2025-10-28
Update instructions LSP tool

b-per:
Very minor update to the LSP tool instructions for getting column level lineage.
I couldn't get it to works at first because it was just using the model name instead of its ID.

I am suggesting to get the project name from `dbt_project.yml` but I am open to other suggestions.

- we could also get it from running the list tool but in that case we would need the list tool to accept the JSON output so that we can request the `unique_id`
- or we could accept a model name in addition to its unique ID and build the unique ID ourselves in the logic of the tool


DevonFulcher:
Actually, perhaps tools should accept just unique IDs instead, because we plan to make the server multi-project. We will have to think through that more.

b-per:
Thanks. Yes, we can revisit, but for now this PR improved the tool on my side.

PR #415, opened 2025-10-28
Check that host doesn't contain prefix

b-per:
## Summary

Add a check to provide an error message if the host value contains the mult-icell prefix

## What Changed

Extra check in the config

## Why

People got confused in how to set the host sometimes which led to the MCP server starting correctly but then not being able to connect to the dbt platform


b-per:
After discussion this likely is not relevant for some use cases (e.g. ST), so let's not implement this

Issue #418, opened 2025-10-28
[Feature]: Explicit allowlist for tools and toolsets


DevonFulcher:
### Is your feature request related to a problem?

Currently, tool configuration is based on a disallow list system. Almost all tools are enabled by default, and they have to be disabled with environment variables (for local MCP) and headers (for remote MCP). This is problematic when we add new tools or toolsets that are enabled by default, as they become available to agents even if the user doesn't intend to use them. This is particularly problematic when users design handcrafted production agents that could change behavior unpredictably.

### Describe the solution you'd like

We should have an explicit allowlist system that supersedes tools and toolsets that are on by default. For instance, we should have environment variables of this form: `DBT_MCP_ENABLE_SEMANTIC_LAYER` & `DBT_MCP_ENABLE_TOOLS` and headers like `x-dbt-enable-tools` & `x-dbt-enable-toolsets`. When these configurations are present, all tools are disabled except for the specific ones, which are enabled.

We intend to make our configuration system as unambiguously lenient as possible (https://github.com/dbt-labs/dbt-mcp/issues/380). So, if users set these new configuration options and the pre-existing disabled options, the disabled options will simply be ignored. The enable options take precedence over the disable options. Let me know if there are concerns with this approach.

### Describe alternatives you've considered

We considered making all new tools & toolsets disabled by default, but determined that this would be too divergent from how the system currently works. This approach would result in an artificial bifurcation of tools into those implemented before and after this change. So, we decided against this approach.

xrpza:
I can take a stab at this one @DevonFulcher , if you'd like the help.

VDFaller:
If you used a [ruff like system](https://docs.astral.sh/ruff/settings/#lint_ignore), You could enable all semantic tools, but more specific disabling would take precedence (or vice versa).  If you default `None` for those headers/environment variables to be true, it might solve both problems.  I just feel if I explicitly enable and disable something, it should be disabled.  At the very least warn that there's a tie.

DevonFulcher:
Yes, I hear you @VDFaller. I think I want to take back this statement:

> So, if users set these new configuration options and the pre-existing disabled options, the disabled options will simply be ignored.

I think this should be the precedence:

1. Disable individual tools with `DISABLE_TOOLS` and `x-dbt-disable-tools`
2. Enable individual tools with `DBT_MCP_ENABLE_TOOLS` and `x-dbt-enable-tools`
[...]

PR #420, opened 2025-10-29
Add `get_model_lineage_dev` CLI tool

VDFaller:
## Summary
Added a tool for parsing the manifest to get the lineage.  

## What Changed
Just parses then reads the manifest.  

## Why
So that it's usable by non-cloud customers. 

## Checklist

## Additional Notes
There are some differences in the outputs.  I can try to line them up if needed. 

## Prompt Call & Response
* "Get me the lineage of this model"
    * Bad - doesn't use mcp
* "use the dbt mcp server to tell me the full lineage of this model"
    * Good - calls both directions, recursively
* "use the dbt mcp server to tell me the lineage of this model"
    * Bad - ran `jq` on the manifest
    * 2nd run - Fine - calls both directions recursively
* "use the dbt mcp server to get me the children of this model, including tests"
    *  Good - Gets non-recursive children (used the name not the unique_id) and had the tests included.


src/dbt_mcp/discovery/models/lineage_types.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class Ancestor(BaseModel):
+    model_id: str
+    parents: list[Ancestor] = Field(default_factory=list)
+
+
+class ModelLineage(BaseModel):
+    model_id: str
+    ancestors: list[Ancestor] | None = Field(default_factory=list)
```
VDFaller: Should maybe change this to parents and children to be consistent?

src/dbt_mcp/discovery/models/lineage_types.py
```diff
@@ -0,0 +1,132 @@
...
+    descendants: list[Descendant] | None = Field(default_factory=list)
+
+    @field_validator("descendants", mode="after")
+    @classmethod
+    def _sanitize_descendants(cls, value: list[Descendant] | None) -> list[Descendant]:  # noqa: D401 - simple filter
```
VDFaller: Currently `get_model_parents` doesn't sanitize, maybe I make that an arg?

src/dbt_mcp/prompts/dbt_cli/get_lineage.md
```diff
@@ -0,0 +1,2 @@
+The get_lineage command retrieves the lineage information for a specified model in the dbt project, including its parent and child models.
```
VDFaller: delete me

b-per:
Thanks Vince!

I am wondering if we should not create new tools instead of making `get_model_parents`/`get_model_children` be able to either query the Metadata API or the local artefacts.

Here are my Pros/Cons of having new tools for `get_model_parents/children` dedicated to the CLI/manifest

- Pros
  - it would be easy to activate those along the rest of the CLI tools and deactivate the dbt platform ones if needed (just activating the CLI toolset)
  - people could query both the `get_model_parents` from the metadata API and the new local tool in a single LLM session/context to compare the changes that they are introducing
  - it feels simpler to understand what tool does what and easier to know which ones someone might want to activate/deactivate
- Cons
  - we already have many tools and it would add more (but realistically most people shouldn't activate all tools and tweak those to their use case)
  - we'd need to find good names and descriptions to explain to the LLM the difference between the children from metadata and the children from manifest if we want to avoid the LLM to get confused


So, I am in the camp of adding this functionality but in new tools ; and I'd be keen to hear other people's opinions about it.

--- 

As a side note, would you be able to set up [signed commits](https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#signed-commits) for this repo? We can bypass this check at the PR level as repo admins, but this repo expects all commits to be signed now.

VDFaller:
@b-per Crap, that's how I originally had it (shouldn't have squashed 😢 )

I didn't like it because when I was trying it out, exactly like you pointed to, it seemed to arbitrarily pick which tool it used.  So I could run very similar queries and it would give two different results.  We could give better names/descriptions so the tool wouldn't get confused but I don't think the user would necessarily know if they were getting the answer they expected.  


* me
    * Get me the model parents for jaffle_shop.orders. 
* MCP 
    * Okay there we've got some options, do you want production parents, or local parents, recusive or not? 
* me 
    * What? 

I think it would just run the tool and the user would see it asking to run  `get_model_lineage(...)` and go "that seems right", without knowing the nuance.  

If it were to be two separate tools, would you think it should be a cli tool or a discovery tool.  My entire thought process was "Discovery is NOT just platform", especially after listening to Jason's talk at Coalesce where he talked about them abstractly.  This very much relates to #418 in my head.  

Rebased with gpgSign on, no idea why it was set to false for this repo. 

Also on this 

> we already have many tools and it would add more (but realistically most people shouldn't activate all tools and tweak those to their use case)

* Do y'all have data to show that's the case?  I'd expect people to just give it everything.

DevonFulcher:
Hey @VDFaller and @b-per, I'm sorry for the back-and-forth on this. I told Vince that I was appreciative of sticking with the existing `get_model_parents`/`get_model_children` tools and routing between the local or remote version depending on the user's config. Let's get aligned on this. I think the `Pros` you listed are valid, Benoit, but they may not be the features worth optimizing greatly for.

I like the router approach because the agent typically doesn't care whether the information is coming from a local or remote source; the user cares more about that. Also, with the latest config changes, it is quite easy to point the agent to local or remote. If `DBT_HOST` is present, use GQL; otherwise, use the local version. Turning on/off more tool options depending on local or remote usage is more flexible, but it is also more complex, and I don't think most users want to use both local and remote at the same time.

Furthermore, this router approach can be applied to the Semantic Layer tools in the future. It is a source of frustration for some users that these tools don't work locally.

src/dbt_mcp/dbt_cli/models/lineage_types.py
```diff
@@ -0,0 +1,138 @@
...
+        if direction in ("both", "parents"):
+            if not recursive:
+                # direct parents only
+                for pid in parent_map.get(model_id, []):
+                    parents.append(Ancestor.model_validate({"model_id": pid}))
+            else:
+                # Build nested ancestor trees. We prevent cycles using path tracking.
+                def _build_ancestor(node_id: str, path: set[str]) -> Ancestor:
+                    if node_id in path:
+                        # cycle detected, return node without parents
```
VDFaller: This is a safety thing, But since they should be DAGs, I think I could get rid of this so we don't lose parents/children that have already been seen.

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -177,6 +179,34 @@ def show(
...
+            manifest = json.load(f)
+        return manifest
+
+    def get_model_lineage_dev(
+        model_id: str,
+        direction: Literal["parents", "children", "both"] = "both",
+        exclude_prefixes: tuple[str, ...] = ("test.", "unit_test."),
+        *,
+        recursive: bool,
+    ) -> dict[str, Any]:
```
DevonFulcher, line 199: Non-blocking: I think this should have essentially the same function signature as the other lineage tool: https://github.com/dbt-labs/dbt-mcp/pull/461/files#diff-6d91f0721d8dcde8199de504338811a7063757ec13f32eca508bfbc8b663a54bR390-R396

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -177,6 +179,34 @@ def show(
         args.extend(["--output", "json"])
         return _run_dbt_command(args)
 
+    def _get_manifest() -> dict[str, Any]:
+        """Helper function to load the dbt manifest.json file."""
+        import json
```
DevonFulcher: Can we import this at the top of the file?

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -177,6 +179,34 @@ def show(
...
+
+        _run_dbt_command(["parse"])  # Ensure manifest is generated
+        cwd_path = config.project_dir if os.path.isabs(config.project_dir) else None
```
DevonFulcher, line 187: Non-blocking: when the server starts up, we should make all paths absolute if they aren't already.

src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -177,6 +179,34 @@ def show(
         args.extend(["--output", "json"])
         return _run_dbt_command(args)
 
+    def _get_manifest() -> dict[str, Any]:
```
DevonFulcher: I think this function should return a Pydantic model rather than a dictionary. We don't have to create a model with fields for every property of the manifest, just the ones we use.

src/dbt_mcp/dbt_cli/models/lineage_types.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if direction in ("both", "children"):
+            if not recursive:
+                children = [
+                    Descendant.model_validate({"model_id": cid})
+                    for cid in child_map.get(model_id, [])
+                ]
+            else:
+                # Build nested descendant trees. Prevent cycles using path tracking.
+                def _build_descendant(node_id: str, path: set[str]) -> Descendant:
```
DevonFulcher: It seems like we could consolidate the two versions of this function into one. The differences are minimal.

src/dbt_mcp/dbt_cli/tools.py
VDFaller: I'll make a separate PR once they're both in to align them.  Cool?

src/dbt_mcp/dbt_cli/models/lineage_types.py
VDFaller: had to do some funky casting to make mypy happy.  let me know if you feel like it's better.

src/dbt_mcp/dbt_cli/tools.py
DevonFulcher: Sounds good, thank you!

src/dbt_mcp/dbt_cli/models/lineage_types.py
DevonFulcher: Yes, I see. I do think it is better. It is a bit easier to read. Thanks for making that change!

Issue #408, opened 2025-10-16
[Feature]: Make it easier to use dbt-mcp with local dbt project without depending on dbt Platform


moomindani:
### Is your feature request related to a problem?

When I wanted to talk to my local MCP server without dbt Platform access, I needed to disable several features. (Otherwise it redirects to dbt Platform with OAuth).

```
{
  "mcpServers": {
    "dbt-mcp": {
      "command": "uvx",
      "args": [
        "dbt-mcp"
      ],
      "env": {
        "DBT_PROJECT_DIR": "/path/to/my/project/",
        "DBT_PROFILES_DIR": "/path/to/my/profile/",
        "DBT_PATH": "dbt",
        "DISABLE_SEMANTIC_LAYER": "true",
        "DISABLE_DISCOVERY": "true",
        "DISABLE_SQL": "true",
        "DISABLE_ADMIN_API": "true"
      },
      "disabled": false,
      "autoApprove": [],
      "disabledTools": []
    }
  }
}
```

### Describe the solution you'd like

It would be great if we can have easy way to force local mode without redirecting to dbt Platform, or at least better documentation for that.

DevonFulcher:
Thanks for sharing! We have discussed adding one environment variable for disabling all the dbt platform tools: `DISABLE_DBT_PLATFORM`. We could also improve our documentation.

moomindani:
Sounds great, thank you.

jeremyholtzman:
I wasted a whole day figuring that out. If you hadn't posted this issue I never would have, so thank you!!

DevonFulcher:
Hey y'all, I'm sorry for the frustration regarding this issue. We should have documented this better! My colleague is working on documentation improvements [here](https://github.com/dbt-labs/docs.getdbt.com/pull/8137), including better information on how to run dbt MCP without dbt Platform. Additionally, PR #424 improves the configuration system so that it does not require `DISABLE_*` environment variables to run tools that don't depend on dbt Platform.

I hope that helps! I am going to close this issue for now, but let me know if you run into any more issues. Additionally, the #tools-dbt-mcp channel in [our community Slack](https://www.getdbt.com/community/join-the-community) is a great place to discuss any other issues you may run into.

PR #424, opened 2025-10-30
#380 better handling of settings misconfig

### LINKED ISSUE #380: [Feature]: Better handling of Settings Misconfiguration.

Settings Misconfiguration doesn't warn the user if something is wrong and sometimes causes a barrier to entry.  I want to have pydantic validate its fields through pydantic. So erroring and logging happens in the settings themselves.  This would be a small refactor so I wanted buy in before I did it.  I plan on just changing #373 to match this.  

I think these basic rules should exist.  

* if environment variables aren't set, disable tools that rely on them and log with a warning that tools are being disabled
* if environment variables are set, but not valid (like `dbt_host` starting with `metadata`, error out and let the user know
* else, let er rip

I've made a small PoC below to kind of explain my thinking.  

<details> 
<summary> PoC Code </summary> 

```python
# %%
import os
import logging
import re

from pydantic import Field, field_validator, model_validator
from pydantic_settings import BaseSettings, SettingsConfigDict

class DbtMcpSettings(BaseSettings):
    model_config = SettingsConfigDict(
        env_prefix="",
        case_sensitive=False,
        env_file=".env",
        env_file_encoding="utf-8",
        extra="ignore",
        frozen=True
    )
    dbt_host: str | None = Field(None, alias="DBT_HOST")
    dbt_prod_env_id: int | None = Field(None, alias="DBT_PROD_ENV_ID")
    dbt_token: str | None = Field(None, alias="DBT_TOKEN")

    ...
    disable_semantic_layer: bool = Field(False, alias="DISABLE_SEMANTIC_LAYER")
    disable_discovery: bool = Field(False, alias="DISABLE_DISCOVERY")
    disable_sql: bool | None = Field(None, alias="DISABLE_SQL")
    disable_admin_api: bool = Field(False, alias="DISABLE_ADMIN_API")
    ...

    def validate_platform(self) -> list[str]:
        errors = []
        if self.dbt_host is None:
            errors.append("DBT_HOST is not set.")
        if self.dbt_prod_env_id is None:
            errors.append("DBT_PROD_ENV_ID is not set.")
        if self.dbt_token is None:
            errors.append("DBT_TOKEN is not set.")
        return errors

    @field_validator("dbt_host", mode="after")
    @classmethod
    def validate_host(cls, v: str | None) -> str | None:
        host_ok = re.compile(r"^(?!metadata\.)(?!semantic-layer\.)[a-zA-Z0-9.-]+$")
        if v is not None and not host_ok.fullmatch(v):
            raise ValueError(
                "DBT_HOST must not start with 'metadata' or 'semantic-layer'."
            )
        return v


    @model_validator(mode="after")
    def auto_disable_platform(self) -> "DbtMcpSettings":
        errors = self.validate_platform()
        if errors:
            # object.__setattr__ is used in case we want to set values on a frozen model
            object.__setattr__(self, "disable_semantic_layer", True)
            object.__setattr__(self, "disable_discovery", True)
            object.__setattr__(self, "disable_sql", True)
            object.__setattr__(self, "disable_admin_api", True)
        
            
            logger.warning(f"Platform features disabled due to misconfigurations:\n    {"    \n ".join(errors)}")
        
        return self
# %%
os.environ["DBT_HOST"] = "cloud.getdbt.com"
os.environ["DBT_PROD_ENV_ID"] = "123"
os.environ["DBT_TOKEN"] = "abc"
settings = DbtMcpSettings() 
print(settings.model_dump())  # {'dbt_host': 'cloud.getdbt.com', 'dbt_prod_env_id': 123, 'dbt_token': 'abc', 'disable_semantic_layer': False, 'disable_discovery': False, 'disable_sql': None, 'disable_admin_api': False}
# %%
# if the host isn't set, the platform features are disabled
os.environ.pop("DBT_HOST", None)
settings = DbtMcpSettings() 
# logs
# Platform features disabled due to misconfigurations:
#     DBT_HOST is not set.
# now the platform features are disabled
print(settings.model_dump())  # {'dbt_host': None, 'dbt_prod_env_id': 123, 'dbt_token': 'abc', 'disable_semantic_layer': True, 'disable_discovery': True, 'disable_sql': True, 'disable_admin_api': True}
# %%
# if the host is set incorrectly, a value error is raised
os.environ["DBT_HOST"] = "metadata.cloud.getdbt.com"
settings = DbtMcpSettings() # Value error, DBT_HOST must not start with 'metadata' or 'semantic-layer'
print(settings.model_dump())  
```
</details> 

#373 is currently my alternative. But I think this is cleaner.  

### Additional context

_No response_

VDFaller:
## Summary
Setting up the config wrong (through environment variables) now gives better information about possible mistakes. 

## What Changed
* Badly configured settings now fail on settings validation. 
* Unconfigured settings just turn off tools and warn user. 
* More validation on what the fields should look be. 
* Priority is on disabling tools, but anything that explicitly misconfigured, I error out loudly.  

Also added in a generic fixture `env_setup` for anything that needs auto-disable to NOT happen in testing. 

## Why
Barrier to entry/trust erosion if people don't know why things aren't working right.  
Also, less to setup and don't let mcp use tools that have guaranteed no chance to work.  

## Related Issues
Closes #380

## Checklist


src/dbt_mcp/config/settings.py
```diff
@@ -144,6 +146,58 @@ def usage_tracking_enabled(self) -> bool:
...
+    @field_validator("dbt_path", mode="after")
+    @classmethod
+    def validate_file_exists(cls, v: str | None, info: ValidationInfo) -> str | None:
+        """Validate a path exists in the system.
+
+        This will only fail if the path is explicitly set to a non-existing path.
+        It will auto-disable upon model validation if it can't be found AND it's not $PATH.
+        """
+        # Allow 'dbt' as a special case as it's expected to be on PATH
+        if v == "dbt":
```
DevonFulcher: Let's add `dbtf` here, as I believe that is a common alias for dbt Fusion.

src/dbt_mcp/config/settings.py
```diff
@@ -166,6 +220,33 @@ def parse_disable_tools(cls, env_var: str | None) -> list[ToolName]:
...
+        """Auto-disable features based on required settings."""
+        # platform features
+        if (
+            not self.actual_host
+        ):  # host is the only truly required setting for platform features
+            # object.__setattr__ is used in case we want to set values on a frozen model
+            object.__setattr__(self, "disable_semantic_layer", True)
+            object.__setattr__(self, "disable_discovery", True)
+            object.__setattr__(self, "disable_admin_api", True)
+            object.__setattr__(self, "disable_sql", True)
```
DevonFulcher, line 260: Nice!

tests/conftest.py
```diff
@@ -0,0 +1,79 @@
+from pathlib import Path
+import pytest
+from contextlib import contextmanager
+import os
+
+
+@pytest.fixture
+def env_setup(tmp_path: Path, monkeypatch):
```
DevonFulcher, line 26: Nice!

DevonFulcher:
This will also help address #408

src/dbt_mcp/config/settings.py
```diff
@@ -85,6 +87,32 @@ class DbtMcpSettings(BaseSettings):
     # Developer settings
     file_logging: bool = Field(False, alias="DBT_MCP_SERVER_FILE_LOGGING")
 
+    def __repr__(self):
```
VDFaller, line 90: This is just so I can see in the GHActions if it's an error is due to auto-disable or not.  pytest truncates when it shows `DbtMCPSettings(some_stuff="soiefjo", .... , final_arg="aoijf")`

DevonFulcher (approved):
Looks great!

PR #428, opened 2025-11-03
Improve environment variable configuration in CONTRIBUTING.md

DevonFulcher:
## Summary

We no longer use `dotenv`, so this section of the docs is a bit outdated.

## What Changed

Small docs update

## Checklist


b-per:
Are we 100% sure about this?
We still load `.env` from `pydantic_settings` `BaseSettings`: https://github.com/dbt-labs/dbt-mcp/blob/5216c66a03518dd465d7642259a1345ea5684710/src/dbt_mcp/config/settings.py#L43-L44

PR #430, opened 2025-11-04
Extend get_job_run_error to support warnings

### LINKED ISSUE #409: [Feature]: Allow MCP server to return jobs with warnings

Currently, it does not seem possible to get warning results or a list of jobs that succeeded with warnings. This is a big hurdle for my workflows - it would be very helpful to be able to analyze warnings and potential resolutions similar to errors using and MCP/LLM workflow.

The MCP server has a tool to return a list of jobs that succeeded with warnings.

It doesn't seem like warning information is returned in any part of the logs available to the MCP server, so I don't think there is alternative. If there is, would be interested in knowing!

### Additional context

_No response_

jairus-m:
## Summary
Extends `get_job_run_error` to include an option to output targeted information of all warnings at three major levels:
1. Source freshness, parsed from `source.json` 
2. Run results warnings, parsed from `run_results.json`
3. Log warnings, parsed from the log text

Inspect `src/dbt_mcp/prompts/admin_api/get_job_run_error.md` to get a better sense of the structure of the output. 

## What Changed

- Add a new `WarningFetcher` class that closely mirrors `ErrorFetcher`
  - get run steps --> parse artifacts/logs --> build structured output --> return output 
- Biggest changes overall were consolidating classes in a new generalized module `src/dbt_mcp/dbt_admin/run_artifacts/` to cohesively house the error and warn specific classes in `src/dbt_mcp/dbt_admin/run_artifacts/parser.py`
  - Kept as much implementation parity between `WarningFetcher` and `ErrorFetcher` 
  - Created base class, `JobRunFetcher` for shared helper methods

## Why

Address the request for a tool that allows for analyzing "warnings and potential resolutions similar to errors using an MCP/LLM workflow".

## Related Issues

Closes #409
- This tool does not return a list of jobs with warnings (future extension?)

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com)
    - [PR](https://github.com/dbt-labs/docs.getdbt.com/pull/8153)

#### Claude Desktop

**1. Error Only**
Screenshot 2025-11-08 at 7 52 33 PM

**2. Warnings Only**
Screenshot 2025-11-08 at 7 53 06 PM

**3. Both Error + Warnings**
Screenshot 2025-11-08 at 7 54 32 PM

### DIFF
--- a/src/dbt_mcp/dbt_admin/run_artifacts/parser.py
+++ b/src/dbt_mcp/dbt_admin/run_artifacts/parser.py
@@ -545,6 +545,9 @@ class WarningFetcher(JobRunFetcher):
         if "[WARNING]" not in step.logs:
             return []
 
+        # TODO: Fusion logs differ from core (currently core only)
+        # Need to explore and implement a solution for this
+
         # Remove ANSI color codes to focus on text
         ansi_escape = re.compile(r"\x1b\[[0-9;]*m")
         clean_logs = ansi_escape.sub("", step.logs)
@@ -644,7 +647,7 @@ class WarningFetcher(JobRunFetcher):
     def _create_summary(
         self,
         warning_steps: list[OutputStepSchema],
-        log_warnings: list[OutputResultSchema],
+        log_warning_steps: list[OutputResultSchema],
     ) -> dict[str, int]:
         """Create a summary of warning counts."""
         test_warnings = 0
@@ -658,14 +661,14 @@ class WarningFetcher(JobRunFetcher):
                 elif "source." in unique_id:
                     freshness_warnings += 1
 
-        log_warnings_count = len(log_warnings)
-        total_warnings = test_warnings + freshness_warnings + log_warnings_count
+        log_warnings = len(log_warning_steps)
+        total_warnings = test_warnings + freshness_warnings + log_warnings
 
         return {
             "total_warnings": total_warnings,
             "test_warnings": test_warnings,
             "freshness_warnings": freshness_warnings,
-            "log_warnings": log_warnings_count,
+            "log_warnings": log_warnings,
         }
 
     def _empty_response(self, reason: str = "No warnings found") -> dict[str, Any]:


b-per:
Thanks for starting the work here! 🙌 

### thinking out loud

I haven't checked all of it in details, but my first comment is whether we'd want to have 2 different tools for
`get_job_run_error` and `get_job_run_warnings` or if we'd want to consolidate them with for example `get_job_run_error` accepting an argument called `include_warnings` which might be set to `False` by default.

I am wondering what flow would benefit from 2 tools vs 1.
Would there be a case where someone is interested just in warnings but doesn't care about errors at all?

### semver consideration

We just need to keep in mind that we'd want any change to be backward compatible if we want to stay in the `1.` semver. For example, renaming `get_job_run_error` to `get_job_run_error_and_warning` would be a breaking change that would require us moving to 2.y.z but I wouldn't consider adding a new tool `get_job_run_error_and_warning` to be a breaking change once we have [the concept of "allowlist"](https://github.com/dbt-labs/dbt-mcp/issues/418) set up.

--- 

_Also, right now we have `get_job_run_error` with no "s" and `get_job_run_warnings` with an "s". We might want to get the same convention for both._

jairus-m:
thanks for taking a peak ! @b-per 

> whether we'd want to have 2 different tools or 1

Yes, I think that's a great point on potentially consolidating. There are two reasons I'm thinking of that may make two separate tools appropriate (?):
1. The returned context of having `get_job_run_error(include_warning=true)` may be massive depending on if the error returns a truncated logout + multiple warnings across steps, but this may be a worth it for streamlining the amount of tools?
2. Outside of dbt-autofix and having agents focus on deprecations, having `get_job_run_warnings` explicitly may be a valuable add-on for being able to hunt down warnings and deprecations via an LLM??

But even as I think about those two cases, I'm starting to lean towards one tool now lol. I'm thinking about how minimal the use case may be for folks to actually want to get warnings explicitly over errors. Should there also be a param to set for `include_errors` so there's an option to return warnings only?

> semver consideration

heard and noted!

> get_job_run_error with no "s" and get_job_run_warnings with an "s". We might want to get the same convention for both.

I did think about this for a quick second 😅  and also that non-alignment.  My rationale was that get an error for a job run involves one failed step and usually one main error (tho you can have multiple test failure that fail in parallel) and that warnings span multiple steps and they are usually quite many (at least for what I've seen).

I do think it does feel funny - what do you think on the rationale here? And if we updated to `get_job_run_errors` with an 's' , that would be a breaking change that requires moving to "2.x.y" right?

@ttusing would love to get your input here !

ttusing:
If I'm understanding correctly, I agree that it's pretty narrow to want to get warnings and definitely exclude errors.

Like you enumerated, there are multiple places where warnings occur (where maybe errors do not as much). It may be useful at some point to filter warnings and errors by type (deprecation, freshness, etc) or other current future features (count of downstream models, rows failing in data test, sub-severity level). Do errors and warnings share enough commonality to be grouped together?

My gut is yes, in most cases they are the same, so a single tool makes sense. I've personally thought about a future with more logging severity levels (which is kind of possible by printing to the log in tests/etc) and I think treating this like standard programming logging levels as opposed to separate object types is preferable.

jairus-m:
@b-per @ttusing 

With the latest [94ba14c](https://github.com/dbt-labs/dbt-mcp/pull/430/commits/94ba14c25258eeb8539bc30c96287129079f6289), I have:

```python
# As a POC but does not feel right
get_job_run_error(
      run_id: int,
      include_warnings: bool,
      warnings_only: bool
  )
```

But should the name be changed? And should the input params be something completely different like:

```python
get_job_run_issues(
      run_id: int,
      severity: list[str] = ["error", "warn"]  # or ["warn"] and default is ["error"]
  )
```

This would obviously be a breaking change but just thought to throw out these ideas! And to brainstorm what could be done.

ttusing:
Hmm maybe `error` for the name is fine? Feel like some parent name for `warning` and `error` would be helpful like `issue`, but I don't think thats used in the DBT lexicon currently. I think `warning` is a severity of error loosely? I think `message` is the term for Python logging.

https://docs.getdbt.com/reference/resource-configs/severity

https://docs.python.org/3/library/logging.html

b-per:
I prefer option 1.

While getting warnings is a nice feature, I feel like the majority of people will be focusing on wanting the errors ; the warning might be more "niche", so I am OK with keeping error in the name of the tool.

src/dbt_mcp/dbt_admin/run_artifacts/parser.py
```diff
@@ -0,0 +1,684 @@
...
+                    freshness_warnings += 1
+
+        log_warnings_count = len(log_warnings)
+        total_warnings = test_warnings + freshness_warnings + log_warnings_count
+
+        return {
+            "total_warnings": total_warnings,
+            "test_warnings": test_warnings,
+            "freshness_warnings": freshness_warnings,
+            "log_warnings": log_warnings_count,
```
b-per: Not a huge deal at all but we call this one `..._count` and not the others.

src/dbt_mcp/dbt_admin/run_artifacts/parser.py
```diff
@@ -0,0 +1,684 @@
...
+    ) -> OutputStepSchema | None:
+        """Check for source freshness warnings in sources.json artifact."""
+        return await self._check_source_freshness_results(
+            step=step,
+            status_filter=["warn"],
+            message_prefix="Source freshness warning",
+            result_status="warn",
+        )
+
+    def _extract_log_warnings(self, step: RunStepSchema) -> list[OutputResultSchema]:
```
b-per, line 540: This function might be OK for now with Core, and we can get it merged, but we should also check it for Fusion. logs are quite different and we will likely need two different implementations for Core and Fusion.

b-per (approved):
LGTM. I haven't reviewed it line by line but the overall flow looks OK.

I have added 2 comments that are non-blocking, but we might want to at least raise an issue for following-up on making it compatible with Fusion logs.

Those logs are still evolving, so it is not an issue we'd fix today, but we'll likely want to have it working with Fusion when it gets to GA.

src/dbt_mcp/dbt_admin/run_artifacts/parser.py
jairus-m: Yeah that's a good call - i definitely should've called that out more explicitly. I'm going to add to the comments just as a reminder for future me that this is an outstanding exploration that will likely need added development in the future.

src/dbt_mcp/dbt_admin/run_artifacts/parser.py
jairus-m: @b-per the function that creates this summary has a `log_warnings` param that will conflict with a variable that's names the same:


```python
    def _create_summary(
        self,
        warning_steps: list[OutputStepSchema],
        log_warnings: list[OutputResultSchema],
    ) -> dict[str, int]:
        """Create a summary of warning counts."""
    ...
    log_warnings = len(log_warnings) # this would be the conflict
    
```

I'll make some changes to make that not-the-case and to make it consistent!

The param should probably be `log_warnings_steps` and then you are right, the variable below should be `log_warnings`

jairus-m:
@b-per I can create an issue pertaining to the fusion v core shenanigans !

PR #434, opened 2025-11-10
[feat] Add `get_source_details` tool

### LINKED ISSUE #433: [Feature]: Expose Column and Data Type Information for get_all_sources

This issue is a follow-up to [issue #336](https://github.com/dbt-labs/dbt-mcp/issues/336). While that issue covered important semantic information, it did not include the ability to retrieve column names and data types for dbt sources.

This capability is fundamental to a productive developer experience, particularly when building staging models or starting new transformation chains. Its absence is a significant friction point in the development workflow.

#### Why Developers Need Source Column Information

When writing dbt SQL (especially staging models), a developer's primary task is to transform raw data from upstream source tables. To do this effectively, they constantly need to know the schema of those sources.

_No response_

### Additional context

**The key development tasks that require this information are:**

* Fulfilling Development Requests: The entire workflow often starts with a request like, "Please add this column from this source table to our new model." A developer cannot even begin this task without being able to see what columns are available in the source.

* Mapping Raw Fields: What is the exact column name in the raw source table? Is it customer_id, cust_id, or ID?

* Verifying Data Types:

  * Can I SUM() this total_amount column, or is it a string that needs to be CAST() first?

  * Is this created_at column a TIMESTAMP or a VARCHAR?
* Selecting Columns: What columns are even available in this source?

xrpza:
## Summary

Added `get_source_details` tool to retrieve detailed source information including column-level metadata (names, types, descriptions). This provides pattern alignment with the `get_all_models` --> `get_model_details` approach, allowing users to inspect source structures without having to return extraneous information for all sources at once. 

LLM clients can now efficiently browse all available sources first and _then_ drill-down to specific resource details only when such close inspection is needed. The goal here is to be considerate of token budgets and context window sizes by not requiring `get_all_sources` to return all metadata about all sources all at once.

## What Changed

- **Added Discovery API GraphQL Query**: New `GET_SOURCE_DETAILS` query in `client.py` to fetch source metadata
- **Implemented Tool Method**: New `fetch_source_details()` method in `SourcesFetcher` class supporting lookup by `uniqueId` or `source_name`
- **Tool Registration**: Registered tool in discovery toolset with `METADATA` policy behavior
  - Added to `tool_names.py`
  - Added to `policy.py`
  - Added to `toolsets.py`
  - Registered in `tools.py`
- **Prompt**: Created `get_source_details.md`
- **Tests**: Added tests for the new feature

## Why

Users would like to inspect source table structures (columns, types, descriptions) before writing dbt transformations, etc. Previously, `get_all_sources` only provided high-level metadata, while model-level tool-use pathways like `get_all_models` --> `get_model_details` offered features like column-level inspection. 

This PR's proposed two-tiered approach of `get_all_sources` --> `get_source_details` will provide that detailed inspection without over-burdening context windows or token budgets.

## Related Issues

Closes #433 

## Checklist
- [X] I have performed a self-review of my code
- [X] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/dbt_mcp/discovery/client.py
+++ b/src/dbt_mcp/discovery/client.py
@@ -741,7 +741,7 @@ class SourcesFetcher:
             "sourcesFilter": source_filters,
             "first": 1,
         }
-        
+
         result = await self.api_client.execute_query(
             GraphQLQueries.GET_SOURCE_DETAILS, variables
         )
--- a/tests/unit/discovery/test_sources_fetcher.py
+++ b/tests/unit/discovery/test_sources_fetcher.py
@@ -440,17 +440,7 @@ async def test_fetch_source_details_with_source_name(sources_fetcher, mock_api_c
 async def test_fetch_source_details_source_not_found(sources_fetcher, mock_api_client):
     """Test fetching source details when source identifier doesn't match any sources."""
     # Mock response with no matching sources (empty edges)
-    mock_response = {
-        "data": {
-            "environment": {
-                "applied": {
-                    "sources": {
-                        "edges": []
-                    }
-                }
-            }
-        }
-    }
+    mock_response = {"data": {"environment": {"applied": {"sources": {"edges": []}}}}}
 
     mock_api_client.execute_query.return_value = mock_response
 
@@ -458,7 +448,7 @@ async def test_fetch_source_details_source_not_found(sources_fetcher, mock_api_c
 
     # Should return empty dict when not found
     assert result == {}
-    
+
     # Verify the API was called with identifier filter
     call_args = mock_api_client.execute_query.call_args
     variables = call_args[0][1]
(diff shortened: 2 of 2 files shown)


src/dbt_mcp/discovery/client.py
```diff
@@ -679,3 +717,38 @@ async def fetch_sources(
...
+
+        if unique_id:
+            source_filter["uniqueIds"] = [unique_id]
+        elif source_name:
+            # For backward compatibility, though unique_id is preferred
+            all_sources = await self.fetch_sources()
+            matching = [s for s in all_sources if s.get("name") == source_name]
+            if not matching:
+                return {}
+            source_filter["uniqueIds"] = [matching[0]["uniqueId"]]
```
DevonFulcher: This looks good, thanks!

xrpza: I actually meant to change this one. Originally, I thought the `sources` Discovery API didn't allow the search on names, but it can through `identifier` 🤦‍♂️ . I'm adjusting the approach to be closer to `get_model_details`.

toshihue:
Thank you @xrpza !

PR #440, opened 2025-11-12
Add discovery tools context

wiggzz:
## Summary

Define a discovery tool context that can be created from the FastMCP context in order to allow the discovery tools to be used more flexibly.

## What Changed

- Add a helper function `adapt_with_mapper` that can create a new function to accept different context parameters so we can map between contexts (e.g. `DiscoveryToolContext` and FastMCP's `Context` object)
- Accept a `DiscoveryToolContext` in all the discovery tools.
- Introduce a `@dbt_mcp_tool` decorator to more easily create the `ToolDefinitions`
- Use the `adapt_with_mapper` helper function to adapt the discovery tools before registration in order to avoid binding the tools to the configuration.

## Why

We use these tool definitions in dbt Platform where binding the tools to the configuration is more problematic - we need to be able to parameterize the tools per request more easily.

## Checklist

### DIFF
--- a/src/dbt_mcp/tools/injection.py
+++ b/src/dbt_mcp/tools/injection.py
@@ -86,6 +86,7 @@ def adapt_with_mapper[R](
             return await invoke_func(bound_args, mapped_value)
 
         awrapper.__signature__ = new_sig  # type: ignore[attr-defined]
+        awrapper.__annotations__ = get_annotations(new_sig)
         return cast(Callable[..., R], awrapper)
 
     else:
--- a/tests/unit/tools/test_injection.py
+++ b/tests/unit/tools/test_injection.py
@@ -284,3 +284,12 @@ def test_adapt_with_mapper_provides_type_hints():
 
     assert hints["ctx"] is Context
     assert hints["return"] is str
+
+
+def test_adapt_with_mapper_provides_type_hints_for_async_functions():
+    adapted = adapt_with_mapper(async_greet_user_id, extract_user_id)
+
+    hints = get_type_hints(adapted)
+
+    assert hints["ctx"] is Context
+    assert hints["return"] is str


src/dbt_mcp/discovery/tools.py
```diff
@@ -1,194 +1,209 @@
...
+)
+async def get_source_details(
+    context: DiscoveryToolContext,
+    source_name: str | None = None,
+    unique_id: str | None = None,
+) -> dict:
+    return await context.sources_fetcher.fetch_source_details(source_name, unique_id)
+
+
+DISCOVERY_TOOLS = [
```
DevonFulcher, line 178: We already have this list with tests [here](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/tools/toolsets.py#L37).

src/dbt_mcp/discovery/tools.py
```diff
@@ -1,194 +1,209 @@
...
+    get_model_health,
+    get_exposures,
+    get_exposure_details,
+    get_all_sources,
+    get_source_details,
+]
+
+
+def adapt_dbt_mcp_context(context: Context) -> DiscoveryToolContext:
+    dbt_mcp = cast(DbtMCP, context.fastmcp)
```
DevonFulcher: Would it be better to use an `assert` rather than `cast` here?

src/dbt_mcp/discovery/tools.py
wiggzz: Right... this is just the reimagination of the existing list that was [here](https://github.com/dbt-labs/dbt-mcp/pull/440/files#diff-6d91f0721d8dcde8199de504338811a7063757ec13f32eca508bfbc8b663a54bL81-L182). So if we want to consolidate those lists we can, but they are currently separate.

DevonFulcher (approved):
Looks great, thanks!

DevonFulcher (approved):
Typical type tantrums result in typified tautologies.

Edit: only minor help from ChatGPT on that.

DevonFulcher (approved):

PR #441, opened 2025-11-12
Add search tools to toolsets

DevonFulcher:
## Summary

Add search tools to the toolset dictionary. Documentation: https://github.com/dbt-labs/docs.getdbt.com/pull/8171

## Checklist


b-per:
Do we need a new toolset? Isn't it the same use case as the existing discovery toolset?

DevonFulcher:
> Do we need a new toolset? Isn't it the same use case as the existing discovery toolset?

Yes, I can make these a part of the Discovery toolset. They are implemented separately, though, since they aren't here in OSS.

DevonFulcher:
Adding this tools as part of this PR instead: https://github.com/dbt-labs/dbt-mcp/pull/442

PR #445, opened 2025-11-13
Semantic layer tool context

wiggzz:
## Summary

This extracts the configuration for semantic layer tools into a context object to make the tools static.

## What Changed

- Extract a SemanticLayerToolContext object which holds the semantic layer fetcher
- Define the tools statically with the new @dbt_mcp_tool annotation
- Bind this when creating the tools for dbt-mcp

## Why

This allows us more flexibility in how we use these tools in other contexts (no pun intended) such as in dbt Platform.

## Checklist


DevonFulcher (approved):
🚀

PR #446, opened 2025-11-14
(WIP)Add get_model_ancestors and get_model_descendants  Tool 

### LINKED ISSUE #110: Feature: support for lineage beyond parent/child
I am working with a client who built an add-on for their dbt MCP server, and use Claude Desktop to have a catalog-like experience with dbt and Snowflake.

They are requesting that dbt-mcp makes it possible to:

> Get the full lineage of a model, not only it’s children/parents. To parse the full lineage with the official MCP it would require quite a few calls to the tool to parse through all layers, which would be nice to get the server doing async instead providing the full list as a return.

In their dbt MCP server, they implemented a `dbt_model_lineage` tool that uses the dbt API to succinctly power that experience.

Muizzkolapo:
## Summary
This change adds two new tools to get model lineage separately: one for ancestors (upstream dependencies) and one for descendants (downstream dependencies).

## What Changed
Added two new tools to the Discovery API toolset for retrieving model lineage:

1. **`get_model_ancestors`** - Retrieves all upstream dependencies (ancestors) for a specific dbt model
   - Fetches sources, seeds, snapshots, and models that the target model depends on
   - Returns all ancestors at any depth in a single GraphQL query

2. **`get_model_descendants`** - Retrieves all downstream dependencies (descendants) for a specific dbt model
   - Fetches models, metrics, and exposures that depend on the target model
   - Uses BFS (Breadth-First Search) traversal algorithm for efficient descendant collection
   - Batches queries by level to minimize API calls

2. **`refactor test_get_all_sources_tool to use a config_provider fixture `**

## Why
Users working with dbt MCP in Claude Desktop and other AI agents need to understand model lineage. Splitting lineage into two separate tools provides:

1. **Flexibility**: Users can query just what they need (upstream OR downstream)
2. **Performance**: Avoid fetching unnecessary data when only one direction is needed

Previously, this would require making multiple sequential API calls to `get_model_parents()` and `get_model_children()` to traverse the lineage tree manually.

## Related Issues
Closes #110

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required https://github.com/dbt-labs/docs.getdbt.com/pull/8198

### DIFF
--- a/tests/integration/discovery/conftest.py
+++ b/tests/integration/discovery/conftest.py
@@ -13,7 +13,8 @@ from dbt_mcp.discovery.client import (
 
 
 @pytest.fixture
-def api_client() -> MetadataAPIClient:
+def config_provider() -> DefaultDiscoveryConfigProvider:
+    """Provides a configured DefaultDiscoveryConfigProvider for tests."""
     # Set up environment variables needed by DbtMcpSettings
     host = os.getenv("DBT_HOST")
     token = os.getenv("DBT_TOKEN")
@@ -28,8 +29,12 @@ def api_client() -> MetadataAPIClient:
     # DbtMcpSettings will automatically pick up from environment variables
     settings = DbtMcpSettings()  # type: ignore
     credentials_provider = CredentialsProvider(settings)
-    config_provider = DefaultDiscoveryConfigProvider(credentials_provider)
+    return DefaultDiscoveryConfigProvider(credentials_provider)
 
+
+@pytest.fixture
+def api_client(config_provider: DefaultDiscoveryConfigProvider) -> MetadataAPIClient:
+    """Provides a configured MetadataAPIClient for tests."""
     return MetadataAPIClient(config_provider)
 
 
--- a/tests/integration/discovery/test_discovery.py
+++ b/tests/integration/discovery/test_discovery.py
@@ -447,36 +447,18 @@ async def test_fetch_sources_with_filter(sources_fetcher: SourcesFetcher):
 
 
 @pytest.mark.asyncio
-async def test_get_all_sources_tool():
+async def test_get_all_sources_tool(config_provider):
     """Test the get_all_sources tool function integration."""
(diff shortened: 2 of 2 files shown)


DevonFulcher:
So cool @Muizzkolapo! I'm a bit worried about context window limits. Would this be better as two separate tools: `get_model_ancestors` and `get_model_descendants`?

Muizzkolapo:
> So cool @Muizzkolapo! I'm a bit worried about context window limits. Would this be better as two separate tools: `get_model_ancestors` and `get_model_descendants`?

Good catch @DevonFulcher  I was actually in the process of doing that, should push the changes shortly.

src/dbt_mcp/discovery/client.py
```diff
@@ -270,6 +273,36 @@ class GraphQLQueries:
...
+    GET_MODEL_ANCESTORS = (
+        textwrap.dedent("""
+        query GetModelLineage(
+            $environmentId: BigInt!,
+            $modelsFilter: ModelAppliedFilter
+            $first: Int,
+        ) {
+            environment(id: $environmentId) {
+                applied {
+                    models(filter: $modelsFilter, first: $first) {
```
DevonFulcher, line 285: I'm sorry I didn't point this out before, I just discovered it myself, but the discovery GQL endpoint already has the ability to query lineage:
```
query Lineage($filter: LineageFilter!, $environmentId: BigInt!) {
  environment(id: $environmentId) {
    applied {
      lineage(filter: $filter) {
        
      }
    }
  }
}
```
This query has the benefit of being one http request instead of multiple. I think this is the route we should go. With the [selector syntax](https://docs.getdbt.com/reference/node-selection/syntax), we can still limit to descendants & ancestors.

Muizzkolapo: Thanks @DevonFulcher  I agree, it really makes it easier and simple to work with I will raise a change for it.

Muizzkolapo:
closing this as I have used a different approach.

PR #448, opened 2025-11-18
[feat] Explicit Tool Allow List 

### LINKED ISSUE #418: [Feature]: Explicit allowlist for tools and toolsets

xrpza:
# Summary

Implements an explicit allowlist system for tool configuration, enabling users to lock down tool availability. This prevents new tools from auto-enabling in production agents when dbt-mcp is upgraded.

# What Changed

## Before
Register all tools _except_ denylist
```mermaid
graph LR
    A[Env Vars] -->|DISABLE_TOOLS| B[Settings]
    B --> C[Config]
    C --> D[register_tools]
    D -->|Skip if in exclude_tools| E[FastMCP]
```
**Problem**: Default behavior is "enable everything except disabled" → new tools auto-enable for all users

## After
Register _only_ the allowed tools when allowlist →else "all" (default fallback for backward compatibility)
```mermaid
graph LR
    A[Env Vars] -->|DISABLE_TOOLS<br/>DBT_MCP_ENABLE_TOOLS<br/>DISABLE_*<br/>DBT_MCP_ENABLE_TOOLSETS| B[Settings]
    B --> C[Config]
    C -->|enable_tools<br/>enabled_toolsets<br/>disabled_toolsets| D[register_tools]
    D -->|Precedence Check| E{should_register?}
    E -->|Yes| F[FastMCP]
    E -->|No| G[Skip]
```
**Solution**: When enablement configuration exists, experience becomes "disable unless allowed"

# Why

**From Issue #418**:
> "_This is problematic when we add new tools or toolsets that are enabled by default, as they become available to agents even if the user doesn't intend to use them. This is particularly problematic when users design handcrafted production agents that could change behavior unpredictably._"

# Solution Discussion
Backwards-compatible updates to dbt-mcp behavior allowing users to _explicitly_ control which tools are made available.

**Examples**
```bash
# Lock down production agent to specific tools
export DBT_MCP_ENABLE_TOOLS="query_metrics,list_metrics"

# Enable full toolset
export DBT_MCP_ENABLE_SEMANTIC_LAYER=true

# Enable toolset but exclude specific tool
export DBT_MCP_ENABLE_SEMANTIC_LAYER=true
export DISABLE_TOOLS="get_metrics_compiled_sql"

# Enable multiple toolsets
export DBT_MCP_ENABLE_SEMANTIC_LAYER=true
export DBT_MCP_ENABLE_ADMIN_API=true
```

## Precedence Rules (Highest to Lowest)

1. **Individual tool enable** - Always enables the tool
2. **Individual tool disable** - Always disables the tool (overrides toolset enable)
3. **Toolset enable** - Enables all tools in toolset
4. **Toolset disable** - Disables all tools in toolset
5. **Fallback** - Only reached when rules 1-4 don't apply:
   - If NO enable configuration exists anywhere → **Enabled** (original default behavior, backward compatible)
   - If ANY enable configuration exists → **Disabled** (allowlist mode activated)

**Examples**:
```bash
export DBT_MCP_ENABLE_SEMANTIC_LAYER=true  # Enable whole toolset
export DISABLE_TOOLS="get_dimensions"       # But not this one tool
```

# Related Issues

Closes #418 

# Checklist
- [X] I have performed a self-review of my code
- [X] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

# Additional Notes


xrpza:
Before moving out of draft, I'm working through:
- [X] ~~Some test failures my changes caused related to proxied tools~~
- [X] ~~Checking that this will work with remote server~~

xrpza:
**Note**: I don't think #450 will break the implementation in this PR, but it might be worth merging that change first, having me update this branch, and then confirm that things behave as expected.

@DevonFulcher

src/dbt_mcp/config/config.py
```diff
@@ -125,8 +129,31 @@ def load_config(enable_proxied_tools: bool = True) -> Config:
...
+    disabled_toolsets: set[Toolset] = {
+        ts
+        for ts in {
+            Toolset.SEMANTIC_LAYER if settings.disable_semantic_layer else None,
+            Toolset.ADMIN_API if settings.disable_admin_api else None,
+            Toolset.DBT_CLI if settings.disable_dbt_cli else None,
+            Toolset.DBT_CODEGEN if settings.disable_dbt_codegen else None,
+            Toolset.DISCOVERY if settings.disable_discovery else None,
+            Toolset.DBT_LSP if settings.disable_lsp else None,
+            Toolset.SQL if settings.actual_disable_sql else None,
```
DevonFulcher: It would be great to use a pattern similar to the one found in `_get_disabled_toolsets`. That way this is always updated when new toolsets are added.

src/dbt_mcp/config/settings.py
```diff
@@ -78,6 +78,14 @@ class DbtMcpSettings(BaseSettings):
     )
     disable_lsp: bool | None = Field(None, alias="DISABLE_LSP")
 
+    # Enable tool settings (allowlist)
+    enable_tools: Annotated[list[ToolName] | None, NoDecode] = Field(
+        None, alias="DBT_MCP_ENABLE_TOOLS"
+    )
+    enable_toolsets: Annotated[list[str] | None, NoDecode] = Field(
+        None, alias="DBT_MCP_ENABLE_TOOLSETS"
```
DevonFulcher: This is okay, but I was imagining environment variables that mirror the current `DISABLE_*` environment variables. So, for example, we currently have `DISABLE_SEMANTIC_LAYER`, and there would be a matching environment variable `DBT_MCP_ENABLE_SEMANTIC_LAYER`. What do you think?

src/dbt_mcp/config/settings.py
```diff
@@ -227,24 +238,55 @@ def validate_dir_exists(cls, v: str | None, info: ValidationInfo) -> str | None:
...
+            List of toolset name strings in canonical lowercase_underscore format
+
+        Raises:
+            ValueError: If any toolset names are invalid
+        """
         if not env_var:
             return []
+
+        # Import here to avoid circular dependency
+        from dbt_mcp.tools.toolsets import Toolset
```
DevonFulcher: I don't think we need to import in the middle of the file here. I think it is better at the top with the rest of the imports. AI has a bad habit of adding imports anywhere.

src/dbt_mcp/config/settings.py
```diff
@@ -227,24 +238,55 @@ def validate_dir_exists(cls, v: str | None, info: ValidationInfo) -> str | None:
...
+        toolset_names: list[str] = []
+
+        for toolset_name in env_var.split(","):
+            toolset_name_stripped = toolset_name.strip()
+            if not toolset_name_stripped:
                 continue
             try:
-                tool_names.append(ToolName(tool_name_stripped))
+                toolset_normalized = toolset_name_stripped.lower().replace("-", "_")
+                Toolset(toolset_normalized)
```
DevonFulcher: Is this line necessary?

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -270,11 +270,18 @@ def register_admin_api_tools(
     dbt_mcp: FastMCP,
     admin_config_provider: ConfigProvider[AdminApiConfig],
     exclude_tools: Sequence[ToolName] = [],
+    *,
+    enabled_tools: set[ToolName] | None = None,
+    enabled_toolsets: set | None = None,
+    disabled_toolsets: set | None = None,
```
DevonFulcher: Do you mind adding more specific type information here and to similar edits? What type of set is this?

xrpza:
I'll work through the comments @DevonFulcher thanks for the feedback

src/dbt_mcp/config/settings.py
xrpza: Good catch. I've repositioned the stray import.

src/dbt_mcp/config/settings.py
xrpza: Looks like a remnant from some validation logic I was playing with. Removing.

src/dbt_mcp/config/config.py
xrpza: Good call. Implementing.

src/dbt_mcp/config/settings.py
xrpza: I see where you're coming from and the advantages of sticking to pattern. I'll do some refactoring.

src/dbt_mcp/dbt_admin/tools.py
xrpza: Good catch!

src/dbt_mcp/proxy/tools.py
```diff
@@ -147,6 +158,17 @@ async def register_proxied_tools(
...
+        # Apply precedence logic to each proxied tool
+        # Proxied tools bypass the standard register_tools() flow, so we check here
+        if not should_register_tool(
+            tool_name_str=tool.name,
+            enabled_tools=enabled_tools,
+            disabled_tools=disabled_tools,
+            enabled_toolsets=enabled_toolsets,
+            disabled_toolsets=disabled_toolsets,
+        ):
+            continue  # Skip this tool if precedence rules say it shouldn't be registered
```
DevonFulcher: I think it is better if this were part of `resolve_proxied_tools_configuration`. That way, if all proxied tools are disabled, there will be no network calls.

src/dbt_mcp/tools/toolsets.py
```diff
@@ -88,3 +95,28 @@ class Toolset(Enum):
...
+    unmapped = set(ToolName) - set(TOOL_TO_TOOLSET.keys())
+    if unmapped:
+        unmapped_names = [tool.value for tool in unmapped]
+        raise ValueError(
+            f"The following tools are not mapped to toolsets: {', '.join(unmapped_names)}"
+        )
+
+
+# Validate at import time to catch errors early
+validate_tool_mapping()
```
DevonFulcher, line 124: Very nice!

src/dbt_mcp/tracking/tracking.py
```diff
@@ -55,35 +55,11 @@ def __init__(
...
-                    if settings.disable_lsp:
-                        disabled_toolsets.append(toolset)
-                case _:
-                    assert_never(toolset)
-        return disabled_toolsets
+        return [
+            toolset
+            for toolset, attr_name in TOOLSET_TO_DISABLE_ATTR.items()
+            if getattr(settings, attr_name, False)
+        ]
```
DevonFulcher, line 62: Much better!

src/dbt_mcp/proxy/tools.py
xrpza: I'll dive into this later this afternoon @DevonFulcher

DevonFulcher: I hope you don't mind, but I went ahead and did this!

src/dbt_mcp/config/config.py
```diff
@@ -15,9 +15,30 @@
...
+
+TOOLSET_TO_ENABLE_ATTR = {
+    Toolset.SEMANTIC_LAYER: "enable_semantic_layer",
+    Toolset.ADMIN_API: "enable_admin_api",
+    Toolset.DBT_CLI: "enable_dbt_cli",
+    Toolset.DBT_CODEGEN: "enable_dbt_codegen",
+    Toolset.DISCOVERY: "enable_discovery",
+    Toolset.DBT_LSP: "enable_lsp",
+    Toolset.SQL: "enable_sql",
+}
```
wiggzz, line 40: I wonder if we could generate these names as opposed to having a mapping? Would improve consistency and predictability of the naming.

wiggzz (approved):

src/dbt_mcp/config/config.py
DevonFulcher: I will do this in a follow up.

PR #451, opened 2025-11-19
Refactor discovery fetchers to share pagination

DevonFulcher:
## Summary

This PR consolidates the pagination logic of the discovery tools. This will make it easier to add new types of resources in the future.

## What Changed

Added `PaginatedResourceFetcher` to share pagination between the different resource types.

## Why

Consolidates code and makes it easier to add resource types in the future.

## Checklist


src/dbt_mcp/discovery/client.py
```diff
@@ -415,32 +415,40 @@ async def execute_query(self, query: str, variables: dict) -> dict:
...
+        current = payload
+        for key in path:
+            current = current[key]
+        return current
+
+    def _parse_edges(self, result: dict) -> list[dict]:
         raise_gql_error(result)
-        edges = result["data"]["environment"]["applied"]["models"]["edges"]
+        edges = self._extract_path(result, self._edges_path)
+        if result.get("errors"):
```
alan-andrade: If the result has an error, `_extract_path` would be a no-op. We can report the error first before extracting the path.

src/dbt_mcp/discovery/client.py
```diff
@@ -450,6 +458,66 @@ def _parse_response_to_json(self, result: dict) -> list[dict]:
             parsed_edges.append(node)
         return parsed_edges
 
+    def _should_continue(
+        self,
+        page_info: dict,
```
alan-andrade: qq: sounds like page_info may be standard for all graphql endpoints. Do you see benefit on having it as a pydantic model instead of interacting with dicts?

src/dbt_mcp/discovery/client.py
```diff
@@ -450,6 +458,66 @@ def _parse_response_to_json(self, result: dict) -> list[dict]:
...
+    ) -> list[dict]:
+        environment_id = await self.get_environment_id()
+        collected: list[dict] = []
+        base_variables = variables.copy()
+        current_cursor: str | None = None
+        while True:
+            if len(collected) >= self._max_node_query_limit:
+                break
+            remaining_capacity = self._max_node_query_limit - len(collected)
+            request_variables = base_variables.copy()
```
alan-andrade: I don't see a mutation for `base_variables` Can we remove the variable and just copy `variables` into `request_variables` ?

src/dbt_mcp/discovery/client.py
alan-andrade: I think this `if` is repetitive because `raise_gql_error` already does the same.

alan-andrade (approved):
🚀

PR #452, opened 2025-11-19
Upgrade js-yaml version

DevonFulcher:
## Summary

Overriding js-yaml package version to address [this](https://github.com/dbt-labs/dbt-mcp/security/dependabot/9) security issue.

## Checklist


wiggzz:
@DevonFulcher you should be able to do this without an override by just running `pnpm update js-yaml` - `js-yaml` is required by `@eslint/eslintrc` and it specifies minor and patch versions can be updated https://github.com/eslint/eslintrc/blob/v3.3.1/package.json#L77

DevonFulcher:
> @DevonFulcher you should be able to do this without an override by just running `pnpm update js-yaml` - `js-yaml` is required by `@eslint/eslintrc` and it specifies minor and patch versions can be updated https://github.com/eslint/eslintrc/blob/v3.3.1/package.json#L77

Excellent, thank you!

PR #453, opened 2025-11-21
Get resource details for search resources

DevonFulcher:
## Summary

Recently, we added a new tool, `search`, which provides information on a variety of dbt resources. However, we didn't provide a way to get the details for such resources. This PR fixes that so that details can be fetched for any resource returned by the `search` tool. This change also includes consolidation of the existing resource details tools: `get_model_details`, `get_exposure_details`, and `get_source_details`. Now, all of these tools operate in a similar fashion.

## What Changed

Added tools for retrieving resource details and consolidated how the existing resource details tools work.

## Why

Enable getting details for resources that are returned by the `search` tool and consolidate how these tools work.

## Checklist
- [] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here


DevonFulcher:
> This may be far fetched, but It would be useful to list **what** metadata is available. Could we rely on gql type generation to embed these into the prompt? Maybe manually for now.

Good idea! I'll take that on as a follow-up to improve the tool descriptions to include the returned fields.

src/dbt_mcp/prompts/discovery/get_model_health.md
```diff
@@ -29,8 +29,8 @@ model_name: The name of the dbt model. Only use this when uniqueId is unavailabl
...
 1. PREFERRED METHOD - Using uniqueId (always use this when available):
-   get_model_details(uniqueId="model.my_project.customer_orders")
-   
+   get_model_health(uniqueId="model.my_project.customer_orders")
+
 2. FALLBACK METHOD - Using only model_name (only when uniqueId is unknown):
-   get_model_details(model_name="customer_orders")
-</examples>
\ No newline at end of file
+   get_model_health(model_name="customer_orders")
```
wiggzz, line 35:
```suggestion
   get_model_health(name="customer_orders")
```

PR #456, opened 2025-11-24
Streamline d2 docs and auto-generate tools in README

### LINKED ISSUE #455: [Under the Hood]: Streamline d2 docs and auto-generate tools in README
A current policy exists to keep d2 docs updated as contributors add new tools. There are two minor road bumps that arise:
- Difficult to consistently enforce this as contributors have to remember to do this and manually run `d2` locally
- The d2 diagram is getting crowded and is not the most friendly thing to read in its current state

Solution discussed w/ @DevonFulcher :
- Maintain toolsets in d2 only (i.e remove tools from diagram)
- Auto-generate toolset + tools in `README.md`

With this, I was thinking to incorporate that automation as a pre-commit hook and/or CI check.

This way, the d2 diagram is cleaner, the docs in the `README.md` are automatically kept up-to-date/checked, and now, the main place to maintain and update docs would be in the dbt docs [about-mcp.md](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/docs/dbt-ai/about-mcp.md).

jairus-m:
## Summary

Remove tools from d2 diagram and instead, auto-generate them in the `README.md`. In addition, auto-generate toolsets in d2 diagram.

## What Changed

- Added `scripts/generate_docs.py`
  - Can either auto-generate toolset + tools or simply check if `README.md` is updated
  - Can either auto-generate toolset or simply check if `diagram.d2` is updated
  - Is this folder appropriate?
- Added `task docs:tools` that wraps a `uv run` call to script
- Added `task docs:tools` to pre-commit and CI workflow

## Why

- Refer to issue

## Related Issues

Closes #455 

## Checklist

## Additional Notes

Behavior:

1. Pre-commit hook is triggered when `src/dbt_mcp/tools/tool_names.py` or `src/dbt_mcp/tools/toolsets.py` or `README.md` or `docs/diagram.d2` is updated
    - The hook will regenerate the `## Tools` section in the README
    - The hook will regnerate the toolset section in d2 diagram as well as the PNG file
2. CI check
    - On every PR, checks if README and d2 config/diagram is in sync with registered tool definitions


scripts/generate_readme_tools.py
```diff
@@ -0,0 +1,110 @@
...
+
+TOOLSET_TO_README_HEADING = {
+    Toolset.DBT_CLI: "dbt CLI commands",
+    Toolset.SEMANTIC_LAYER: "Semantic Layer",
+    Toolset.DISCOVERY: "Discovery",
+    Toolset.SQL: "SQL",
+    Toolset.ADMIN_API: "Admin API",
+    Toolset.DBT_CODEGEN: "Codegen",
+    Toolset.DBT_LSP: "dbt LSP",
+}
```
jairus-m: instead of mapping these again into a "nicely-formatted" enum, should we just use the raw `Toolset.<tool>.value` as the strings? So we can fully automate the process?

Although this enum is nice, if a new toolset were to be added,`TOOLSET_TO_README_HEADING` would have to be manually updated.

docs/diagram.d2
jairus-m, line 1: In `diagram.d2`, toolsets would still need to be maintained in addition to dbt docs. Is that fine? Only the tools themselves are auto-generated in the README

scripts/generate_readme_tools.py
VDFaller: What if you just used a [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum). So that you don't even have to deal with value?

DevonFulcher: Ya, it would be great if we didn't have to update `TOOLSET_TO_README_HEADING`. We can likely take `Toolset.<tool>.value` and use some simple string manipulation to make the values more readable.

docs/diagram.d2
DevonFulcher: Automating `diagram.d2` generation would be another great enhancement, but this PR already improves the process, so it isn't necessary to get this PR merged in.

jairus-m: @DevonFulcher totally! I'm gonna take some time today or tomorrow to think through the the cleanest approach here for getting the d2 diagram included and will try to consolidate that with the README approach.

PR #461, opened 2025-11-27
implement get_lineage tool for complete lineage across dbt resources

Muizzkolapo:
## Summary

A User building a catalog-like experience with dbt and Snowflake via Claude Desktop requested full lineage capabilities https://github.com/dbt-labs/dbt-mcp/issues/110:

> "Get the full lineage of a model, not only its children/parents. To parse the full lineage with the official MCP it would require quite a few calls to the tool to parse through all layers, which would be nice to get the server doing async instead providing the full list as a return."

---

## Discovery Phase: GraphQL Introspection

We used GraphQL introspection to validate design decisions and understand the Discovery API's capabilities.

### Why we used Sequential Search Strategy

**Question:** Does `LineageFilter` support searching by name/identifier? [LineageFilter](https://metadata.cloud.getdbt.com/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4TADpIAR4D6hKAngA4IAUSAhountiADICWSCtA5ggGJsANigQAnZgEo8OfATwdyMFAISCwAZ2m45cuom068YBOqii25FGwhIDO4wDNaMYQDVagmAjtyylLbKGeHregYYA1hxgPjoQDgAqFAgBQTohMRFRGbEJSSmpafShBTqRSNFhQQC%2B2Xg1lXU%2B9QT1VSBVQA)

**Finding:**  - `LineageFilter` only accepts full unique IDs (no `identifier` or `name` field). Due to this Sequential search across resource types is necessary.

---

### Two API Calls for "both" Direction

**Question:** Does `ModelLineageNode` include direction metadata to categorize ancestors vs descendants? 
[View ModelLineageNode Introspection](https://metadata.cloud.getdbt.com/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4TADpIAR4D6hKAngA4IAUSAhountiALIRgIA2AMgJZIJaAcwQA5dgmYBKPDnwE8dRLgV4AZry5gAzrJWrF9SfNUdtUAE69yKXhCT6CAX30ukTkE6A)

**Finding:** - No direction metadata exists (no `relationshipDirection`, `isAncestor`, `isDescendant`, or `dependsOn` fields). Two separate API calls required for "both" direction.

---

### Graphql Client-Side Pagination

**Question:** Does the `lineage` query support server-side pagination parameters?

[Introspect](https://metadata.cloud.getdbt.com/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4TADpIAR4D6hKAngA4IAUSAhountiAILnkA2AlgmAMopaKBMwCUeHPgJ4AZjw5gAzhNzTpdRKrV5aAJwDmyydvX0RUk2UoqLJvBvN3pAay5IwWkwF9P0n7f8Cfy8QLyA
)

**Finding:** - No pagination parameters available (no `first`, `after`, `offset`, or `limit` arguments). Client-side 50-node limit is necessary.

---

### List Of Resource Search Support

**Question:** Which resource types support searching by name (via `identifier` field)?

[link Introspection](https://metadata.cloud.getdbt.com/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4TADpIAR5wRgIA2AMgJZIICGA5gungPosoCeADggBRK1EzbCACyxMlRoMEAOQkiAlHhz4CeAYlzq8AM0pkwAZxUbBCPJx6nNFgNbUweCLoAq3C8DOI8DpE4BfPCCA7WCwowgYACcoBCk6RmY2Kz5bYRAAZSjY%2BOpEuQUQZVUdWzCCfUMTL1tLDxtzX0dnNwbapr9A4PC1ULUjBAQwBJlk9g9%2BcwzMoZH8mXkSJRUK7wQ1qtJjRp9U3fsWl3drDp8unpCw-oIjAS4jAAsIFFGk1gmeKaE8EUz7p4vN6FZbFVZqAjlCF6AzbGrreqnBEXY7tZEtEK9Ag3PAIAAeXAgRhieWk7xSk3SvxAAFECUSScClhswaV1FCdFsdmcLPsec1-K0Tp50YLMTicYgUNFKFBgeNUt8mNTRAhpbKmUUSmsOeoufC6nzRU5UUi6hdxdcIgg4LRULLxCQKAtyZ80tNqbNbfaoI7JC6QSztdDdZVYdyEUbzUc2mbOhjLlaBrQAG7DACKMAQ0Q48o%2Biqpf1TGazOc1oODZXMm3DBqaUfjgtNIujYsTfTCtqg0QgeYpX0LYlo3d7AeZKzZkOr0P1B0RLcbJtjC-OCauHbUKAQRleAYVlI9Ilc293ZMDE51085tbnDdXTeXc4t7exuACIACQA)

**Finding:** - 4 out of 7 resource types support name search:
- ✅ Models (via `ModelAppliedFilter.identifier`)
- ✅ Sources (via `SourceAppliedFilter.identifier`)
- ✅ Seeds (via `GenericMaterializedFilter.identifier`)
- ✅ Snapshots (via `GenericMaterializedFilter.identifier`)
- ❌ Exposures (no `identifier` field)
- ❌ Tests (no `identifier` field)
- ❌ Metrics (no filter fields)

### Complete Resource Search Coverage

Based on introspection findings:

| Resource Type | Filter Type | Has `identifier`? | Search Support |
|--------------|-------------|-------------------|----------------|
| **Models** | `ModelAppliedFilter` | ✅ Yes | ✅ Implemented |
| **Sources** | `SourceAppliedFilter` | ✅ Yes | ✅ Implemented |
| **Seeds** | `GenericMaterializedFilter` | ✅ Yes | ✅ Implemented |
| **Snapshots** | `GenericMaterializedFilter` | ✅ Yes | ✅ Implemented |
| **Exposures** | `ExposureFilter` | ❌ No | ❌ Requires unique_id |
| **Tests** | `TestAppliedFilter` | ❌ No | ❌ Requires unique_id |
| **Metrics** | `MetricFilter` | ❌ No fields | ❌ Requires unique_id |

---

### Key Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| **Name Resolution** | Internal (resolve inside `get_lineage` tool) | Better UX - single call instead of requiring users to manually resolve names to unique IDs |
| **Search Order** | Models → Sources → Seeds → Snapshots | Optimized by frequency: Models (~90%), Sources (~8%), Seeds (~1%), Snapshots (~1%) - fast path for common cases |
| **Search Implementation** | Sequential search across resource types | No unified search API exists - must query each filter type separately  |
| **Disambiguation** | MCP Elicitation with clarifying fallback prompt| Try interactive selection first using elicitation to get the response from user if the client supports it and, fall back to structured response on timeout/error - guides both LLM and human users |
| **Direction="both"** | Two API calls + merge | No direction metadata in responses - we had to make separate `+uniqueId` and `uniqueId+` calls to categorize ancestors vs descendants |
| **Pagination** | Hard limit: 50 nodes/direction | No server-side pagination available - client-side limit protects LLM token budget |
| **Unsearchable Types** | Clarifying prompt like error messages with examples to guide LLM/User to next steps when users request for non supported resource types| When name search finds no matches, return helpful error that: (1) Lists which types ARE searchable (models, sources, seeds, snapshots), (2) Suggests the user might be searching for an exposure/test/metric, (3) Shows exact `unique_id` format examples for their input (e.g., `exposure.project.{name}`), (4) Explains API limitation so users understand why |

### 1. **Disambiguation Flow** (Which is one of the Important UX Feature in this change). This model was selected because it will trigger eliciation.
**Prompt:**
```
"Get lineage for customers"
```

image

**Use mcp inspector to force elicitation**
image
image

### 2. **Prompt like Error Message** (Resource Type Limitation Handling)
**Prompt:**
```
"Show lineage for customer_dashboard"
```
image

### 3. **Full Lineage** (Feature Completeness)
**Prompt:**
```
"Show full lineage for orders_snapshot"
```
image

## Why

Implementing to close https://github.com/dbt-labs/dbt-mcp/issues/110

## Related Issues

Closes #https://github.com/dbt-labs/dbt-mcp/issues/110
Related to #https://github.com/dbt-labs/dbt-mcp/issues/110

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here https://github.com/dbt-labs/docs.getdbt.com/pull/8233

## Related Issues

Closes #
Related to #

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here https://github.com/dbt-labs/docs.getdbt.com/pull/8233

### DIFF
--- a/src/dbt_mcp/prompts/discovery/get_lineage.md
+++ b/src/dbt_mcp/prompts/discovery/get_lineage.md
@@ -6,9 +6,9 @@ Retrieves lineage for a dbt resource in a single API call. Supports name-based s
 - `types`: Optional filter for specific resource types (Model, Source, Seed, Snapshot, Exposure, Metric, Test, etc.)
 
 **Response:**
-- `target`: The resource queried (matchesMethod=true)
-- `ancestors/descendants`: Dependency nodes (based on direction)
-- `pagination`: Metadata (max 50 nodes per direction, truncation info)
+- `target`: The resource queried (matchesMethod=true).
+- `ancestors/descendants`: Dependency nodes (based on direction).
+- `pagination`: Metadata (max 50 nodes per direction, truncation info).
 
 **Resource Types:**
 Models, Sources, Seeds, Snapshots, Exposures, Metrics, SemanticModels, SavedQueries, Tests, Macros


src/dbt_mcp/discovery/client.py
```diff
@@ -307,6 +308,60 @@ class GraphQLQueries:
         }
     """)
 
+    GET_SEEDS = textwrap.dedent("""
```
DevonFulcher: I just merged a PR that adds GQL queries for seeds & snapshots. Can we reuse those queries here?

src/dbt_mcp/discovery/tools.py
```diff
@@ -175,6 +187,114 @@ async def get_source_details(
...
+    destructive_hint=False,
+    idempotent_hint=True,
+)
+async def get_lineage(
+    context: DiscoveryToolContext,
+    ctx: Context,
+    name: str | None = None,
+    unique_id: str | None = None,
+    direction: str = "both",
+    types: list[str] | None = None,
```
DevonFulcher: Would it be better if these were enums or some other stronger type than strings?

src/dbt_mcp/discovery/tools.py
```diff
@@ -175,6 +187,114 @@ async def get_source_details(
...
+
+                if result.action == "accept":
+                    # Validate the selected unique_id is in matches
+                    selected_id = result.data.unique_id
+                    if selected_id in [m["uniqueId"] for m in matches]:
+                        resolved_unique_id = selected_id
+                    else:
+                        raise InvalidParameterError(
+                            f"Selected unique_id '{selected_id}' not in available matches"
+                        )
```
DevonFulcher: This is pretty interesting! How do you think it compares to the approach I took in my recent changes [here](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/discovery/client.py#L625-L650)?

src/dbt_mcp/discovery/client.py
```diff
@@ -750,3 +934,295 @@ async def fetch_source_details(
...
+
+        # For "both" direction, make two separate API calls to correctly
+        # categorize ancestors and descendants (BUG-001 fix)
+        if direction == LineageDirection.BOTH:
+            ancestors_result = await self._fetch_lineage_single_direction(
+                unique_id, LineageDirection.ANCESTORS, types
+            )
+            descendants_result = await self._fetch_lineage_single_direction(
+                unique_id, LineageDirection.DESCENDANTS, types
+            )
```
DevonFulcher, line 881: Is the reason for the two API calls to simplify labeling ancestors and descendants? The selector syntax allows for both directions [here](https://github.com/dbt-labs/dbt-mcp/pull/461/files#diff-b16884b86662d778854b1c2ae6a357f93184a90fd784c0e1469eaf86bef5fbc5R1079), but I assume the separate calls make it easier to label the nodes. Is that right?

One suggestion, if we keep two separate calls, they can be done concurrently with asyncio.gather().

src/dbt_mcp/discovery/tools.py
```diff
@@ -175,6 +187,114 @@ async def get_source_details(
...
+            # Multiple matches - try elicitation first, fallback to disambiguation
+            try:
+                # Format matches for display
+                match_descriptions = [
+                    f"{m['resourceType']}: {m['uniqueId']}" for m in matches
+                ]
+                message = (
+                    f"Multiple resources found with name '{name}':\n"
+                    + "\n".join(f"  {i+1}. {desc}" for i, desc in enumerate(match_descriptions))
+                    + "\n\nSelect the unique_id of the resource you want:"
```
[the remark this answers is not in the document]
Muizzkolapo: added _format_disambiguation_message

src/dbt_mcp/discovery/client.py
```diff
@@ -750,3 +934,295 @@ async def fetch_source_details(
...
+        config = _RESOURCE_SEARCH_CONFIG[resource_type]
+
+        # Build GraphQL variables
+        variables = {
+            "environmentId": await self.get_environment_id(),
+            config["filter_key"]: {"identifier": name},
+            "first": PAGE_SIZE,
+        }
+
+        # Execute query - already a direct reference to the query string
```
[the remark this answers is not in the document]
Muizzkolapo: Comment clarified added in branch

tests/unit/discovery/test_get_lineage_tool.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+
+    async def test_multiple_matches_returns_disambiguation(self, mock_discovery_context, mock_mcp_context):
+        """Should return disambiguation response when elicitation fails."""
+        mock_discovery_context.lineage_fetcher.search_all_resources.return_value = [
+            {"uniqueId": "model.test.customers", "name": "customers", "resourceType": "Model"},
+            {"uniqueId": "source.test.raw.customers", "name": "customers", "resourceType": "Source"},
+        ]
+        # Simulate elicitation failure
+        mock_mcp_context.elicit.side_effect = Exception("Elicitation not supported")
```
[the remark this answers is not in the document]
Muizzkolapo: done

src/dbt_mcp/discovery/tools.py
```diff
@@ -175,6 +187,114 @@ async def get_source_details(
...
+            except Exception as e:
+                # Elicitation failed, timed out, or not supported - return disambiguation response
+                logger.debug(f"Elicitation not completed: {type(e).__name__}: {e}")
+                return {
+                    "status": "disambiguation_required",
+                    "message": f"Multiple resources found with name '{name}'",
+                    "matches": matches,
+                    "instruction": "Please call get_lineage again with the unique_id parameter set to one of the matches above.",
+                }
+
```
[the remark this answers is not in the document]
Muizzkolapo: done

src/dbt_mcp/discovery/client.py
Muizzkolapo: I like the new GQL folder pattern, I have placed the GQLs in a folder, but I dont think we can reuse the existing GQLs. The reason is because the queries use resources() with AppliedResourcesFilter, which requires knowing the unique_id upfront (via the uniqueIds array). 

The new search gql queries we created for the lineage are lightweight searches using the specific seeds()/snapshots() endpoints with GenericMaterializedFilter's identifier field for name-based matching. 

Through introspection, I noticed that GenericMaterializedFilter has native identifier support for name matching, so we use that for 1-call lookups instead of the package enumeration pattern.

src/dbt_mcp/discovery/tools.py
Muizzkolapo: Good catch! I have Implemented this.

src/dbt_mcp/discovery/client.py
Muizzkolapo: yup! the reason for the two calls is so we can group them into ancestors vs descendants. When using +uniqueId+, the API returns everything mixed together, so we'd need extra logic to figure out which nodes are uancestors vs descendants. I have implemented the asyncio.gather() for parallel execution.

src/dbt_mcp/discovery/tools.py
Muizzkolapo: I actually came up with this pattern after introspection of the Discovery API since ModelAppliedFilter, SourceAppliedFilter, and GenericMaterializedFilter all have an identifier field for name-based searching, I just used that native support instead which means we can just make fewer calls.

src/dbt_mcp/discovery/client.py
```diff
@@ -667,3 +675,303 @@ async def fetch_details(
         if not edges:
             return []
         return [e["node"] for e in edges]
+
+
+# ============================================================================
+# Lineage Tool Classes and Configuration
+# ============================================================================
```
DevonFulcher: Rather than a demarcation like this, should we move this to a different file? It seems we could break this file up a bit more in general.

src/dbt_mcp/discovery/tools.py
```diff
@@ -333,6 +339,145 @@ async def get_test_details(
...
+                    }
+            except Exception as e:
+                # Elicitation failed, timed out, or not supported - return disambiguation response
+                logger.debug(f"Elicitation not completed: {type(e).__name__}: {e}")
+                return {
+                    "status": "disambiguation_required",
+                    "message": f"Multiple resources found with name '{name}'",
+                    "matches": matches,
+                    "instruction": "Please call get_lineage again with the unique_id parameter set to one of the matches above.",
+                }
```
DevonFulcher: I think we should handle this situation in a different way. Not all clients support elicitation, so we should fall back to a different method. Perhaps we should return a list of trees? Additionally, are you familiar with how elicitation is handled with remote MCP? We should ensure this tool is compatible with remote MCP.

src/dbt_mcp/discovery/tools.py
DevonFulcher: I think either approach works well, and I feel okay with trying this out. It will be our first use of elicitation, so that is exciting! It is a creative solution. Thanks for that.

At some point, we should consolidate our approaches, though. I would like the Discovery tools to work uniformly.

Muizzkolapo: hmm yeah definitely, I think we should be able to get elicitation to work with either approaches. Question for you then can we assume the approach here would be the standard going forward https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/discovery/client.py#L625-L650?

DevonFulcher: I think the approach could be: try elicitation. If the client doesn't support elicitation, fall back to the method you linked.

DevonFulcher: My point is that this could be the method for all discovery tools, not just lineage. That uniformity doesn't have to be done in this PR, though.

Muizzkolapo: alright I will just raise a separate issue to standardize the approach for discovery tools

src/dbt_mcp/discovery/tools.py
Muizzkolapo: I think I will remove elicitation for now as I am not sure how it would work in remote mcp, I will create a different issue to look into this approach.

Muizzkolapo:
Hi @DevonFulcher  I just made some changes to align better, I will raise the issues for discovery api uniformity and exploring elicitation after this.

.changes/unreleased/Enhancement or New Feature-20251129-121002.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Adds Get_lineage tool for getting lineage across resources.
```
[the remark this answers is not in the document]
Muizzkolapo: Fixed

src/dbt_mcp/discovery/lineage/tools.py
```diff
@@ -0,0 +1,172 @@
...
+)
+async def get_lineage(
+    context: "DiscoveryToolContext",
+    name: str | None = None,
+    unique_id: str | None = None,
+    direction: str = "both",
+    types: list[str] | None = None,
+) -> dict:
+    normalized_name = name.strip().lower() if name else None
+    normalized_unique_id = unique_id.strip().lower() if unique_id else None
```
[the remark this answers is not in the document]
Muizzkolapo: The Discovery API handles unique IDs case-insensitively so this is fine.

src/dbt_mcp/discovery/client.py
```diff
@@ -667,3 +677,320 @@ async def fetch_details(
...
+
+        config = _RESOURCE_SEARCH_CONFIG[resource_type]
+        environment_id = await self.get_environment_id()
+
+        # Fetch packages to construct possible uniqueIds (Devon's pattern)
+        # Note: Using "model" resource for package fetching works for all types
+        packages_result = await self.api_client.execute_query(
+            ResourceDetailsFetcher.GET_PACKAGES_QUERY,
+            variables={"resource": "model", "environmentId": environment_id},
+        )
```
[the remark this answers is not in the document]
Muizzkolapo: removed comments

tests/unit/discovery/test_lineage_fetcher.py
```diff
@@ -0,0 +1,494 @@
...
+            "model.jaffle_shop.customers", LineageDirection.BOTH
+        )
+        assert result == "+model.jaffle_shop.customers+"
+
+    def test_invalid_direction_raises_error(self, lineage_fetcher):
+        """Should raise ValueError for invalid direction."""
+        with pytest.raises(ValueError, match="Invalid direction"):
+            lineage_fetcher._build_selector(
+                "model.jaffle_shop.customers", "invalid_direction"
+            )
```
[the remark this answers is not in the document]
Muizzkolapo: Added tests for None, int, list, dict, and empty string types

src/dbt_mcp/discovery/lineage/tools.py
```diff
@@ -0,0 +1,172 @@
...
+
+def _fix_lineage_annotations():
+    """Replace string annotations with actual DiscoveryToolContext class."""
+    from dbt_mcp.discovery.tools import DiscoveryToolContext
+
+    # Fix get_lineage annotations
+    get_lineage.fn.__annotations__['context'] = DiscoveryToolContext
+
+    # Fix _fetch_all_lineage_trees annotations
+    _fetch_all_lineage_trees.__annotations__['context'] = DiscoveryToolContext
```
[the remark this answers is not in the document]
Muizzkolapo: Reverted the get_lineage tools to single file approach to avoid this circular dependency, this will be addressed in future discovery API uniformity refactoring.

src/dbt_mcp/discovery/tools.py
DevonFulcher: Sorry, I still see that it is a string type and a list of strings. I think these would be better as an enum and list of enums:
```
    direction: str = "both",
    types: list[str] | None = None,
```

Muizzkolapo: oh sorry forgot that on the refactor, I have switched it back to using enums.

src/dbt_mcp/discovery/tools.py
```diff
@@ -333,6 +340,98 @@ async def get_test_details(
...
+        raise InvalidParameterError("Either name or unique_id must be provided")
+    if normalized_name and normalized_unique_id:
+        raise InvalidParameterError(
+            "Only one of name or unique_id should be provided, not both"
+        )
+
+    resolved_unique_id = normalized_unique_id
+    if not normalized_unique_id:
+        assert normalized_name is not None, "Name must be provided"
+        matches = await context.lineage_fetcher.search_all_resources(normalized_name)
```
DevonFulcher, line 410: Non-blocking: I still think there is greater uniformity we could apply here, such that all of the Discovery tools handle missing `unique_id` in a consistent way, but we can address that later.

PR #462, opened 2025-12-01
Fix semantic layer gql blocking requests

wiggzz:
## Summary

Semantic layer requests are blocking the event loop due to blocking network calls inside async functions.

## What Changed

- Introduce httpx
- Use httpx to make semantic layer requests

## Why

Semantic layer requests are blocking the event loop

## Checklist


wiggzz:
> looks good! it's generally recommended to reuse the `client` since it holds a socket connection pool internally which helps reuse connections instead of having to re-establish a connection on every request.

Yea, I figured as much. This can be a follow-on since it will require a bit more refactoring.

PR #466, opened 2025-12-03
Fix semantic layer client blocking requests

wiggzz:
## Summary

Fix semantic layer blocking requests

## What Changed

- Use `asyncio.to_thread` to run semantic layer client requests 

## Why

Blocking inside an async method will hang the event loop and could prevent other activities from making progress.

## Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20251203-125939.yaml
+++ b/.changes/unreleased/Under the Hood-20251203-125939.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Fix semantic layer client blocking requests
+time: 2025-12-03T12:59:39.630853-06:00


src/dbt_mcp/semantic_layer/client.py
```diff
@@ -352,7 +353,8 @@ async def get_metrics_compiled_sql(
                 parsed_order_by: list[OrderBySpec] = self._get_order_bys(
                     order_by=order_by, metrics=metrics, group_by=group_by
                 )
-                compiled_sql = sl_client.compile_sql(
```
DevonFulcher, line 355: Should we be using the async version of the library instead: https://github.com/dbt-labs/semantic-layer-sdk-python?tab=readme-ov-file#asyncio?

DevonFulcher (approved):
Thank you!

src/dbt_mcp/semantic_layer/client.py
wiggzz: Yes, good idea, I didn't realize that was available. I'll do it in a follow-on because it will require a bigger change and also require changes in codegen.

PR #468, opened 2025-12-04
Upgrade Python MCP SDK

DevonFulcher:
## Summary

Upgrade the Python MCP SDK version to address community issue: https://getdbt.slack.com/archives/C08NGJYD563/p1764769425425089

## Checklist


src/dbt_mcp/proxy/tools.py
```diff
@@ -34,6 +36,28 @@
 logger = logging.getLogger(__name__)
 
 
+# Based on this: https://github.com/modelcontextprotocol/python-sdk/blob/9ae4df85fbab97bf476ddd160b766ca4c208cd13/src/mcp/server/fastmcp/utilities/func_metadata.py#L179
+def _get_typed_annotation(annotation: Any, globalns: dict[str, Any]) -> Any:
```
reneruck-dbt, line 40: did that function get dropped from the lib? 
seems like bits and pieces of func_metadata get dropped and we implement them locally....
Are there no alternatives?

DevonFulcher: Yes, this function was removed. We are using the internals of dbt MCP here, which is fine because we are pinning the minor version. The Python SDK doesn't natively allow for proxying in the way that we depend on, so we have to update the internal methods. I think this is preferable to the alternative, ejecting from the framework entirely. One day, we may decide to upstream these changes.

PR #470, opened 2025-12-04
Fix discovery prompts

wiggzz:
## Summary

Fix discovery prompts

## What Changed

#453 introduced some parameter name changes, ensure that we reflect those in the tool description to avoid confusion

## Why

I don't know what the impact of this will be, but possible mis-use of the tool.

## Checklist


DevonFulcher (approved):
Thank you!

PR #472, opened 2025-12-05
Add x-dbt-partner-source header to discovery tools

DevonFulcher:
## Summary

This helps with telemetry of the Discovery API.


src/dbt_mcp/config/headers.py
```diff
@@ -29,6 +29,7 @@ def headers_from_token(self, token: str) -> dict[str, str]:
         return {
             "Authorization": f"Bearer {token}",
             "Content-Type": "application/json",
+            "X-Dbt-Partner-Source": "dbt-mcp",
```
reneruck-dbt, line 32: aren't we lowercasing header names by default?

DevonFulcher: I don't think we are for direct API calls from dbt MCP to the Discovery API. We may be doing that for our internal use of this tool. Regardless, this casing is how the header appears in the Discovery API source code.

PR #475, opened 2025-12-05
Fix Decimal JSON serialization in semantic layer query results

nustiueudinastea:
DISCLAIMER: this PR was created with the assistance of Claude Code, but I reviewed the small change set. If the project policy is to not accept these please let me know and I will close the PR.

## Summary

Fix JSON serialization error when querying metrics that return Decimal values from the dbt Semantic Layer.

## What Changed
- Added `from decimal import Decimal` import in `src/dbt_mcp/semantic_layer/client.py`
- Renamed `DateTimeEncoder` to `ExtendedJSONEncoder` and extended it to handle `Decimal` objects by converting them to `float`
- Updated the `json.dumps()` call to use the renamed encoder class

## Why
When querying metrics from the dbt Semantic Layer, PyArrow tables with `decimal128`/`decimal256` columns return Python `Decimal` objects via `table.to_pylist()`. The existing `DateTimeEncoder` only handled `datetime` and `date` objects, causing `json.dumps()` to fail with:

```
Object of type Decimal is not JSON serializable
```

MCP response showing the error:
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Object of type Decimal is not JSON serializable"
      }
    ],
    "isError": false
  }
}
```

## Related Issues
Closes #
Related to #

## Checklist
- [ ] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- No documentation changes required

## Additional Notes
Manually verified with a `query_metrics` call that previously failed:

```json
{
  "limit": 3,
  "metrics": ["sqls"],
  "group_by": [{"name": "metric_time", "grain": "MONTH", "type": "time_dimension"}],
  "order_by": [{"name": "metric_time", "descending": true}]
}
```

Now returns successfully:
```json
[
  {"metric_time__month": "2025-12-01T00:00:00+00:00", "sqls": 245.0},
  {"metric_time__month": "2025-11-01T00:00:00+00:00", "sqls": 134.0},
  {"metric_time__month": "2025-10-01T00:00:00+00:00", "sqls": 9373.0}
]
```

### DIFF
--- a/.changes/unreleased/Bug Fix-20251205-114146.yaml
+++ b/.changes/unreleased/Bug Fix-20251205-114146.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Fix JSON serialization error when querying metrics that return Decimal values from the dbt Semantic Layer
+time: 2025-12-05T11:41:46.988675841Z


b-per:
Thanks @nustiueudinastea !

We don't have any problem with people using AI/Claude as long as there is some human analysis/review in the loop, like you did. I am ok with approving the change, but would you be able to add a changelog entry using changie like described [here](https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#changelog)?

nustiueudinastea:
Sure @b-per , let me do that and get back to you.

nustiueudinastea:
@b-per please take a look. I added the change using changie.

b-per (approved):

PR #484, opened 2025-12-11
implement get_lineage tool for complete lineage across dbt resources

Muizzkolapo:
## Summary

A User building a catalog-like experience with dbt and Snowflake via Claude Desktop requested full lineage capabilities https://github.com/dbt-labs/dbt-mcp/issues/110:

> "Get the full lineage of a model, not only its children/parents. To parse the full lineage with the official MCP it would require quite a few calls to the tool to parse through all layers, which would be nice to get the server doing async instead providing the full list as a return."

---

## What Changed

Added a new `get_lineage` tool that enables efficient lineage traversal across all dbt resource types using the Discovery API's `lineage()` endpoint.

### Key Implementation Details

**Query Approach:**
- Uses Discovery API's `lineage()` endpoint to fetch the entire lineage graph in a single API call
- Requests minimal fields: `name`, `uniqueId`, `resourceType`, and `parentIds`
- Pre-builds a children adjacency map for  descendant lookups

**Why Not Use Selector Syntax?**

The  selector parameters (`select: "model+"`, `exclude: "test_*"`)  does not seem to be  available in the public API:

## Test Case
image

Search for any exposures:
image

## Why

Implementing to close https://github.com/dbt-labs/dbt-mcp/issues/110

## Related Issues

Closes #https://github.com/dbt-labs/dbt-mcp/issues/110
Related to #https://github.com/dbt-labs/dbt-mcp/issues/110

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here https://github.com/dbt-labs/docs.getdbt.com/pull/8233

## Related Issues

Closes #
Related to #

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here https://github.com/dbt-labs/docs.getdbt.com/pull/8233

### DIFF
--- a/tests/unit/discovery/test_lineage_fetcher.py
+++ b/tests/unit/discovery/test_lineage_fetcher.py
@@ -11,102 +11,42 @@ def lineage_fetcher(mock_api_client):
     return LineageFetcher(api_client=mock_api_client)
 
 
-@pytest.fixture
-def mock_lineage_response():
-    """Sample lineage response from GraphQL API."""
-    return {
-        "data": {
-            "environment": {
-                "applied": {
-                    "lineage": [
-                        {
-                            "name": "customers",
-                            "uniqueId": "model.test_project.customers",
-                            "resourceType": "Model",
-                            "parentIds": [
-                                "model.test_project.stg_customers",
-                                "source.test_project.raw.customers",
-                            ],
-                        },
-                        {
-                            "name": "stg_customers",
-                            "uniqueId": "model.test_project.stg_customers",
-                            "resourceType": "Model",
-                            "parentIds": ["source.test_project.raw.customers"],
-                        },
-                        {
-                            "name": "customers",
-                            "uniqueId": "source.test_project.raw.customers",
-                            "resourceType": "Source",
-                            "parentIds": [],
-                        },
-                    ]
-                }
-            }
-        }
-    }
-
-
-async def test_fetch_lineage_returns_connected_nodes(
-    lineage_fetcher, mock_api_client, mock_lineage_response
-):
-    """Test fetching lineage returns connected nodes from the API."""
-    mock_api_client.execute_query.return_value = mock_lineage_response
-
-    result = await lineage_fetcher.fetch_lineage(
-        unique_id="model.test_project.customers"
-    )
-
-    # Verify API was called
-    mock_api_client.execute_query.assert_called_once()
-
-    # Verify nodes returned
-    assert len(result) == 3
-    unique_ids = {node["uniqueId"] for node in result}
-    assert "model.test_project.customers" in unique_ids
-    assert "model.test_project.stg_customers" in unique_ids
-    assert "source.test_project.raw.customers" in unique_ids
-
-
-async def test_fetch_lineage_filters_to_connected_subgraph(
-    lineage_fetcher, mock_api_client
-):
-    """Test that lineage returns only nodes connected to the target."""
-    # Mock response with disconnected nodes
+async def test_fetch_lineage_returns_connected_nodes(lineage_fetcher, mock_api_client):
+    """Test that fetch_lineage returns only nodes connected to the target."""
     mock_api_client.execute_query.return_value = {
         "data": {
             "environment": {
                 "applied": {
                     "lineage": [
-                        # Connected to customers
+                        # Connected subgraph
                         {
-                            "name": "customers",
                             "uniqueId": "model.test.customers",
+                            "name": "customers",
                             "resourceType": "Model",
                             "parentIds": ["source.test.raw_customers"],
                         },
                         {
-                            "name": "raw_customers",
                             "uniqueId": "source.test.raw_customers",
+                            "name": "raw_customers",
                             "resourceType": "Source",
                             "parentIds": [],
                         },
                         {
-                            "name": "customer_metrics",
                             "uniqueId": "model.test.customer_metrics",
+                            "name": "customer_metrics",
                             "resourceType": "Model",
                             "parentIds": ["model.test.customers"],
                         },
-                        # NOT connected to customers
+                        # Disconnected subgraph (should be filtered out)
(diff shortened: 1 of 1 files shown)


Muizzkolapo:
Hi @DevonFulcher  for some reason the old branch https://github.com/dbt-labs/dbt-mcp/pull/461  keeps failing the CI checks, I have moved the change onto a new branch and run the precommits locally to confirm all is good. If you can help reapprove this.

src/dbt_mcp/discovery/graphql/get_lineage_details.gql
jairus-m, line 1: Noticed that this was the only `*.gql` file in its own dedicated subdirectory - don't think it matters a whole lot, but want to point out if it's worth considering to keep GraphQL queries flat in `src/dbt_mcp/graphql/` like the others?

src/dbt_mcp/discovery/client.py
```diff
@@ -667,3 +680,640 @@ async def fetch_details(
...
+        "query": GraphQLQueries.GET_SEED_DETAILS,
+        "response_path": "resources",
+        "gql_type": "Seed",
+    },
+    "Snapshot": {
+        "query": GraphQLQueries.GET_SNAPSHOT_DETAILS,
+        "response_path": "resources",
+        "gql_type": "Snapshot",
+    },
+}
```
jairus-m: Similar to `_RESOURCE_QUERY_TEMPLATES` , does this config mapping make sense to be encapsulated within the `LineageFetcher` class as a class var (with clear ownership) rather than at the module-level?

Can also group both class vars at the top together!

tests/unit/discovery/test_lineage_fetcher.py
jairus-m, line 1: Wondering if there's an opportunity to consolidate here! I do acknowledge that this implementation is fairly involved/requires more logic/helper methods than the other fetchers, but when looking at the test structure of the others, most keep the test suite fairly light (4-6 tests) with a focus on high-level behavior (public methods) and less-so on internal implementation details. 

Given this, would it be worth streamlining the approach here? Such that you can:
  - Parameterize the 5 invalid direction tests → 1 test
  - Parameterize the 4 search tests → 1 test
  - Remove tests for private methods (`_build_batch_*`, `_parse_batch_*`)

These are just my thoughts! Would love to hear yours.

src/dbt_mcp/discovery/graphql/get_lineage_details.gql
Muizzkolapo: good catch, that was from the old approach that had multiple gql files, I have moved to the main folder now and renamed the file to better align with the other gql.

tests/unit/discovery/test_lineage_fetcher.py
Muizzkolapo: That is not a bad approach at all. I have parameterized the direction and search tests, but I still want us to keep  the 2 private method tests for _build_batch_node_query and _parse_batch_response because I feel they are the most susceptible to bugs due to the logic.

Muizzkolapo:
> Thank you @Muizzkolapo for letting me take a look :)
> 
> I tested in the MCP inspector and via Claude Desktop successfully! Aside from the minor comments for the code, I did have some observations from using `get_lineage` as a user.
> 
> The first thing I observed, is that when I asked Claude about getting the model lineage for a model, it defaulted to `get_model_lineage_dev` and when it did get a result, it overloaded the context of Claude. I then asked it to get the parents only of the model and Claude then used `get_model_parents`.
> 
> I then explicitly told Claude to use `get_lineage` for parents only and filtered for tests only which did in fact work!
> 
> However, this does have me thinking - what is the best way to properly guide both a user and LLM to use the correct tool?
> 
> All of these are fairly similar:
> 
> * `get_lineage_dev`
> * `get_model_parents`
> * `get_model_children`
> * `get_lineage`
> 
> While there are real differences between these in use and purpose, I can imagine some confusion for users/LLMs for what to use and when, especially for the Discovery API tool set.
> 
> Could there be a future where `get_lineage` is the standard tool for the Discovery API to get lineage ? As in, implement a way to query for "levels" of ancestors and children in addition to what you already have?
> 
> This would make get_model_parents == get_lineage(n=1, direction="ancestors") and get_model_children == get_lineage(n=1, direction="descendants") such that those tools would not be needed. Just a thought!



> Thank you @Muizzkolapo for letting me take a look :)
> 
> I tested in the MCP inspector and via Claude Desktop successfully! Aside from the minor comments for the code, I did have some observations from using `get_lineage` as a user.
> 
> The first thing I observed, is that when I asked Claude about getting the model lineage for a model, it defaulted to `get_model_lineage_dev` and when it did get a result, it overloaded the context of Claude. I then asked it to get the parents only of the model and Claude then used `get_model_parents`.
> 
> I then explicitly told Claude to use `get_lineage` for parents only and filtered for tests only which did in fact work!
> 
> However, this does have me thinking - what is the best way to properly guide both a user and LLM to use the correct tool?
> 
> All of these are fairly similar:
> 
> * `get_lineage_dev`
> * `get_model_parents`
> * `get_model_children`
> * `get_lineage`
> 
> While there are real differences between these in use and purpose, I can imagine some confusion for users/LLMs for what to use and when, especially for the Discovery API tool set.
> 
> Could there be a future where `get_lineage` is the standard tool for the Discovery API to get lineage ? As in, implement a way to query for "levels" of ancestors and children in addition to what you already have?
> 
> This would make get_model_parents == get_lineage(n=1, direction="ancestors") and get_model_children == get_lineage(n=1, direction="descendants") such that those tools would not be needed. Just a thought!

Good idea! I am actually planning a github issue to address this and consolidate the discovery api tool. This was discussed in a previous iteration of this https://github.com/dbt-labs/dbt-mcp/pull/461#discussion_r2585731658

PR #486, opened 2025-12-11
Add Fusion tools to ToolName enum

DevonFulcher:
## Summary

This PR adds remote MCP Fusion tools to the list of available tools. These tools are brand new, but we just missed adding them to the enum immediately.

## Checklist


src/dbt_mcp/config/config.py
```diff
@@ -37,6 +38,7 @@
     Toolset.DISCOVERY: "enable_discovery",
     Toolset.DBT_LSP: "enable_lsp",
     Toolset.SQL: "enable_sql",
+    Toolset.FUSION: "enable_fusion",
```
reneruck-dbt: I think we should really call them LSP tools to avoid confusion

DevonFulcher: This change mirrors our documentation: https://docs.getdbt.com/docs/dbt-ai/about-mcp

DevonFulcher: I moved these tool names under the LSP toolset.

alan-andrade (approved):

PR #492, opened 2025-12-16
Close proxied_tools_manager for exceptions in listing proxied tools

DevonFulcher:
## Summary

I have to do more research on anyio and why this was happening, but first, I want to get this fix out. A combination of factors has led to errors when trying to use Discovery & Semantic Layer tools.

First, we added a requirement that `x-dbt-user-id` header must not be present when service token authentication is used. This change has led to an increase in the number of unauthenticated requests to remote MCP. Proxied tools rely on remote dbt MCP. Typically, if there are failures loading the proxied tools, they are disabled, and this doesn't impact the other tools. With this current bug, that is not the case.

This is because we migrated the Discovery & Semantic Layer tools to use httpx rather than the requests library. For some reason that I still have yet to understand, some underlying cancellation state of anyio caused the Discovery & Semantic Layer httpx tool calls to fail when the API call to list proxied tool calls failed. I will do more research.

This change reinstates the intended behavior, when listing proxied tools fails, the other toolsets are not impacted.

## Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20251216-171641.yaml
+++ b/.changes/unreleased/Under the Hood-20251216-171641.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Migrate admin tools definitions to use the dbt_mcp_tool decorator
+time: 2025-12-16T17:16:41.328774Z
--- a/src/dbt_mcp/dbt_admin/tools.py
+++ b/src/dbt_mcp/dbt_admin/tools.py
@@ -1,4 +1,5 @@
 import logging
+from dataclasses import dataclass
 from typing import Any
 
 from mcp.server.fastmcp import FastMCP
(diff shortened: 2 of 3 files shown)


wiggzz:
Can you link the errors you saw and describe how this fixes it?

wiggzz (approved):

DevonFulcher:
@wiggzz this is the original error message:
```
httpx.HTTPStatusError: Client error '401 Unauthorized' for url 'https://cloud.getdbt.com/api/ai/v1/mcp/'
```
which is being raised [here](https://github.com/dbt-labs/dbt-mcp/blob/16b05ff5b90d7aa183057f419b6563062b0f8b54/src/dbt_mcp/proxy/tools.py#L162) when the remote MCP session is being created to retrieve proxied tools. This is expected because we now prohibit the `x-dbt-user-id` header when a service token is being used. This exception was being caught and didn't provide immediate error messages to the user. However, because we weren't closing the `ProxiedToolsManager` [AsyncExitStack](https://github.com/dbt-labs/dbt-mcp/blob/16b05ff5b90d7aa183057f419b6563062b0f8b54/src/dbt_mcp/proxy/tools.py#L100). This caused issues later on, when the Discovery or Semantic Layer API was called.

From this point, I am still a bit uncertain of the details, but httpx tried to use the shared anyio backend, and, I think, since this stack wasn't closed and the anyio task associated with it wasn't cleanly exited, then this error gets thrown:
```
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

This issue was introduced when we [migrated to using httpx](https://github.com/dbt-labs/dbt-mcp/pull/462/changes) instead of `requests`, and was exacerbated when we changed to prohibit the `x-dbt-user-id` header for service token requests. I believe that was December 10.

PR #494, opened 2025-12-17
Using uv for Python pre-commit hooks

DevonFulcher:
## Summary

uv provides a consistent environment in order to run pre-commit hooks. This change ensures that the hook versions match the versions in pyproject.toml. This should also make CLI path setup easier for developers.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,7 @@ exclude: |
       .mypy_cache/
       | .pytest_cache/
       | .venv/
+      | __pycache__/
   )$
 
 repos:


b-per:
Getting rid of `https://github.com/astral-sh/uv-pre-commit` means that we won't install `uv` automatically and that it will be a requirement for it to be available in the user `PATH`, no?

DevonFulcher:
@b-per `https://github.com/astral-sh/uv-pre-commit` didn't install uv, it checks if `uv.lock` is up-to-date. We maintain that functionality with the `uv lock --check` hook that I added.

b-per:
I think that `https://github.com/astral-sh/uv-pre-commit` actually installs `uv` in the venv used for the pre-commit. To my knowledge, installing a remote hook installs its dependencies in the venv taht pre-commit creates for the check, and the dependency of the astral hook lists `uv`: https://github.com/astral-sh/uv-pre-commit/blob/main/pyproject.toml

I don't think that this is a huge deal, but technically I think that the pre-commit checks before this change didn't require uv to be installed on the machine and in the PATH but the updated config requires it.

DevonFulcher:
@b-per oh got it! Ya, I don't think that is a big concern. We already mention installing `uv` in the CONTRIBUTING.md.

DevonFulcher:
@b-per I think the benefits of this change outweigh that feature of pre-commit and the `uv-pre-commit` hook.

b-per (approved):

PR #518, opened 2026-01-14
Add tool for getting project details

b-per:
## Summary

Add a new admin tool for getting dbt project details.
This can be useful for agents who want to know more about specific failing job runs.

For example, the flow could be

1. admin-tool-call: get the failing runs
2. admin-tool-call:get the information about the project (incl. the repo)
3. check in the repo if anything was changed recently (and clone it if required)
4. create PR in the repo with the fix

## What Changed

Adding a new tool to retrieve the detailed project information

## Checklist

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20260113-153007.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20260113-153007.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Extend detect_binary_type timeout to account for long-running dbt executable
+time: 2026-01-13T15:30:07.124284-08:00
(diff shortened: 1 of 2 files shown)


src/dbt_mcp/dbt_admin/client.py
```diff
@@ -121,6 +121,16 @@ async def get_job_details(self, account_id: int, job_id: int) -> dict[str, Any]:
         )
         return result.get("data", {})
 
+    async def get_project_details(
+        self, account_id: int, project_id: int
+    ) -> dict[str, Any]:
+        """Get details for a specific project."""
+        result = await self._make_request(
+            "GET",
+            f"/api/v2/accounts/{account_id}/projects/{project_id}/",
```
DevonFulcher: Is there a reason to use v2 rather than v3?

b-per: I didn't know it was available in v3!  Many endpoints aren't.
I can update the code and test it tmr!

b-per: Moved to v3 and tested with `mcpc`. OK to approve?

PR #541, opened 2026-01-22
fix: convert HTTPError to ArtifactRetrievalError in get_job_run_artifact

b-per:
When an artifact doesn't exist (e.g., `run_results.json` for a job that failed early), the dbt Cloud API returns a 404 which raises `HTTPError`.

The parser's` _fetch_run_results_artifact` method only catches `ArtifactRetrievalError`, causing the uncaught `HTTPError` to bubble up through asyncio.gather and get silently swallowed, resulting in an empty failed_steps array instead of useful error information.

This fix catches `HTTPError` and converts it to `ArtifactRetrievalError`, allowing the parser to properly fall back to returning truncated logs.

## Checklist


b-per:
I tested it locally and could see useful logs after the change that weren't there before

PR #545, opened 2026-01-23
Align `get_lineage` and `get_model_lineage_dev` signatures. 

### LINKED ISSUE #477: Align get_model_lineage functions.
Non-blocking: I think this should have essentially the same function signature as the other lineage tool: https://github.com/dbt-labs/dbt-mcp/pull/461/files#diff-6d91f0721d8dcde8199de504338811a7063757ec13f32eca508bfbc8b663a54bR390-R396

_Originally posted by @DevonFulcher in https://github.com/dbt-labs/dbt-mcp/pull/420#discussion_r2590967351_

VDFaller:
## Summary
Better alignment between `get_lineage` and `get_model_lineage_dev` tools. 

## What Changed
* Got the signatures to mostly match, (minus disovery context)
* Renamed to `get_lineage_dev` since it's not just models
  * if necessary I can keep but deprecate `get_model_lineage_dev` if we're worried about it. Depends on how many hard coded tool names in the wild there are (like `allowed-tools` in a skill).  
* Made the functionality match better
* Added a shortcut where depth of 0 = 999 depth in discovery, infinite in cli

## Why
#477. 

## Related Issues

Closes #477

## Checklist

## Additional Notes
I created a common fields script @ `src/dbt_mcp/tools/fields.py`. not sure if there's a better place. Also LineageResourceType is in there. 🤮 Suggestions welcome. 
Also, I don't think `semantic_model` and `saved_query` go into the manifest so I'm just not allowing them right now.

### DIFF
--- a/src/dbt_mcp/discovery/tools.py
+++ b/src/dbt_mcp/discovery/tools.py
@@ -2,6 +2,7 @@ import logging
 from dataclasses import dataclass
 
 from mcp.server.fastmcp import FastMCP
+from pydantic import Field
 
 from dbt_mcp.config.config_providers import ConfigProvider, DiscoveryConfig
 from dbt_mcp.discovery.client import (


src/dbt_mcp/dbt_cli/tools.py
```diff
@@ -190,20 +196,17 @@ def _get_manifest() -> Manifest:
...
         return Manifest(**manifest_data)
 
-    def get_model_lineage_dev(
-        model_id: str,
-        direction: Literal["parents", "children", "both"] = "both",
-        exclude_prefixes: tuple[str, ...] = ("test.", "unit_test."),
-        *,
-        recursive: bool,
+    def get_lineage_dev(
+        unique_id: str = UNIQUE_ID_REQUIRED_FIELD,
```
jairus-m, line 200: The tool name itself has changed and also, now there is no model_id/name param (replaced by unique_id). Obviously, a breaking change but wondering what the best path forward for dealing with this/communicating this out?

src/dbt_mcp/prompts/dbt_cli/get_lineage_dev.md
```diff
@@ -0,0 +1,27 @@
...
+<instructions>
+Retrieves the lineage of a specific dbt resource from the local development manifest. Returns both upstream (parents) and downstream (children) dependencies.
+
+This tool ONLY pulls from the local development manifest. For production lineage, use `get_lineage` instead.
+</instructions>
+
+<parameters>
+unique_id: str => Fully-qualified unique ID of the resource (e.g. `model.analytics.stg_orders`).
+types: list[LineageResourceType] | None = None => List of resource types to include. If None, includes all types. Valid types: Model, Source, Seed, Snapshot, Exposure, Metric, Test.
+depth: int = 5 => The depth of the lineage graph to return. Controls how many levels to traverse (1 = immediate neighbors only).
```
jairus-m: Could you include info on depth=0 ?

src/dbt_mcp/dbt_cli/models/lineage_types.py
```diff
@@ -25,48 +63,48 @@ class ModelLineage(BaseModel):
...
         cls,
         manifest: Manifest,
-        model_id: str,
-        direction: Literal["parents", "children", "both"] = "both",
-        exclude_prefixes: tuple[str, ...] = ("test.", "unit_test."),
-        *,
-        recursive: bool = False,
+        unique_id: str,
+        types: list[LineageResourceType] | None = None,
+        depth: int = 5,
```
jairus-m, line 68: For this method, should there be validation or an error raised if depth < 0 is passed?

src/dbt_mcp/tools/fields.py
```diff
@@ -0,0 +1,57 @@
...
+    description="The depth of the lineage graph to return. "
+    "Controls how many levels to traverse from the target node.",
+)
+
+TYPES_FIELD = Field(
+    default=None,
+    description="List of resource types to include in lineage results. "
+    "If not provided, includes all types. "
+    "Valid types: Model, Source, Seed, Snapshot, Exposure, Metric, SemanticModel, SavedQuery, Test.",
+)
```
jairus-m, line 42: I know you mentioned in PR that SemanticModel and SavedQuery are not valid resources for CLI and also block them later on in `lineage_types.py` , so just wanted to flag that maybe worth nothing that these are only available for the dbt Platform users that go through the Discovery API?

src/dbt_mcp/dbt_cli/tools.py
VDFaller: Since the call comes from an agent, I'm not personally worried about a breaking change as long as the documentation to the agent matches the new stuff.  I figure it's going to do a dynamic call based on what's in the doc for it.  Thoughts?

src/dbt_mcp/tools/fields.py
VDFaller: I put the check into the cli with an error message telling the agent that those two aren't allowed.  

And if you do `None` for type it doesn't break.  It will just [return True](https://github.com/dbt-labs/dbt-mcp/pull/545/files#diff-509ea69f08f6e051651fe2ecbe1aa420c836a5bbede45c8887ff493492771a95R43).  

When I tested it, the only way to even get the error was to ask for semantic models explicitly, which I think makes sense for it to error out.

src/dbt_mcp/dbt_cli/tools.py
jairus-m: Totally hear you on that. The thing I'm thinking about kinda refers to users who may have experienced [this](https://github.com/dbt-labs/dbt-mcp/pull/484#pullrequestreview-3614869230) and have needed to use either the `DISABLE_TOOLS` or `DBT_MCP_ENABLE_TOOLS` env vars.

If someone had 

```bash
DISABLE_TOOLS=get_model_lineage_dev
```

in order to only use the Platform discover tool `get_lineage` when using the MCP, that may be enough of a hurdle to consider since their setup will now silently change in behavior. 

I'll leave this thought here for now and welcome other POVs - I ack that it may not be that much of a friction point (and do see that tool use is dynamically called) but wanted to flag it!

VDFaller: I'm happy to change it back if we don't want to deal with it.  I made it its own commit knowing this might happen.

jairus-m: No worries! We can see what others think - just wanted to make sure we talk about it since this may be the first time a tool name would be changed (?). I believe it will be good to get alignment on how we think about and navigate deprecations 👍 (it may have already been discussed, but not sure aha)

VDFaller:
@jairus-m You think I making a `src/dbt_mcp/common/` folder for things like `fields` and ... `parameters` ... (still haven't found a name I like) makes enough sense?

jairus-m:
>You think I making a src/dbt_mcp/common/ folder for things like fields and ... parameters ... (still haven't found a name I like) makes enough sense?

this may be just personal preference but i feel like `common/` is too broad (also wondering how common it will be to have shared params outside of local CLI vs discovery-specific tools in the future). i think if you asked for my strict opinion, since `src/dbt_mcp/tools/fields.py` is specifically shared tool params between the CLI / discovery get_lineage tools, id personally just leave the module there but rename it to `src/dbt_mcp/tools/parameters.py`. but again, just take this as my 2c's !

.changes/unreleased/Breaking Change-20260124-205900.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Breaking Change
```
DevonFulcher: I don't think we should submit this as a breaking change. I updated the guidelines here with what I think the policy should be: https://github.com/dbt-labs/dbt-mcp/pull/566.

Thanks for bringing this up! Especially regarding the `DBT_MCP_ENABLE_TOOLS` configuration. We need to make a fix so there are no breaking changes. I'm working on that.

VDFaller:
Cool.  Following #567 and will fix everything once that's in.  Thanks @DevonFulcher

src/dbt_mcp/dbt_cli/models/lineage_types.py
```diff
@@ -1,9 +1,47 @@
...
+    LineageResourceType.SEMANTIC_MODEL,
+    LineageResourceType.SAVED_QUERY,
+}
+
+
+def _resource_types_to_prefix(resource: LineageResourceType) -> tuple[str, ...]:
+    """Convert LineageResourceType to tuple of unique_id prefixes."""
+    if resource in UNALLOWED_RESOURCE_TYPES:
+        raise ValueError(
+            f"Resource types '{', '.join(UNALLOWED_RESOURCE_TYPES)}' are not supported for lineage in dbt CLI."
```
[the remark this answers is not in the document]
VDFaller: Not True. It's a StrEnum.

VDFaller:
Should be good.

src/dbt_mcp/discovery/tools.py
```diff
@@ -2,14 +2,12 @@
 from dataclasses import dataclass
 
 from mcp.server.fastmcp import FastMCP
-from pydantic import Field
```
jairus-m, line 5: @VDFaller i checked out your branch and was going to test locally, but came across an error on server startup. I then ran the unit tests and there were some failures! Just flagging since CI checks skipped the tests.

```bash
ERROR tests/unit/discovery/test_get_model_performance.py - NameError: name 'Field' is not defined
ERROR tests/unit/lsp/test_lsp_tools.py - NameError: name 'Field' is not defined
ERROR tests/unit/tools/test_disable_tools.py - NameError: name 'Field' is not defined
ERROR tests/unit/tools/test_tool_names.py - NameError: name 'Field' is not defined
ERROR tests/unit/tools/test_toolsets.py - NameError: name 'Field' is not defined
```

`Field` is used in `get_model_performance` (lines 190 and 197)

VDFaller: Yeah, I added it back in.  My bad.  Not sure how that got past pre-commit.

PR #548, opened 2026-01-26
feat: add automatic OAuth token refresh at startup

b-per:
## Summary
- Automatically refresh OAuth access tokens at startup using the refresh token
- Eliminates unnecessary re-authentication flows after token expiry (~1 hour)

## Problem
When the MCP server restarts after the access token has expired, users were required to go through the full OAuth login flow again, even though they had a valid refresh token saved.

## Solution
Before triggering a full OAuth flow, attempt to use the refresh token to obtain a new access token:
1. Check if context is complete (has all required fields)
2. If token is still valid, use it directly
3. If token is expired, try to refresh using the refresh token
4. Only fall back to full OAuth login if refresh fails

## Test plan
- [x] New unit tests for `_is_context_complete`, `_is_token_valid`, `_try_refresh_token`
- [x] Existing OAuth tests pass

### DIFF
--- a/src/dbt_mcp/config/settings.py
+++ b/src/dbt_mcp/config/settings.py
@@ -232,9 +232,9 @@ class DbtMcpSettings(BaseSettings):
         if v in ["dbt", "dbtf"]:
             return v
         if v:
-            p = Path(v)
+            p = Path(v).expanduser()
             if p.exists():
-                return v
+                return str(p)
 
             field_name = (
                 getattr(info, "field_name", "None") if info is not None else "None"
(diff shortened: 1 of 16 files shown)


src/dbt_mcp/config/settings.py
```diff
@@ -359,18 +428,21 @@ async def get_dbt_platform_context(
...
+            assert dbt_ctx is not None  # for type checker
+            # If token is still valid, use context directly
+            if _is_token_valid(dbt_ctx):
+                return dbt_ctx
+            # Token expired, try to refresh
+            refreshed_ctx = _try_refresh_token(
+                dbt_ctx, dbt_platform_url, dbt_platform_context_manager
+            )
+            if refreshed_ctx:
+                return refreshed_ctx
```
DevonFulcher, line 444: I was under the assumption that the token can only be refreshed before it expires. Is that not the case?

b-per: The access token and refresh token both expire, but they have different expiration time (the refresh one expiring later).

So, it is possible for the access one to be expired but not the refresh one yet.

PR #549, opened 2026-01-26
feat: add production environment picker during OAuth setup

b-per:
## Summary
- Add ability to select production environment from a dropdown during OAuth setup instead of relying on auto-detection
- Dropdown shows all non-development environments with name, deployment type, and ID
- Production environments are pre-selected by default if available
- Added new `/environments` endpoint to fetch available environments for a project

### DIFF
--- a/src/dbt_mcp/oauth/fastapi_app.py
+++ b/src/dbt_mcp/oauth/fastapi_app.py
@@ -224,10 +224,10 @@ def create_app(
         return dbt_platform_context_manager.read_context() or DbtPlatformContext()
 
     @app.post("/environments")
-    def get_environments(
+    def get_deployment_environments(
         request: GetEnvironmentsRequest,
     ) -> list[DbtPlatformEnvironmentResponse]:
-        """Get all environments for a project, excluding development environments."""
+        """Get all deployment environments for a project, excluding development environments."""
         logger.info("Getting environments for project %s", request.project_id)
         if app.state.decoded_access_token is None:
             raise RuntimeError("Access token missing; OAuth flow not completed")
@@ -247,9 +247,7 @@ def create_app(
         return [
             env
             for env in environments
-            if not (
-                env.deployment_type and env.deployment_type.lower() == "development"
-            )
+            if not (env.type and env.type.lower() == "development")
         ]
 
     @app.post("/selected_project")
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -614,7 +614,7 @@ export default function App() {
                       className="form-label"
                       id="environment-select-label"
                     >
-                      Production Environment
+                      Deployment Environment
                     </label>
                     <select
                       id="environment-select"
@@ -624,7 +624,7 @@ export default function App() {
                     >
                       {environments.map((env) => (
                         <option key={env.id} value={env.id}>
-                          {env.name} ({env.deployment_type || "unknown"}, {env.id})
+                          {env.name} ({env.deployment_type ? `${env.deployment_type} - ${env.id}` : env.id})
                         </option>
                       ))}
                     </select>
@@ -687,7 +687,7 @@ export default function App() {
 
               {dbtPlatformContext.prod_environment && (
                 <div className="context-item">
-                  <strong>Production Environment:</strong>
+                  <strong>Deployment Environment:</strong>
                   <div className="environment-details">
                     <span className="env-name">
                       {dbtPlatformContext.prod_environment.name}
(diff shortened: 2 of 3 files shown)


b-per:
Updated UI with the selector

image

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -222,6 +223,35 @@
         logger.info("Selected project received")
         return dbt_platform_context_manager.read_context() or DbtPlatformContext()
 
+    @app.post("/environments")
+    def get_environments(
+        request: GetEnvironmentsRequest,
+    ) -> list[DbtPlatformEnvironmentResponse]:
+        """Get all environments for a project, excluding development environments."""
+        logger.info(f"Getting environments for project {request.project_id}")
```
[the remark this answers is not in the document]
b-per: Good bot! But Opus suggested a better fix

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -222,6 +223,35 @@ def get_dbt_platform_context() -> DbtPlatformContext:
         logger.info("Selected project received")
         return dbt_platform_context_manager.read_context() or DbtPlatformContext()
 
+    @app.post("/environments")
+    def get_environments(
```
DevonFulcher:
```suggestion
    def get_production_environments(
```

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -222,6 +223,35 @@ def get_dbt_platform_context() -> DbtPlatformContext:
...
+            project_id=request.project_id,
+            headers=headers,
+            page_size=100,
+        )
+        # Filter out development environments since this is for selecting production
+        return [
+            env
+            for env in environments
+            if not (
+                env.deployment_type and env.deployment_type.lower() == "development"
```
DevonFulcher: This filter is a bit confusing to me because the image you posted in the PR displays a `Development` environment. Can users select a development environment?

src/dbt_mcp/oauth/fastapi_app.py
DevonFulcher: Actually, it seems we should start using the term "deployment":

```suggestion
    def get_deployment_environments(
```

b-per:
Updating the code and Dev is gone
image

src/dbt_mcp/oauth/fastapi_app.py
b-per: Fixed

src/dbt_mcp/oauth/fastapi_app.py
```diff
@@ -222,6 +223,33 @@ def get_dbt_platform_context() -> DbtPlatformContext:
...
+            account_id=request.account_id,
+            project_id=request.project_id,
+            headers=headers,
+            page_size=100,
+        )
+        # Filter out development environments since this is for selecting production
+        return [
+            env
+            for env in environments
+            if not (env.type and env.type.lower() == "development")
```
DevonFulcher, line 250: Should we also filter out staging-type environments?

ui/src/App.tsx
```diff
@@ -522,6 +589,74 @@ export default function App() {
...
+                !environmentsError &&
+                selectedProjectId !== null &&
+                environments.length > 0 && (
+                  <div className="form-group">
+                    <label
+                      htmlFor="environment-select"
+                      className="form-label"
+                      id="environment-select-label"
+                    >
+                      Production Environment
```
DevonFulcher: This is a list of deployment environment's, right? Not necessarily production environments
```suggestion
                      Deployment Environment
```

There are some other places in this PR where "production" is used instead of "deployment". Ideally, we are consistent codebase-wide, but especially so in this PR.

src/dbt_mcp/oauth/fastapi_app.py
b-per: No, it is OK to connect the MCP server to a staging env if some people want to query the SL or the Discovery API on staging.

ui/src/App.tsx
b-per: Fixed in this PR. We might want to do a bigger refactor later to remove most mentions of PROD in the existing codebase

DevonFulcher: Ya, I agree. It also seems like "deployment" may not suffice since it doesn't encapsulate staging? It is okay for now, but perhaps there is a better term for it.

PR #550, opened 2026-01-26
feat: add Dockerfile for containerized deployment

b-per:
## Summary

Add Dockerfile and .dockerignore to enable running dbt-mcp in a Docker container.

This is a prerequisite for listing dbt-mcp on the [Docker MCP Hub](https://hub.docker.com/mcp).

## Changes

- **Dockerfile**: Multi-stage build using uv for fast dependency installation
  - Uses `python:3.12-slim-bookworm` as base
  - Runs as non-root user (`appuser`) for security
  - Supports `SETUPTOOLS_SCM_PRETEND_VERSION` build arg for versioning without .git
  - Default entry point is `dbt-mcp` command

- **.dockerignore**: Excludes unnecessary files from build context (tests, docs, .env, IDE files, etc.)

## Build & Test

```bash
# Build the image
DOCKER_BUILDKIT=1 docker build \
  --build-arg SETUPTOOLS_SCM_PRETEND_VERSION=1.7.0 \
  -t dbt-mcp:test .

# Test it runs
docker run --rm dbt-mcp:test

# Verify version
docker run --rm --entrypoint python dbt-mcp:test \
  -c "from importlib.metadata import version; print(version('dbt-mcp'))"
```

## Notes

- OAuth flow is not supported in containers (requires browser) - users should provide `DBT_TOKEN` (service token) instead
- CLI tools are not included (require local dbt project and dbt executable)
- Platform features (Semantic Layer, Discovery, Admin API, SQL) work with service token authentication


DevonFulcher:
Nice! Should we add instructions in CONTRIBUTING.md for building & running this? Should we add steps to CI to build and run it to ensure it works correctly? Do we need CI steps to publish this to MCP hub for new release versions?

b-per:
Docker needs us to have a Dockerfile in the root, but then, they are the ones building images with this info.

I wouldn't expect many folks using it themselves, but this is a requirement to be part of the Docker Hub and to be added to Docker Desktop easily.

jairus-m:
@b-per can you help me understand how this gets used downstream? I checked out your branch and ran the test command which worked great. But I'm curious, how would people control env vars for tool control with this deployment?

Would it be as simple as having this MCP available in the docker MCP hub and then running a command and passing vars in-line or referencing a .env file?

Like the following:

```json
"mcpServers": {
  "dbt-mcp": {
    "command": "docker",
    "args": [
      "run", 
      "-i", 
      "--rm", 
      "--env-file", "abs/path/to/.env", 
      "{dbtlabs-dockerhub}/dbt-mcp:latest"
    ]
  }
}
```

b-per:
Again, this is not so much for "end users" consumption.
This is the format requested by the Docker MCP Hub/Catalog: https://github.com/docker/mcp-registry/blob/main/CONTRIBUTING.md

They then build some UI on top of it when accessed via Docker Desktop

jairus-m:
Ah my bad. Misread and misunderstood. Thanks for the clarification!

PR #565, opened 2026-01-28
Add hatchling build hook to build UI during package installation

### LINKED ISSUE #560: [Bug]: Can't call `uvx --env-file .env git+https://github.com/dbt-labs/dbt-mcp@main` with OAuth due to ui files built at release time
### What happened?

I tried to run dbt-mcp from `@main` on github, with OAuth and got an error due to the fact that the `ui` file are not stored in GitHub but are generated during the release process via `task build`.

Allowing dbt-mcp from GitHub can be a great way sometimes to test code that has been merged but not yet released to PyPi.

If we wanted to keep supporting this approach we would need to either:
1. add the ui files to the repo
    - as the js/css files contain hashes, this would lead to deleted/new files every time we change the UI
    - we could hard code the js/css file names but in that case we might have some issues with cache invalidation not working for those when we update the UI
2. or add a hatchling build hook that would run `pnpm install` and `pnpm build`
    - we could then remove `(cd ui && pnpm install && pnpm build)` from the build task
    - but this would require for users to have `pmpm` installed on their machine when building from GitHub

I think that we could potentially go with either options (a slight preference for option 2), but what do people think?

### Steps to Reproduce

1. set an env var file with just the DBT_HOST
2. run `uvx --env-file .env git+https://github.com/dbt-labs/dbt-mcp@main`

### Deployment

Local MCP server

### Environment

_No response_

### Relevant log output

```shell

```

b-per:
## Summary

Fixes #560

This PR adds a hatchling build hook that automatically builds the UI assets during package installation, enabling users to install dbt-mcp directly from GitHub:

```bash
uvx git+https://github.com/dbt-labs/dbt-mcp@main
```

## Changes

- Add `hatch_build.py` with `UIBuildHook` that runs `pnpm install` and `pnpm build`
- Configure the hook in `pyproject.toml` for both wheel and sdist targets
- Include `ui/` source files in sdist (excluding `node_modules`)
- Remove manual UI build from Taskfile `build` task (hook handles it now)
- Fix `pnpm-workspace.yaml` to include `packages` field required by pnpm v10

## Notes

- Users installing from GitHub will need `pnpm` installed (as discussed in the issue, this is an acceptable tradeoff)
- GitHub Actions already has pnpm setup via `pnpm/action-setup`
- The hook provides clear error messages if pnpm is not available

## Test plan

- [x] Verified `uv build` works and triggers the hook
- [x] Verified sdist includes UI source files but excludes `node_modules`

### DIFF
--- a/hatch_build.py
+++ b/hatch_build.py
@@ -1,0 +1,66 @@
+"""Hatchling build hook to build the UI before packaging."""
+
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface  # type: ignore[import-not-found]
+
+
+class UIBuildHook(BuildHookInterface):
+    """Build hook that builds the UI assets before packaging."""
+
+    PLUGIN_NAME = "ui-build"
+
+    def initialize(self, version: str, build_data: dict) -> None:
+        """Build the UI assets before the package is built."""
+        if self.target_name not in ("wheel", "sdist"):
+            return
+
+        root = Path(self.root)
+        ui_dir = root / "ui"
+
+        if not ui_dir.exists():
+            self.app.display_warning(f"UI directory not found: {ui_dir}")
+            return
+
+        # Check if pnpm is available
+        pnpm_cmd = shutil.which("pnpm")
+        if not pnpm_cmd:
+            self.app.display_error(
+                "pnpm is not installed. Please install pnpm to build the UI. "
+                "You can install it with: npm install -g pnpm"
+            )
+            sys.exit(1)
+
+        self.app.display_info("Building UI assets...")
+
+        try:
+            # Run pnpm install
+            self.app.display_info("Running pnpm install...")
+            subprocess.run(
+                [pnpm_cmd, "install"],
+                cwd=ui_dir,
+                check=True,
+                capture_output=True,
+                text=True,
+            )
+
+            # Run pnpm build
+            self.app.display_info("Running pnpm build...")
+            subprocess.run(
+                [pnpm_cmd, "build"],
+                cwd=ui_dir,
+                check=True,
+                capture_output=True,
+                text=True,
+            )
+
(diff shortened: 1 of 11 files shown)


hatch_build.py
```diff
@@ -0,0 +1,66 @@
...
+
+    PLUGIN_NAME = "ui-build"
+
+    def initialize(self, version: str, build_data: dict) -> None:
+        """Build the UI assets before the package is built."""
+        if self.target_name not in ("wheel", "sdist"):
+            return
+
+        root = Path(self.root)
+        ui_dir = root / "ui"
```
DevonFulcher, line 22: Would it be better to keep this path the same as before, `src/dbt_mcp/ui`? The app expects that specific path, [here](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/oauth/login.py#L52).

pyproject.toml
```diff
@@ -71,7 +71,14 @@ requires = ["hatchling", "hatch-vcs"]
...
 [tool.hatch.build.targets.sdist]
-include = ["src/dbt_mcp/**/*", "README.md", "LICENSE"]
+include = ["src/dbt_mcp/**/*", "README.md", "LICENSE", "ui/**/*"]
+exclude = ["ui/node_modules/**/*"]
+
+[tool.hatch.build.targets.wheel.hooks.custom]
+path = "hatch_build.py"
+
+[tool.hatch.build.targets.sdist.hooks.custom]
+path = "hatch_build.py"
```
jairus-m, line 81: I've never worked w/ build hooks before so was generally curious about the behavior as well as the diff between sdist and wheel distributions. 

I ran `uv build -v` locally and saw that the "pnpm install + pnpm build" step ran twice. Is this intended?

From my *new* understanding, the wheel dist is the built package that gets directly installed and used. The source dist is one step before that (has raw src files and no build assets) and is the basis for then building a wheel. So given this, is the hook on sdist necessary? Or would it make sense to just keep it on the wheel target?

PR #568, opened 2026-01-28
Enhance docs gen script to include d2.png updates / proper MCP casing

jairus-m:
## Summary

Enhance `docs/generate_docs.py` to include d2.png updates / proper MCP casing

## What Changed

- "Mcp" --> "MCP" in docs (automatic)
- Include update to `docs/d2.png` IFF `docs/diagram.d2` is automatically updated
  - Prior behavior, a new tool set would automatically populate in the `diagram.d2` but no updates to `d2.png`
  - Now,  both `pre-commit` and `task docs:generate` handle the update of `d2.png` with the addition of new tool sets to the `Toolset` enum

## Why

- Styling/casing of MCP
- Improve automation of docs

## Checklist


scripts/generate_docs.py
```diff
@@ -164,16 +216,21 @@ def update_all(check_only: bool = False) -> bool:
         True if all docs are up to date (or were updated), False if updates needed.
     """
     readme_ok = update_readme(check_only)
-    diagram_ok = update_d2_diagram(check_only)
+    diagram_ok, diagram_changed = update_d2_diagram(check_only)
```
b-per, line 219: Should we use `task` instead for that? 
It uses a cache mechanism that "skips" the d2 generation if the `.d2` file hasn't changed.

d2 can be a bit long to run sometimes

PR #570, opened 2026-01-29
Make ToolDefinition.name required to eliminate fallback checks

alan-andrade:
## Summary
Client code that makes use of tool names is forces to check where the name lives with this check:

`tool.name or tool.fn.__name__`

This update is to clean up code. To access the tool name just do `tool.name`.

## What Changed
Set tool definition name at creation time and update type to have the name be non-nullable.

## Why
For code cleanup purposes

## Checklist

## Additional Notes

The tradeoff is that the name is explicit when defining `ToolDefinition` which I think is better than an IF check.


DevonFulcher:
Thanks Alan! However, I find this a bit redundant. I would prefer not to repeat the function name as a string in tools. Would it not suffice to use `get_name` when a non-nullable value is required?

PR #575, opened 2026-02-02
feat: auto-detect DBT_PATH and DBT_PROJECT_DIR when not set

joellabes:
## Summary

- When `DBT_PATH` is not set, automatically find dbt using `which dbt` or `which dbtf`
- When `DBT_PROJECT_DIR` is not set, use the current working directory if it's a valid dbt project
- Validates the project directory by running `dbt parse` and checking for "not a project" errors

## Details

The `dbt parse` validation:
- Terminates early once success or failure is determined (doesn't wait for full parse)
- Only fails for "not a project" errors (e.g., "no dbt_project.yml found", "not a dbt project dir")
- Other errors (syntax errors, missing dependencies, etc.) still indicate a valid project directory
- Has a 30-second timeout to avoid blocking startup
- Works with `dbt_project.yml` in parent directories (dbt handles this automatically)

If auto-detection fails, the existing `auto_disable` logic kicks in to disable CLI features gracefully.

## Test plan

- [x] Existing unit tests pass (358 passed)
- [x] Manual test: Run MCP server from a dbt project directory without setting env vars
```
                                                                                                                                                                                                                                                   
❯ which dbt models are here                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                   
⏺ dbt-mcp-local - dbt list (MCP)(resource_type: ["model"])                                                                                                                                                                                         
  ⎿  =======================Selected nodes=======================                                                                                                                                                                                  
     jaffle_shop.marts.customers                                                                                                                                                                                                                   
     jaffle_shop.daily_active_users                                                                                                                                                                                                                
     … +18 lines (ctrl+o to expand)                                                                                                                                                                                                                
                                                                                                                                                                                                                                                   
⏺ This project has 15 dbt models organized into layers:                                                                                                                                                                                            
                                                                                                                                                                                                                                                   
  Staging models:                                                                                                                                                                                                                                  
  - stg_customers                                                                                                                                                                                                                                  
  - stg_locations                                                                                                                                                                                                                                  
  - stg_order_items                                                                                                                                                                                                                                
  - stg_orders                                                                                                                                                                                                                                     
  - stg_products                                                                                                                                                                                                                                   
  - stg_supplies                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                   
  Marts models:                                                                                                                                                                                                                                    
  - customers                                                                                                                                                                                                                                      
  - locations                                                                                                                                                                                                                                      
  - metricflow_time_spine                                                                                                                                                                                                                          
  - order_items                                                                                                                                                                                                                                    
  - orders                                                                                                                                                                                                                                         
  - products                                                                                                                                                                                                                                       
  - supplies                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                   
  Other models (root level):                                                                                                                                                                                                                       
  - daily_active_users                                                                                                                                                                                                                             
  - subscriptions                                                                                                                                                                                                                                  

```
- [x] Manual test: Run MCP server from a non-dbt directory and verify CLI features are disabled
```
❯ which dbt models are in this project                                                                                                                                                                                                             
                                                                                                                                                                                                                                                   
⏺ dbt-mcp-local - dbt list (MCP)(resource_type: ["model"])                                                                                                                                                                                         
  ⎿  error: dbt1001: The current dir is not a dbt project dir                                                                                                                                                                                      
     Hint: `cd` into it; pass a <path> to it via `--project-dir <path>`; or run `db init` to scaffold a project                                                                                                                                    
                                                                                                                                                                                                                                                   
⏺ The dbt MCP server isn't configured with a dbt project directory. Let me check if there's a dbt project within this repository.                                                                                                                  
                                                                                                                                                                                                                                                   
⏺ Searched for 1 pattern (ctrl+o to expand)                                                                                                                                                                                                        
                                                                                                                                                                                                                                                   
⏺ This repository (dbt-mcp) is not a dbt project itself - it's the Python codebase for the dbt MCP server. There's no dbt_project.yml file here.                                                                                                   
                                                                                                                                                                                                                                                   
  To list dbt models, you would need to:                                                                                                                                                                                                           
                                                                                                                                                                                                                                                   
  1. Configure the dbt MCP server to point to an actual dbt project by setting:                                                                                                                                                                    
    - DBT_PROJECT_DIR - path to your dbt project                                                                                                                                                                                                   
    - DBT_PATH - path to the dbt executable (or let it auto-detect based on your recent commit)                                                                                                                                                    
  2. Or navigate to a dbt project directory before using the dbt MCP tools                                                                                                                                                                         
                                                                                                                                                                                                                                                   
  Do you have a dbt project you'd like to connect to, or were you asking about something else?                                                                                                                                                     
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/dbt_mcp/config/settings.py
+++ b/src/dbt_mcp/config/settings.py
@@ -296,9 +296,7 @@ class DbtMcpSettings(BaseSettings):
             resolved_dir = _resolve_project_dir_default(resolved_dbt_path)
             if resolved_dir:
                 object.__setattr__(self, "dbt_project_dir", resolved_dir)
-                logger.info(
-                    f"Auto-detected dbt project directory: {resolved_dir}"
-                )
+                logger.info(f"Auto-detected dbt project directory: {resolved_dir}")
 
         return self
 
@@ -637,7 +635,10 @@ def _validate_dbt_project_with_parse(
                 # Use select on Unix to avoid blocking indefinitely
                 if sys.platform != "win32":
                     ready, _, _ = select.select(
-                        [process.stdout], [], [], 1.0  # type: ignore[list-item]
+                        [process.stdout],
+                        [],
+                        [],
+                        1.0,  # type: ignore[list-item]
                     )
                     if not ready:
                         continue


src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+def _resolve_dbt_path_default() -> str | None:
+    """
+    Attempt to find a dbt executable on the system PATH.
+
+    Returns:
+        The full path to the dbt executable, or None if not found
+    """
+    # Try common dbt executable names
+    for executable in ["dbt", "dbtf"]:
+        resolved = shutil.which(executable)
```
b-per, line 725: Do we know what happens with `shutil.which` with different MCP clients?

I guess that for Claude Code it uses whatever $PATH and/or venv is activated when we run `claude code` ; but what happens for Claude Desktop for example?

src/dbt_mcp/config/settings.py
```diff
@@ -266,6 +270,36 @@ def parse_disable_tools(cls, env_var: str | None) -> list[ToolName] | None:
...
+    def apply_cli_defaults(self) -> "DbtMcpSettings":
+        """Apply sensible defaults for dbt CLI settings when env vars are not set.
+
+        If DBT_PATH is not set (or is the default "dbt"), attempt to find dbt on PATH.
+        If DBT_PROJECT_DIR is not set, attempt to use the current working directory
+        if it appears to be a valid dbt project (validated with `dbt parse`).
+        """
+        # Skip if CLI is explicitly disabled
+        if self.disable_dbt_cli:
+            return self
```
b-per, line 283: This would be a behavior change.

- Before this change, if `DBT_PATH` or `DBT_PROJECT_DIR` are not set, then all the CLI tools are disabled.  
- After this update, if people were running the MCP server from a dbt project, the CLI tools would be activated. 

It feels like a better experience, but it is a change that we might want to highlight as it might make current production workflow behave worst with the LLM trying to call the CLI tools instead of the other tools that were configured. The impact is pretty low, but it is there.

src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+                    ready, _, _ = select.select(
+                        [process.stdout],
+                        [],
+                        [],
+                        1.0,  # type: ignore[list-item]
+                    )
+                    if not ready:
+                        continue
+
+                line = process.stdout.readline()  # type: ignore[union-attr]
```
b-per, line 646: Asked Claude Code for a review of this function and it highlighted this line:

> On Windows, select doesn't work on pipes, so the code skips it and falls through to readline() which can block forever if dbt produces no
>   output. This could hang the MCP server startup on Windows.
> 
>   Fix: Use threading with a timeout, or use asyncio.subprocess with proper timeout handling for both platforms.

src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+                # Note: "Running with dbt=" is printed before validation, so don't use it
+                if "found" in line_lower and any(
+                    word in line_lower
+                    for word in ["model", "source", "test", "seed", "snapshot"]
+                ):
+                    process.terminate()
+                    process.wait(timeout=5)
+                    return True
+
+        except Exception:
```
b-per, line 677: Can we add logging details here to help debugging? 
This PR is doing it in another part of the code.

src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+
+                # Check for specific "not a project" indicators only
+                # dbt Core: "No dbt_project.yml found" or similar
+                # dbt Fusion: "not a dbt project" or "is not a dbt project dir"
+                # These indicate the directory itself is wrong, not just a parsing error
+                if (
+                    "no dbt_project.yml" in line_lower
+                    or "not a dbt project" in line_lower
+                    or "is not a dbt project" in line_lower
+                ):
```
b-per, line 661: This feels quite brittle, no?
Has it been checked to be working for Core, Fusion and the Cloud CLI?

If we keep this logic let's add as many tests as possible then

src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+
+    except FileNotFoundError:
+        # dbt executable not found
+        return False
+    except Exception as e:
+        logger.debug(f"Error validating dbt project with parse: {e}")
+        return False
+
+
+def _resolve_dbt_path_default() -> str | None:
```
DevonFulcher, line 716: We could also check the dbt project for a venv or look in the default Fusion installation path, which I think is `~/.local/bin/dbt`. These options would help when there are multiple dbt installations or when dbt is not on the user's path. Knowing the specific location may be more reliable than `shutil.which`, as Benoit suggests in his comment.

src/dbt_mcp/config/settings.py
DevonFulcher: Agreed. Can we check the exit code instead?

src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+    Args:
+        dbt_path: Path to the dbt executable
+        project_dir: Directory to validate as a dbt project
+        timeout: Maximum seconds to wait for validation (default 30)
+
+    Returns:
+        True if the directory is (or contains) a valid dbt project, False otherwise
+    """
+    try:
+        process = subprocess.Popen(
```
DevonFulcher, line 608: I think using `subprocess.run` instead would be a simpler approach, as it would allow you to remove much of the code related to timeouts and process termination.

src/dbt_mcp/config/settings.py
joellabes: I was not pleased with the way I had to detect a project either! 

>Can we check the exit code instead?
 
 I don't wait for the exit code - parsing can take a long time so as soon as there's some signs of life I kill it. 

>dbt debug but with changes to debug in Core/Fusion/Cloud CLI so that we don't check the warehouse connection. I think that today there is no way to deactivate it

This would be my preference (I coulda sworn there was a `no-debug-connection` flag or something, but it turns out i was mistaking it for `--skip-profile-setup`) but:
- I didn't want to chase it down across three repos
- It would also only work for updated dbt versions, whereas the MCP works with any version of dbt, right?
- I considered it unacceptable to accidentally open a Snowflake OAuth page as a result of initing the MCP server - it even annoys me when the VSCE does it, and I know what's happening there. 

>more simple heuristics: if it has a dbt_project.yml and it has any of the model/data/macros/tests/snapshots paths (we need to handle the case where those are modified in the dbt_project.yml ; then it is a dbt project -- This would be much faster than triggering any dbt command

I chose to not do this because dbt will also search in parent directories, so if your cwd is `/models` and you run `dbt run`, it'll work. As fragile as reading logs is, replicating dbt's "am I in a project" logic felt more fragile. 

Maybe the move is to optimistically check for `dbt_project.yml`, and only go to "invoke dbt and check the logs" as a fallback?

b-per: > I chose to not do this because dbt will also search in parent directories, so if your cwd is /models and you run dbt run, it'll work. As fragile as reading logs is, replicating dbt's "am I in a project" logic felt more fragile.

This is not what we do today with the current implementation.

The code calls
```
[dbt_path, "parse", "--project-dir", project_dir]
```

with `project_dir` being `cwd`. In that case, dbt doesn't check if we are in a subfolder of a project, it checks if we are at the root.

In a dbt project:
- `dbt parse` works in the `models` folder
- `dbt parse --project-dir .` doesn't work in the `models` folder


But t doesn't mean that it would stop us from looking at upstream folders (maybe up to 2 or 3 layers) to check if there is a `dbt_project.yml` in there and something that looks like a dbt project.

--- 

> Maybe the move is to optimistically check for dbt_project.yml, and only go to "invoke dbt and check the logs" as a fallback?

That could also work as well as a backup solution. Let's give a great experience to folks following the most common use case and a still-good experience to the ones who do weird stuff

src/dbt_mcp/config/settings.py
b-per: We could maybe look at `~/.local/bin/dbt` but at the last option. Otherwise with Claude Code for example, if folks have activated their venv with dbt Core we might find the Fusion binary at `~/.local/bin/dbt` when what the user wanted was dbt Core from the activated venv.

src/dbt_mcp/config/settings.py
b-per: I don't think this will work as the current approach uses `Popen` to get more control on when to stop etc... Like Joel mentioned, parse can be slow to run but by reading the first few lines of output (even if the command is not finished yet), we can already know if we are in a dbt project or not.

DevonFulcher: True, I missed that part.

src/dbt_mcp/config/settings.py
```diff
@@ -549,6 +583,180 @@ def validate_dbt_cli_settings(settings: DbtMcpSettings) -> list[str]:
...
+        dbt_path: Path to the dbt executable (required for validation)
+
+    Returns:
+        The current working directory if it's a valid dbt project, None otherwise
+    """
+    if not dbt_path:
+        # Can't validate without dbt
+        return None
+
+    cwd = os.getcwd()
```
VDFaller, line 749: I didn't do this originally, because this actually resolves to the mcp server folder in VSCode agent mode.  This is most likely due to the [recommended setup for manual testing](https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#manual-testing). Not sure if that matters.  

```
2026-02-04 22:27:31.195 [warning] [server stderr]   Value error, CWD is /home/me/repos/dbt-mcp [type=value_error, input_value={}, input_type=dict]
```

Issue #569, opened 2026-01-29
[Feature]: Human tool descriptions for improved README documentation


DevonFulcher:
### Is your feature request related to a problem?

Currently, the README contains all tool names, but there is no description of what the tools are. This is because that section of the README is autogenerated. Also, it is a pain to keep our doc site up-to-date with the current set of tool names.

### Describe the solution you'd like

Add a `human_description` field to the `dbt_mcp_tool` decorator that is used to populate the README. For example:
```
@dbt_mcp_tool(
    human_description="List all metrics defined in the project's semantic layer." # for humans
    description=get_prompt("semantic_layer/list_metrics"), # for AI
    title="List Metrics",
    read_only_hint=True,
    destructive_hint=False,
    idempotent_hint=True,
)
```
This field should be a short sentence rather than a long description for AI.

We should also remove this duplicated list from our docs site. We can link to the README on our docs site.

jairus-m:
I can continue the automated doc generation work ( #568 , #576) and pick this up since this is within that same realm =D

jairus-m:
@DevonFulcher if we use `human_description` from the `dbt_mcp_decorator` as the source of truth for the auto-population of the README , this would not work for the proxied tools right? As they don't have implemented tool defs in this repo?

What are your thoughts on creating some time of mapping instead that would be the basis for maintaining the descriptions? This instead would be directly used within the doc autogen script. Further, if `human_description` is still useful metadata, we can still have it but populated by the mapping? If not, can forgo the added field.

DevonFulcher:
@jairus-m ya, I think a mapping sounds like a good idea, similar to how we group toolsets!

PR #579, opened 2026-02-03
Add get_all_macros discovery tool

### LINKED ISSUE #535: [Feature]: Add "get_all_macros" tool

There is `get_all_models` and `get_all_sources`, and also a `get_macro_details`, but there is no `get_all_macros`. I am working a team who would like to review the existing macros and all types of templates like custom materialization via MCP.

`get_all_macros` returns an overview of all the macros

_No response_

### Additional context

_No response_

jairus-m:
## Summary
Add `get_all_macros` tool to Discovery tool set
- Filters out default dbt packages (refer to `DBT_BUILTIN_PACKAGES` in `src/dbt_mcp/discovery/client.py`
- Option to filter macros by package name
- Option to return package names only 
- Option to include or exclude default dbt package macros

## Why

Refer to issue.

## Related Issues

Closes #535

## Checklist

## Additional Notes

Validation in the comments below

### DIFF
--- a/src/dbt_mcp/discovery/client.py
+++ b/src/dbt_mcp/discovery/client.py
@@ -618,6 +618,7 @@ class MacrosFetcher:
         self,
         package_names: list[str] | None = None,
         return_package_names_only: bool = False,
+        include_default_dbt_packages: bool = False,
     ) -> list[dict] | list[str]:
         """Fetch all macros with optional filtering.
 
@@ -626,6 +627,8 @@ class MacrosFetcher:
             return_package_names_only: If True, returns only unique package names
                 instead of full macro details. Useful for discovering available
                 packages before drilling down.
+            include_default_dbt_packages: If True, includes the default dbt macros that
+                are maintained by dbt Labs.
 
         Returns:
             List of macros with name, uniqueId, description, and packageName,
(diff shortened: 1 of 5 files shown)


jairus-m:
## Validation

1. Filtering for just `amazon_ads` package:
Screenshot 2026-01-30 at 4 19 24 PM

2. Looking for built-in `dbt` package:
Screenshot 2026-01-30 at 4 20 59 PM

3. Filter for `return_package_names_only=true` (using the Codex App):
Screenshot 2026-02-03 at 8 20 51 AM

4. Use MCP Inspector to run `get_all_macros(packages=["dbt", "dbt_snowflake"], return_package_names_only=True, include_default_dbt_packages=True)`:
Screenshot 2026-02-04 at 2 52 23 PM

src/dbt_mcp/discovery/client.py
```diff
@@ -555,6 +605,71 @@ async def fetch_sources(
...
+        macros = await self._paginator.fetch_paginated(
+            GraphQLQueries.GET_MACROS,
+            variables={"filter": macro_filter},
+        )
+
+        # Filter out dbt-labs first-party macros
+        macros = [
+            m
+            for m in macros
+            if not self._is_dbt_builtin_package(m.get("packageName", ""))
```
DevonFulcher: Are we filtering these out because the agent may assume they already exist without a tool call?

jairus-m: It stems from the original realization that get_all_macros for many projects returns a ton of macros and that there can be a lot of noise. In addition, a POV was also shared in the GH issue that this tool should not return the out-of-the-box dbt macros.

In any case, i do think the future implementation of search would make this less of an issue!

Lastly, I'm re-reading the issue and totally missed a user wanting to inspect the default macros and then Beonoit's comment on having a param to turn that on or off. I'll have to make that update.

jairus-m:
Just documenting here that I will be forgoing the second requirement of adding required changes to dbt docs. Near future work (for #569) will eliminate the need to manually maintain the docs site in tandem with the tools introduced here.

PR #586, opened 2026-02-05
Logging Settings

VDFaller:
## Summary
I noticed that logging gets setup after settings so they aren't captured.  But #575 made me realize that more and more is getting done in settings so it should probably get logged. 

## What Changed
I moved Log settings to its own `DbtMcpLogSettings` model so that we can setup logging BEFORE trying to do the pydantic settings. 

## Why
So it listens to our dev calls when doing `DbtMcpSettings` warning and above in settings

## Checklist


DevonFulcher (approved):
Very nice, thank you!

PR #593, opened 2026-02-09
Fix cross-model metric queries failing with false validation error

DevonFulcher:
## Summary
- Removes the client-side `validate_query_metrics_params` preflight validation that incorrectly rejects valid cross-model metric queries
- The validation fetched available dimensions for all metrics at once, but the backend's `dimensionsPaginated` API returns only dimensions common to **all** metrics (intersection). When metrics span different semantic models, `metric_time` drops out of that intersection, causing false `"Group by metric_time not found"` errors
- The backend already validates parameters and correctly handles `metric_time` across semantic model boundaries (matching CLI behavior)
- This helps to address [this](https://getdbt.slack.com/archives/C046L0VTVR6/p1770656430655169) customer issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/langgraph_agent/uv.lock
+++ b/examples/langgraph_agent/uv.lock
@@ -594,7 +594,7 @@ wheels = [
 
 [[package]]
 name = "langsmith"
-version = "0.4.13"
+version = "0.6.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "httpx" },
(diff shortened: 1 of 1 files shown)


DevonFulcher:
> Gave the community Slack issue/thread a look!
> 
> For my notes: If I understand correctly, this removes all client-side validation for query metric params and just lets the existing backend checks handle that - esp as this client-side validation was leading to false negatives

Yup, that is right! I think that agents are more powerful now than they were when this was introduced, so it should no longer be necessary. Regardless, if we want better error messages we should add it to the API for all usages, not just MCP.

PR #606, opened 2026-02-12
Recategorize fetch_performance errors

DevonFulcher:
## Summary

These errors are miscategorized and causing some alerts in our backend systems.

## Checklist


src/dbt_mcp/discovery/client.py
```diff
@@ -947,14 +947,14 @@ async def fetch_performance(
                 name=name,
             )
             if not details:
-                raise ToolCallError(f"Model not found: {name}")
+                raise InvalidParameterError(f"Model not found: {name}")
```
wiggzz: Is this really an InvalidParameterError? I think this would be more like a NotFoundError or something?

DevonFulcher: Fixed

wiggzz (approved):

PR #590, opened 2026-02-07
[Bugfix]: Update args passed to dbt.listNodes

jairus-m:
## Summary

Fix `get_column_lineage` to pass model and column selectors as separate arguments to the dbt LSP.  

## What Changed

Updated `get_column_lineage` in `src/dbt_mcp/lsp/lsp_client.py` to split the single selector into two arguments for `dbt.listNodes`:                              
  - `model_selector: @{model_id}`                                                                                                                           
  - `column_selector: +column:{model_id}.{column_name}+`

This change is based of the working VS Code extension column lineage's dbt LSP trace output (@kbrock91 very helpfully called this out in the issue):

Screenshot 2026-02-07 at 9 47 17 AM

 
## Why

The `get_column_lineage` tool returned no results due to incorrect passing of args. With this update, lineage now works! Refer to the linked issue for a screenshot of my validation of the fix.

## Related Issues

Closes #582 

## Checklist
- [x] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Additional Notes

I'm trying to understand the why/context for this: did something change upstream in the LSP implementation where the previous single model+col selector arg passed to `dbt.listNodes` was working but now requires two separate model _and_ column selectors?

Did `get_column_lineage` work prior? If so, this may need more investigation before merging to better understand the before and after state and if these changes will affect users that are not currently having issues with the original implementation.

PR #607, opened 2026-02-12
Add validation for dbt LSP args in LSPClient

jairus-m:
## Summary

In response to #590 and in talks with @kbrock91 and @DevonFulcher, adds validation to make sure `dbt.listNodes` receives the correct args (and adds tests).

## Checklist

### DIFF
--- a/src/dbt_mcp/lsp/lsp_client.py
+++ b/src/dbt_mcp/lsp/lsp_client.py
@@ -9,6 +9,7 @@ import asyncio
 import logging
 from typing import Any
 
+from dbt_mcp.errors import InvalidParameterError
 from dbt_mcp.lsp.lsp_connection import LspEventName
 from dbt_mcp.lsp.providers.lsp_connection_provider import (
     LSPConnectionProviderProtocol,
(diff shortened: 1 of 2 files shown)


src/dbt_mcp/lsp/lsp_client.py
```diff
@@ -85,7 +85,18 @@ async def get_column_lineage(
...
+
+        Raises:
+            ValueError: If model_id or column_name is empty or invalid
         """
+        # Validate inputs - dbt.listNodes requires both model_selector and column_selector
+        # as separate args (https://github.com/dbt-labs/dbt-mcp/pull/590)
+        if not isinstance(model_id, str) or not model_id:
+            raise ValueError(f"model_id must be a non-empty string, got: {model_id!r}")
+        if not isinstance(column_name, str) or not column_name:
+            raise ValueError(
```
DevonFulcher: I think these need to be InvalidParameterErrors. If they remain a `ValueError`, then we will get alerted when this happens. This is a user-related error, not a server-side error, so we shouldn't get alerted.

jairus-m: thank you for explaining/teaching - will do! in the case of alerts, how do you mean? like in datadog? 

i do think i understand: so user errors are just logged but true serve-side errors are both logged and alerted in the monitoring system (?)

DevonFulcher: > how do you mean? like in datadog?

Yes, in Datadog, which can then trigger Slack messages, pages, and other incident-related automations.

> so user errors are just logged but true serve-side errors are both logged and alerted in the monitoring system

That is correct.

I plan to make this distinction clearer in this repo, since it is pretty opaque to open-source contributors.

PR #610, opened 2026-02-19
fix: show user-friendly error on OAuth setup failure

b-per:
## Summary
- When the local OAuth server becomes unreachable during setup (e.g., process crashed or timed out), the UI previously showed a raw `TypeError: Failed to fetch` in the Response block
- Now shows: *"Something went wrong when setting up the authentication. Please close this window and try again."*
- Non-network errors still display their original message


DevonFulcher (approved):
Nice!

PR #611, opened 2026-02-25
Upgrade MCP SDK version pin from ==1.23.1 to ==1.26.0

wiggzz:
## Why

The exact pin (`mcp[cli]==1.23.1`) prevents downstream consumers from pulling in newer versions of the MCP SDK. We are seeing a memory leak in production related to stateless HTTP session task accumulation in the MCP SDK. Upgrading to 1.26.x is an optimistic update to see if newer SDK versions mitigate the issue.

## What

- Changed the `mcp[cli]` dependency in `pyproject.toml` from `==1.23.1` to `~=1.26.0` (allows 1.26.x patches)
- Updated `uv.lock` to reflect the new constraint

No code changes -- just the dependency pin.

---

Drafted by Claude Opus 4.6 under the direction of @wiggzz


b-per:
> This is pinned because we rely on private variables here: [`main`/src/dbt_mcp/proxy/tools.py](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/proxy/tools.py).

Shouldn't we refactor our code not to depend on private mcp-sdk functions? This might become a problem in the future, no?

DevonFulcher:
> > This is pinned because we rely on private variables here: [`main`/src/dbt_mcp/proxy/tools.py](https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/proxy/tools.py).
> 
> Shouldn't we refactor our code not to depend on private mcp-sdk functions? This might become a problem in the future, no?

@b-per 

Yes, it is a bit of a risk. Pinning this dependency should ensure the SDK maintainers don't push changes that would break us, but pinning does introduce some risks if security patches are missed. A long-term fix might be:
- Submit changes to the Python SDK
- Migrate to Prefect's FastMCP library, but I'm unsure if they have proxy functionality that we can use to replace our usage.

PR #616, opened 2026-02-26
Metadata tool for current branch

DevonFulcher:
## Summary

This metadata tool can ensure we are on the right branch when testing local tools. I tested this with `task client`.

## Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20260226-164726.yaml
+++ b/.changes/unreleased/Under the Hood-20260226-164726.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Metadata tool for current branch
+time: 2026-02-26T16:47:26.161579-06:00


README.md
```diff
@@ -92,6 +92,7 @@ The dbt MCP server architecture allows for your agent to connect to a variety of
 - `get_column_lineage`: Traces column-level lineage locally (requires dbt-lsp via dbt Labs VSCE).
 
 ### MCP Server Metadata
+- `get_mcp_server_branch`: Returns the current git branch of the running dbt MCP server.
```
wiggzz, line 95: **questions** This is only relevant for running in development right? I don't think .git metadata will be present in the pypi package. What kind of situation did you need this for?

DevonFulcher: Yes, getting a git branch is only relevant for development. This toolset is off by default and is mainly intended for debugging. While we didn't actually use this tool during our testing, I planned to use it for testing https://github.com/dbt-labs/dbt-mcp/pull/617. The use case is just to ensure that the MCP server is configured correctly while testing a branch.

PR #617, opened 2026-02-26
Address SL expired oauth token

DevonFulcher:
## Summary

The `token` for `SemanticLayerConfig` was erroneously using the static `settings.dbt_token` rather than the refreshed value `token_provider.get_token()`. This resulted in expired token usage for long-running dbt MCP sessions. I'm trying to get this fix out asap and will follow up with a deeper analysis and testing.

## Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works (I will follow up with this)

### DIFF
--- a/src/dbt_mcp/config/config_providers.py
+++ b/src/dbt_mcp/config/config_providers.py
@@ -75,7 +75,7 @@ class DefaultSemanticLayerConfigProvider(ConfigProvider[SemanticLayerConfig]):
             url=f"http://{host}" if is_local else f"https://{host}" + "/api/graphql",
             host=host,
             prod_environment_id=settings.actual_prod_environment_id,
-            token=settings.dbt_token,
+            token=token_provider.get_token(),
             headers_provider=SemanticLayerHeadersProvider(
                 token_provider=token_provider
             ),
--- a/src/dbt_mcp/config/settings.py
+++ b/src/dbt_mcp/config/settings.py
@@ -608,6 +608,7 @@ class CredentialsProvider:
             self.settings.dbt_host = get_dbt_host(self.settings, dbt_platform_context)
             if not dbt_platform_context.decoded_access_token:
                 raise ValueError("No decoded access token found in OAuth context")
+            # TODO: This is unreliable. We shouldn't set token here because it is static and not refreshed.
             self.settings.dbt_token = dbt_platform_context.decoded_access_token.access_token_response.access_token
 
             self.token_provider = OAuthTokenProvider(


src/dbt_mcp/config/config_providers.py
```diff
@@ -75,7 +75,7 @@ async def get_config(self) -> SemanticLayerConfig:
             url=f"http://{host}" if is_local else f"https://{host}" + "/api/graphql",
             host=host,
             prod_environment_id=settings.actual_prod_environment_id,
-            token=settings.dbt_token,
+            token=token_provider.get_token(),
```
wiggzz, line 78: Ok yea this explains why the second attempt also didn't work. Because after the expiration, the token will never work.

I think the other issue we saw is also relevant.

DevonFulcher: > I think the other issue we saw is also relevant.

Can you elaborate? This seems to fix the issue with: `AuthError(["UNAUTHENTICATED: [FlightSQL] Invalid credentials (Unauthenticated; Prepare)"])`

wiggzz: Yes, this fixes that issue. But I think that it's also still possible for a person to load Claude, then wait a long time, and then Claude would try to run a query, and it would fail because the token was expired. This would be in the `list_metrics`, `list_saved_queries`, `get_dimensions`, `get_entities` calls which use graph ql and the headers provider, _not_ the token.

DevonFulcher: I'm unsure if there is an issue with those tools. As long as we are using `token_provider.get_token()` rather than `settings.dbt_token`, we are using the access token that is either:
- Read and expiry checked on start up from `~/.dbt/mcp.yml` [here](https://github.com/dbt-labs/dbt-mcp/blob/4f54a2ef57bb6de255bb327356f3b0f3b598303e/src/dbt_mcp/config/settings.py#L450).
- Retrieved from the oauth flow on start up if `~/.dbt/mcp.yml` is expired [here](https://github.com/dbt-labs/dbt-mcp/blob/4f54a2ef57bb6de255bb327356f3b0f3b598303e/src/dbt_mcp/config/settings.py#L461).
- Refreshed automatically in the background with a 5 minute buffer [here](https://github.com/dbt-labs/dbt-mcp/blob/4f54a2ef57bb6de255bb327356f3b0f3b598303e/src/dbt_mcp/oauth/token_provider.py#L86).

This is the only place we are reading from `settings.dbt_token` rather than `token_provider.get_token()` which is used for `query_metrics`, not those other tools. For those other tools, we are calling `token_provider.get_token()` [here](https://github.com/dbt-labs/dbt-mcp/blob/4f54a2ef57bb6de255bb327356f3b0f3b598303e/src/dbt_mcp/config/headers.py#L19).

With that said, `token_provider.get_token()` perhaps shouldn't rely on externalized logic to ensure the token expiry valid. It should likely check the expiration on its own and refresh it if necessary to be 100% sure it is valid before returning.

DevonFulcher: I will follow up with that change. I want to keep this PR as minimal as possible as a minimal-viable-fix. I want to fix the structural issues that caused this to happen and difficult to debug in follow up PRs.

wiggzz: Yea, that's fine. Yea, my point was that even though we're using token_provider.get_token() - this relies on the token being not expired. On the _first_ invocation of get_token, we'll use the token that was put into the token provider at start up - which will mean that it could be expired on the first call (depending on whether get_token() had previously been called - it may be called during the set up in some situations it looks like but I don't think we want to rely on this). This risk is mitigated in that that call will also trigger a background refresh, so subsequent calls _should_ succeed. We can address in a separate PR but I think we need to initiate the background refresh process eagerly when we initialize the token provider rather than only the first time calling get_token())

DevonFulcher: Ah, yes. I will add that to the RCA.

DevonFulcher:
Note: This PR was stacked on top of #616, so the Files Changed tab includes changes from both PRs. For the diff specific to this PR, see the merge commit: https://github.com/dbt-labs/dbt-mcp/commit/fe9d442

PR #620, opened 2026-03-02
Adds a product docs toolset to the dbt MCP

mirnawong1:
## Summary

This PR adds a `product_docs` toolset to the dbt MCP server, giving AI agents real-time access to the public dbt documentation at docs.getdbt.com. The toolset exposes **two** MCP tools with a clean separation of concerns: search for pages, then fetch their content.

Product docs tools are included by default when you run dbt-mcp; use `DISABLE_PRODUCT_DOCS=true` to turn them off, or `DBT_MCP_ENABLE_PRODUCT_DOCS=true` when using an allowlist (other `DBT_MCP_ENABLE_*` vars).

#### Flow
```
User asks a question about dbt
│
▼
┌─────────────────────────┐
│ 1. search_product_docs │ Agent sends keywords
│ │ Tool searches llms.txt index (title/description matching)
│ │ If fewer than 3 results, runs `expand_keywords(query)` to get synonyms/abbreviations (e.g. "udf" → "user-defined function"), then does a second search_index call with the expanded query.
Merges and deduplicates results by URL.
Never touches llms-full.txt.
│ │ Returns: titles, URLs, descriptions (NO content)
└───────────┬─────────────┘
│
│ Agent now has URLs for relevant pages
│
▼
┌─────────────────────────────────────────────────┐
│ 2. get_product_doc_pages([paths]) │
│ Fetches 1–10 pages in parallel as Markdown │
└─────────────────────────────────────────────────┘
│
▼
 Agent synthesizes answer from full page content
```

### Two-tool surface (`src/dbt_mcp/product_docs/tools.py`)

Designed with minimal tool count to reduce LLM confusion and MCP context overhead:

- **`search_product_docs`** — keyword search over the llms.txt index (returns metadata: titles, URLs, descriptions). Includes automatic full-text fallback via `llms-full.txt` when keyword search finds fewer than 3 results.
- **`get_product_doc_pages`** — fetch one or more pages by path or URL as Markdown (up to 10 in parallel). A single tool handles both single-page and multi-page fetching — pass a list of one path or many.

### Typed responses (`src/dbt_mcp/product_docs/types.py`)

All tools return typed `@dataclass` responses (following the existing `semantic_layer/types.py` pattern) instead of raw JSON strings:

- `DocSearchResult`, `SearchProductDocsResponse`, `ProductDocPageResponse`, `GetProductDocPagesResponse`

FastMCP natively serializes dataclasses, so no manual `json.dumps()` is needed.

### Client with caching and search ranking (`src/dbt_mcp/product_docs/client.py`)

- `ProductDocsClient` handles HTTP fetching with TTL-based in-memory caching (1h index, 24h full-text, 30m pages).
- Relevance-ranked search with documented scoring weights (extracted into `score_index_entry()` with named constants like `SCORE_KEYWORD_IN_TITLE`, `SCORE_EXACT_TITLE_MATCH`, etc.).
- Abbreviation expansion (e.g. "CI" → "continuous integration", "SL" → "semantic layer").
- Content truncation (28k chars) to keep responses manageable for LLM context windows.
- `client.get_page()` raises `httpx.HTTPStatusError` / `httpx.RequestError` — tool layer catches exceptions and returns typed error responses.

### Supporting changes

- `tool_names.py` / `toolsets.py` — `SEARCH_PRODUCT_DOCS` and `GET_PRODUCT_DOC_PAGES` enum members added.
- `human_descriptions.py` — descriptions for both tools.
- Prompt `.md` files — `search_product_docs.md` and `get_product_doc_pages.md` with guidance on how to present docs content to users.
- `config.py` / `settings.py` / `server.py` — product docs toolset registration with `DISABLE_PRODUCT_DOCS` support.
- `README.md` / `diagram.d2` — auto-updated by pre-commit hook.
- `.changes/unreleased/` — changie entry added.

### Tests

- **Unit tests** (`tests/unit/tools/test_product_docs.py`) — 42 tests covering parsing, URL normalization, search ranking, full-text fallback, page fetching (success, 404, network error, partial failures, 10-page cap), and toolset registration.
- **Integration tests** (`tests/integration/product_docs/test_product_docs.py`) — live tests against docs.getdbt.com for client methods and both MCP tools.

## Why

1. **Minimal tool surface** — Two tools instead of three. A single `get_product_doc_pages` tool accepts a list of paths (1 to 10), eliminating the need for separate single/batch tools. Fewer tools = less LLM confusion about which to pick, less MCP context overhead.
2. **Typed returns** — Returning dataclasses instead of dicts/JSON strings aligns with the rest of the codebase (e.g. semantic layer tools) and gives downstream consumers type safety.
3. **Documented scoring** — Search relevance weights are named constants with a docstring explaining the heuristic, not magic numbers.
4. **Robust error handling** — Raising exceptions from the HTTP layer instead of returning error strings, with typed error responses at the tool level.

## Checklist

- [ ] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required — WILL ADD THIS WHEN PR MERGED

### Additional notes

Product docs tools are on by default. Set `DISABLE_PRODUCT_DOCS=true` to disable them. If you use an allowlist (e.g. `DBT_MCP_ENABLE_SEMANTIC_LAYER=true`), add `DBT_MCP_ENABLE_PRODUCT_DOCS=true` to include product docs. The `read_only_hint=True` annotation is correct per the MCP spec — the in-memory cache is internal process state, not an externally observable side effect.

### DIFF
--- a/src/dbt_mcp/product_docs/client.py
+++ b/src/dbt_mcp/product_docs/client.py
@@ -1,14 +1,16 @@
 """Client for fetching, caching, and searching docs.getdbt.com content.
 
-Provides ``ProductDocsClient``, which handles HTTP requests, TTL-based
+Provides ``ProductDocsClient``, which handles HTTP requests, in-memory
 caching, and keyword search/ranking over the llms.txt index and
 llms-full.txt full-text corpus.
+
+Caches live for the lifetime of the MCP server process; restart the
+server to refresh.
 """
 
-import asyncio
 import logging
 import re
-import time
+from typing import Any
 
 import httpx
 
@@ -17,9 +19,6 @@ logger = logging.getLogger(__name__)
 LLMS_TXT_URL = "https://docs.getdbt.com/llms.txt"
 LLMS_FULL_TXT_URL = "https://docs.getdbt.com/llms-full.txt"
 DOCS_BASE_URL = "https://docs.getdbt.com"
-INDEX_CACHE_TTL_SECONDS = 3600  # 1 hour
-PAGE_CACHE_TTL_SECONDS = 1800  # 30 minutes
-FULL_TEXT_CACHE_TTL_SECONDS = 86400  # 24 hours (large file, changes rarely)
 
 # Cap page content length in tool responses to avoid Cursor IDE freezing.
 MAX_CONTENT_CHARS_PER_PAGE = 28_000
@@ -303,75 +302,38 @@ def expand_keywords(query: str) -> list[str]:
 class ProductDocsClient:
     """Async client for fetching and searching docs.getdbt.com content.
 
-    All caching state lives on the instance, not at module level.
+    Caches are simple dicts that live for the lifetime of the instance
+    (and thus the MCP server process).  Restart the server to refresh.
     """
 
     def __init__(self) -> None:
-        self._index_cache: list[dict[str, str]] | None = None
-        self._index_cache_time: float = 0.0
-
-        self._page_cache: dict[str, tuple[float, str]] = {}
-
-        self._full_text_cache: list[dict[str, str]] | None = None
-        self._full_text_cache_time: float = 0.0
-
-        self._index_lock = asyncio.Lock()
-        self._full_text_lock = asyncio.Lock()
+        self._cache: dict[str, Any] = {}
 
     # -- fetchers ------------------------------------------------------------
 
     async def get_index(self) -> list[dict[str, str]]:
-        """Return the cached llms.txt index, refreshing if stale."""
-        if (
-            self._index_cache is not None
-            and (time.time() - self._index_cache_time) < INDEX_CACHE_TTL_SECONDS
-        ):
-            return self._index_cache
-
-        async with self._index_lock:
-            if (
-                self._index_cache is not None
-                and (time.time() - self._index_cache_time) < INDEX_CACHE_TTL_SECONDS
-            ):
-                return self._index_cache
-
+        """Return the cached llms.txt index, fetching on first call."""
+        if "index" not in self._cache:
             logger.info("Fetching llms.txt index from %s", LLMS_TXT_URL)
             async with httpx.AsyncClient(timeout=30.0, follow_redirects=True) as client:
                 response = await client.get(LLMS_TXT_URL)
                 response.raise_for_status()
-
-            self._index_cache = parse_llms_txt(response.text)
-            self._index_cache_time = time.time()
-            logger.info("Cached llms.txt index: %d pages", len(self._index_cache))
-            return self._index_cache
+            self._cache["index"] = parse_llms_txt(response.text)
+            logger.info("Cached llms.txt index: %d pages", len(self._cache["index"]))
+        return self._cache["index"]
 
     async def get_full_text_index(self) -> list[dict[str, str]]:
-        """Return the cached llms-full.txt page index, refreshing if stale."""
-        if (
-            self._full_text_cache is not None
-            and (time.time() - self._full_text_cache_time) < FULL_TEXT_CACHE_TTL_SECONDS
-        ):
-            return self._full_text_cache
-
-        async with self._full_text_lock:
-            if (
-                self._full_text_cache is not None
-                and (time.time() - self._full_text_cache_time)
-                < FULL_TEXT_CACHE_TTL_SECONDS
-            ):
-                return self._full_text_cache
-
+        """Return the cached llms-full.txt page index, fetching on first call."""
+        if "full_text" not in self._cache:
             logger.info("Fetching llms-full.txt from %s", LLMS_FULL_TXT_URL)
             async with httpx.AsyncClient(
                 timeout=120.0, follow_redirects=True
             ) as client:
                 response = await client.get(LLMS_FULL_TXT_URL)
                 response.raise_for_status()
-
-            self._full_text_cache = parse_llms_full_txt(response.text)
-            self._full_text_cache_time = time.time()
-            logger.info("Cached llms-full.txt: %d pages", len(self._full_text_cache))
-            return self._full_text_cache
+            self._cache["full_text"] = parse_llms_full_txt(response.text)
+            logger.info("Cached llms-full.txt: %d pages", len(self._cache["full_text"]))
+        return self._cache["full_text"]
 
     async def get_page(self, url: str) -> str:
         """Fetch a page with caching.
@@ -380,23 +342,13 @@ class ProductDocsClient:
         Raises httpx.HTTPStatusError on 4xx/5xx responses.
         Raises httpx.RequestError on network/connection failures.
         """
-        cached = self._page_cache.get(url)
-        if cached is not None and (time.time() - cached[0]) < PAGE_CACHE_TTL_SECONDS:
-            return cached[1]
-
-        logger.info("Fetching product doc page: %s", url)
-        async with httpx.AsyncClient(timeout=30.0, follow_redirects=True) as client:
-            response = await client.get(url)
-            response.raise_for_status()
-            content = response.text
-
-        self._page_cache[url] = (time.time(), content)
-
-        if len(self._page_cache) > 50:
-            oldest_key = min(self._page_cache, key=lambda k: self._page_cache[k][0])
-            self._page_cache.pop(oldest_key, None)
-
-        return content
+        if url not in self._cache:
+            logger.info("Fetching product doc page: %s", url)
+            async with httpx.AsyncClient(timeout=30.0, follow_redirects=True) as client:
+                response = await client.get(url)
+                response.raise_for_status()
+            self._cache[url] = response.text
+        return self._cache[url]
 
     # -- search --------------------------------------------------------------
 
--- a/tests/unit/tools/test_product_docs.py
+++ b/tests/unit/tools/test_product_docs.py
@@ -465,8 +465,7 @@ class TestSearchFullText:
         from dbt_mcp.product_docs.client import ProductDocsClient
 
         client = ProductDocsClient()
-        client._full_text_cache = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
-        client._full_text_cache_time = 9999999999.0
+        client._cache["full_text"] = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
         results = await client.search_full_text(["rust"])
         urls = [r["url"] for r in results]
         assert any("fusion" in u for u in urls)
@@ -476,8 +475,7 @@ class TestSearchFullText:
         from dbt_mcp.product_docs.client import ProductDocsClient
 
         client = ProductDocsClient()
-        client._full_text_cache = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
-        client._full_text_cache_time = 9999999999.0
+        client._cache["full_text"] = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
         results = await client.search_full_text(["zzzznonexistent"])
         assert results == []
 
@@ -486,8 +484,7 @@ class TestSearchFullText:
         from dbt_mcp.product_docs.client import ProductDocsClient
 
         client = ProductDocsClient()
-        client._full_text_cache = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
-        client._full_text_cache_time = 9999999999.0
+        client._cache["full_text"] = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
         results = await client.search_full_text(["rust", "deploy"])
         urls = [r["url"] for r in results]
         assert any("fusion" in u for u in urls)
@@ -498,8 +495,7 @@ class TestSearchFullText:
         from dbt_mcp.product_docs.client import ProductDocsClient
 
         client = ProductDocsClient()
-        client._full_text_cache = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
-        client._full_text_cache_time = 9999999999.0
+        client._cache["full_text"] = parse_llms_full_txt(SAMPLE_LLMS_FULL_TXT)
         results = await client.search_full_text(["incremental"])
         assert results[0]["url"].endswith("incremental-models-overview")
 


mirnawong1:
i'm not sure where the failure is coming from :(

.cursor/mcp.json
b-per, line 1: We should not commit this, should we?

.cursor/mcp.json.bak
b-per, line 1: This should be removed

src/dbt_mcp/tools/human_descriptions.py
```diff
@@ -67,6 +67,10 @@
     ToolName.FUSION_GET_COLUMN_LINEAGE: "Traces column-level lineage via dbt Platform.",
     # MCP Server tools
     ToolName.GET_MCP_SERVER_VERSION: "Returns the current version of the dbt MCP server.",
+    # Product docs tools
+    ToolName.SEARCH_PRODUCT_DOCS: "Searches docs.getdbt.com for pages matching a query; returns metadata (titles, URLs, descriptions).",
+    ToolName.GET_PRODUCT_DOC_PAGE: "Fetches the full Markdown content of a single docs.getdbt.com page by path or URL.",
+    ToolName.SEARCH_AND_GET_PRODUCT_DOC: "Searches docs.getdbt.com and returns the full content of the best-matching page(s) in one call.",
```
b-per: Why do we need this last tool which combines both? Wouldn't it confuse the LLM when it will need to pick which tool to use?

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
...
+            "total_matches": len(results),
+            "showing": len(results),
+            "results": results,
+        },
+        indent=2,
+    )
+
+
+@dbt_mcp_tool(
+    description=(
```
b-per: Those descriptions are usually in their own markdown files (you check check how other tools do it)

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
...
+    if not fetched_pages:
+        return json.dumps({"query": query, "error": "Failed to fetch matching pages."})
+
+    remaining = [r for r in results[max_pages:] if r not in pages_to_fetch]
+
+    response: dict[str, object] = {
+        "query": query,
+        "pages": fetched_pages,
+        "other_results": remaining[:5],
+        "usage_guidance": (
```
b-per: Is this field part of the MCP spec? Or where does it come from?

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
+"""Product Docs tools for searching and retrieving docs.getdbt.com content.
+
+Provides tools that let AI agents query the public dbt product
+documentation in real time:
```
b-per: Is it required? What does "real time" mean in this context?

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
...
+                "fully meet the user's need. ALWAYS include the docs page URL "
+                "as a markdown hyperlink at the end of your response, "
+                "e.g. [Page title](url)."
+            ),
+        },
+        indent=2,
+    )
+
+
+def _search_index(query: str) -> list[dict[str, str]]:
```
b-per: Where did this logic come from? Is it taken from somewhere or did the LLM generate it on its own? We should describe what we are trying to do.

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
...
+_index_cache_time: float = 0.0
+
+_page_cache: dict[str, tuple[float, str]] = {}
+
+_full_text_lock = threading.Lock()
+_full_text_cache: list[dict[str, str]] | None = None
+_full_text_cache_time: float = 0.0
+
+_http_client: httpx.Client | None = None
+_http_client_lock = threading.Lock()
```
b-per: This feels very low level. What led to this implementation? If we want to keep it, can we separate it from the `tools.py` file?

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
...
+        return content
+    return (
+        content[:max_chars].rstrip()
+        + f"\n\n---\n*Content truncated for length. Full page: {url}*"
+    )
+
+
+def _get_http_client() -> httpx.Client:
+    """Return a shared HTTP client with connection pooling."""
+    global _http_client
```
b-per: I'm not sure that we can use `global` in the context of the remote MCP

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,748 @@
...
+        response = _get_http_client().get(LLMS_TXT_URL)
+        response.raise_for_status()
+
+        _index_cache = _parse_llms_txt(response.text)
+        _index_cache_time = time.time()
+        logger.info("Cached llms.txt index: %d pages", len(_index_cache))
+        return _index_cache
+
+
+def _get_page(url: str) -> str:
```
b-per: Can we check if we should use async for that?

src/dbt_mcp/product_docs/tools.py
DevonFulcher, line 1: In general, can the functions return dataclasses or Pydantic models rather than dicts?

src/dbt_mcp/tools/human_descriptions.py
mirnawong1: addressed

src/dbt_mcp/product_docs/tools.py
mirnawong1: addressed

src/dbt_mcp/product_docs/tools.py
mirnawong1: removed

src/dbt_mcp/product_docs/tools.py
mirnawong1: reworked flow - replaced search_and_get_product_doc with batch_get_product_doc_pages, fixed `client.get_page() ` to raise exceptions instead of returning error strings

src/dbt_mcp/product_docs/tools.py
mirnawong1: removed

src/dbt_mcp/product_docs/tools.py
mirnawong1: yes w should! removed!

src/dbt_mcp/product_docs/tools.py
mirnawong1: return typed dataclasses instead of JSON strings

scripts/test-docs-approaches.py
DevonFulcher, line 1: This may be better suited as an integration test.

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,230 @@
+"""Product Docs MCP tool definitions.
+
+Registers three tools that let AI agents query the public dbt product
+documentation at docs.getdbt.com in real time:
+
+- ``search_product_docs`` — keyword search against the llms.txt index
+  (with automatic full-text fallback via llms-full.txt)
+- ``get_product_doc_page`` — fetch a single page as clean Markdown
+- ``batch_get_product_doc_pages`` — fetch multiple pages in parallel
```
b-per: Why would we want 2 tools? Can't we use 1 tool where we can provide a list of just 1 page if required?

What we'd want to achieve is:
- less tools = less chances for the LLM to get confused about which one to pick
- less tools = less context that the MCP server takes

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,421 @@
...
+
+_ABBREVIATION_EXPANSIONS: dict[str, list[str]] = {
+    "udf": ["user-defined function", "user-defined functions"],
+    "udfs": ["user-defined function", "user-defined functions"],
+    "ci": ["continuous integration"],
+    "cd": ["continuous deployment"],
+    "ci/cd": ["continuous integration", "continuous deployment"],
+    "sl": ["semantic layer"],
+    "ide": ["studio ide", "cloud ide"],
+    "cli": ["command line", "cloud cli"],
```
b-per: Didn't we update the naming for the Cloud CLI as well (there were Slack conversations a few days ago)

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,421 @@
...
+    "vc": ["version control"],
+    "pr": ["pull request"],
```
b-per, line 74: If we add "pr" let's maybe also add the gitlab one, "mr" for "merge request". Same thing, just a different name

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,421 @@
...
+        "about",
+        "using",
+        "use",
+        "set",
+        "up",
+        "dbt",
+    }
+)
+
+_ABBREVIATION_EXPANSIONS: dict[str, list[str]] = {
```
b-per, line 58: we could add "repo" => "repository" also, no?

b-per: and `gh` : `github` ?

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,421 @@
...
+            return self._full_text_cache
+
+    async def get_page(self, url: str) -> str:
+        """Fetch a page with caching.
+
+        Returns the page content as a string on success.
+        Raises httpx.HTTPStatusError on 4xx/5xx responses.
+        Raises httpx.RequestError on network/connection failures.
+        """
+        cached = self._page_cache.get(url)
```
b-per: I might be missing something, but where do we define this function?

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,421 @@
...
+
+            if len(matching) == len(keywords):
+                score += 15
+
+            if query_lower == title_lower:
+                score += 50
+
+            if title_lower and query_lower in title_lower:
+                focus_ratio = len(query_lower) / len(title_lower)
+                score += focus_ratio * 15
```
b-per: Could we maybe extract this to its own function and explain what those numbers are and do? They feel a bit too magic at the moment.

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,230 @@
...
+        showing=len(doc_results),
+        results=doc_results,
+        search_method="full_text_fallback" if used_full_text else None,
+    )
+
+
+@dbt_mcp_tool(
+    description=get_prompt("product_docs/get_product_doc_page"),
+    title="Get Product Doc Page",
+    read_only_hint=True,
```
b-per, line 146: Is it truly read-only if we create a local cache (not a trick question, I genuinely don't know)

src/dbt_mcp/product_docs/client.py
mirnawong1: yes great catch!

src/dbt_mcp/product_docs/tools.py
DevonFulcher: I think read-only is appropriate. I think the hint pertains to the semantics of the tool rather than the implementation details of that nature.

src/dbt_mcp/product_docs/client.py
mirnawong1: i believed it's defined in line 219 [here](https://github.com/dbt-labs/dbt-mcp/pull/620/files#diff-3caa665068f737d08a9e42f1a17ec4492beac3387720f9ab27673b3f8a4519d7R219)

src/dbt_mcp/product_docs/tools.py
mirnawong1: yes, i believe it's more around 'does it change the environment', which it doesn't. also mentioned here: https://modelcontextprotocol.io/specification/2025-11-25/schema#toolannotations-readonlyhint

scripts/test-docs-approaches.py
mirnawong1: addressed here https://github.com/dbt-labs/dbt-mcp/commit/9e73215451291e39a71cc6d58f5f005a7bb69fbc#diff-4a42f06838422473febc82fb22627076db1aa04527d96bd5e3dc1aad82390d0dR1

src/dbt_mcp/product_docs/client.py
mirnawong1: addressed here https://github.com/dbt-labs/dbt-mcp/commit/9e73215451291e39a71cc6d58f5f005a7bb69fbc#diff-3caa665068f737d08a9e42f1a17ec4492beac3387720f9ab27673b3f8a4519d7R87

tests/integration/product_docs/test_product_docs.py
```diff
@@ -0,0 +1,111 @@
...
+        assert result.total_matches == 0
+
+    @pytest.mark.asyncio
+    async def test_get_product_doc_page(self, context: ProductDocsToolContext):
+        fn = getattr(get_product_doc_page, "fn", get_product_doc_page)
+        page = await fn(context, "/docs/build/incremental-models")
+
+        assert page.error is None
+        assert len(page.content) > 0
+        assert "docs.getdbt.com" in page.url
```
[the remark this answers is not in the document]
mirnawong1: addressed

src/dbt_mcp/product_docs/tools.py
mirnawong1: ah ok good flag and thank you for explaining. have combined both now!

mirnawong1:
great thanks so much @b-per and @DevonFulcher ! ive addressed all the comments and ready for your re-review whenever you have a second. 

one thing that i'm not sure about is the full-text fallback mechanism. I've noticed that when a `search_product_docs` query returns less than 3 results from the lightweight llms.txt index, it automatically falls back to fetching llms-full.txt — the entire dbt documentation corpus. And this this freezes my cursor ide. Not sure how often this would happen but wondering if there's a clever way to circumvent this?

DevonFulcher:
> falls back to fetching llms-full.txt

@mirnawong1 is it possible to remove this fallback? I don't think we want to feed our entire docs site into the LLM.

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+    pages: list[ProductDocPageResponse] = []
+    for i, result in enumerate(results):
+        if isinstance(result, BaseException):
+            logger.warning("Failed to fetch %s: %s", paths[i], result)
+            pages.append(
+                ProductDocPageResponse(
+                    url=paths[i],
+                    content="",
+                    error=f"Failed to fetch page: {paths[i]} ({result})",
```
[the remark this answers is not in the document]
mirnawong1: addressed https://github.com/dbt-labs/dbt-mcp/pull/620/commits/12884729812b2d2c2b987060ae3c1dcd88158170

src/dbt_mcp/product_docs/tools.py
```diff
@@ -0,0 +1,209 @@
...
+    """Register Product Docs tools."""
+
+    def bind_context() -> ProductDocsToolContext:
+        return ProductDocsToolContext()
+
+    register_tools(
+        dbt_mcp,
+        tool_definitions=[
+            tool.adapt_context(bind_context) for tool in PRODUCT_DOCS_TOOLS
+        ],
```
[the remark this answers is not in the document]
mirnawong1: addressed https://github.com/dbt-labs/dbt-mcp/pull/620/commits/12884729812b2d2c2b987060ae3c1dcd88158170

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,467 @@
...
+    if not url.startswith("http"):
+        url = url.lstrip("/")
+        url = f"{DOCS_BASE_URL}/{url}"
+
+    url = url.rstrip("/")
+
+    if not url.endswith(".md"):
+        url = f"{url}.md"
+
+    return url
```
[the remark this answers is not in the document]
mirnawong1: addressed https://github.com/dbt-labs/dbt-mcp/pull/620/commits/12884729812b2d2c2b987060ae3c1dcd88158170

tests/unit/tools/test_product_docs.py
```diff
@@ -0,0 +1,563 @@
...
+        assert not result.pages[0].url.endswith(".md")
+
+    @pytest.mark.asyncio
+    async def test_rejects_external_url(self, context, mock_client):
+        result = await get_product_doc_pages.fn(
+            context, ["https://evil.com/steal-data"]
+        )
+        page = result.pages[0]
+        assert page.error is not None
+        assert "docs.getdbt.com" in page.error
```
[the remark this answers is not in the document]
mirnawong1: addressed

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,473 @@
...
+    """Async client for fetching and searching docs.getdbt.com content.
+
+    All caching state lives on the instance, not at module level.
+    """
+
+    def __init__(self) -> None:
+        self._index_cache: list[dict[str, str]] | None = None
+        self._index_cache_time: float = 0.0
+
+        self._page_cache: dict[str, tuple[float, str]] = {}
```
b-per: Oh, OK, so the cache is just an in-memory dict and not using persistent files.

I think that in that case, we might not need any of the following vars
```
INDEX_CACHE_TTL_SECONDS = 3600  # 1 hour
PAGE_CACHE_TTL_SECONDS = 1800  # 30 minutes
FULL_TEXT_CACHE_TTL_SECONDS = 86400  # 24 hours (large file, changes rarely)
```

The cache will live as long as the MCP server is on, but i doubt that people will keep MCP servers on for more than a couple of hours.

If for testing purpose we want to refresh the cache we'd just restart the MCP server.

Can we just simplify the cache approach with a simple "non-expiring" cache? like

``` py 
self._cache: dict[str, Any] = {}

async def get_index(self):
    if "index" not in self._cache:
        self._cache["index"] = await self._fetch_index()
    return self._cache["index"]

async def get_full_text_index(self):
    if "full_text" not in self._cache:
        self._cache["full_text"] = await self._fetch_full_text()
    return self._cache["full_text"]

async def get_page(self, url: str):
    if url not in self._cache:
        self._cache[url] = await self._fetch_page(url)
    return self._cache[url]
```

src/dbt_mcp/product_docs/client.py
```diff
@@ -0,0 +1,473 @@
...
+
+        self._full_text_cache: list[dict[str, str]] | None = None
+        self._full_text_cache_time: float = 0.0
+
+        self._index_lock = asyncio.Lock()
```
b-per: Is this complexity really required? Similar to the comment above, can we simplify the overall caching approach?

src/dbt_mcp/product_docs/client.py
mirnawong1: thanks so much @b-per ! ok i've removed the vars and also replaced the TTL-based caching with a simple dict

src/dbt_mcp/product_docs/client.py
mirnawong1: addressed https://github.com/dbt-labs/dbt-mcp/pull/620/commits/dafca944743d21ed0ac11762c6854f42a4d91411

src/dbt_mcp/product_docs/client.py
mirnawong1: addressed https://github.com/dbt-labs/dbt-mcp/pull/620/commits/dafca944743d21ed0ac11762c6854f42a4d91411

mirnawong1:
thanks @b-per ! ready for re-review when you can

b-per:
Your commits are not signed though. Before any future contribution, could you check how to set git to sign your commits? https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md

PR #624, opened 2026-03-04
Add meta field passthrough for MCP Apps support

jgiuffrida:
## Summary
- Adds `meta: dict[str, Any] | None = None` to `GenericToolDefinition`, the `@dbt_mcp_tool` decorator, and `register_tools()` passthrough to `FastMCP.add_tool()`
- This is infrastructure-only — no new tools or features. All existing tools are unaffected (`meta` defaults to `None`)
- Enables future MCP App tools to declare associated UIs via `meta={"ui": {"resourceUri": "ui://..."}}`
- Adds `CLAUDE.md` with project architecture overview for AI-assisted development
- Adds "Adding Tools" and "Adding an MCP App tool" sections to `CONTRIBUTING.md`
- Adds 7 new unit tests for meta passthrough through the full pipeline
- Updates `MockFastMCP` in test conftest to capture `tool_kwargs` for future test assertions

## Test plan
- [x] All 386 unit tests pass (`uv run pytest tests/ --ignore=tests/integration -x -q`)
- [x] New tests cover: decorator, `adapt_context()`, `to_fastmcp_internal_tool()`, and `generic_register_tools()` for both `meta=value` and `meta=None`
- [ ] Verify no behavioral changes to existing tools in a client (e.g., Claude, Cursor)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changes/unreleased/Under the Hood-20260303-194726.yaml
+++ b/.changes/unreleased/Under the Hood-20260303-194726.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Loosen version pin to dbt-protos>=1.0.431
+time: 2026-03-03T19:47:26.544664-08:00
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 ]
 dependencies = [
   "authlib==1.6.7",
-  "dbt-protos==1.0.382",
+  "dbt-protos>=1.0.431",
   "dbt-sl-sdk[sync]>=0.13.2",
   "dbtlabs-vortex==0.2.0",
   "fastapi>=0.116.1",
--- a/uv.lock
+++ b/uv.lock
@@ -291,7 +291,7 @@ dev = [
 [package.metadata]
 requires-dist = [
     { name = "authlib", specifier = "==1.6.7" },
-    { name = "dbt-protos", specifier = "==1.0.382" },
+    { name = "dbt-protos", specifier = ">=1.0.431" },
     { name = "dbt-sl-sdk", extras = ["sync"], specifier = ">=0.13.2" },
     { name = "dbtlabs-vortex", specifier = "==0.2.0" },
     { name = "fastapi", specifier = ">=0.116.1" },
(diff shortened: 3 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -85,6 +88,63 @@ Or, if you would like to test with Oauth, use a configuration like this:
...
+   )
+   async def my_tool(context: MyToolContext, param: str) -> dict:
+       return await context.fetcher.do_something(param)
+   ```
+5. **Add it to the tool list** (e.g., `DISCOVERY_TOOLS`) in the same module
+6. The registration function (e.g., `register_discovery_tools`) handles context binding and registration
+
+### Adding an MCP App tool
+
+MCP Apps are tools that have an associated interactive UI rendered by the host. They build on top of regular tools with two additions:
```
DevonFulcher: Interesting wording here. I considered dbt MCP to be the MCP App. It seems a bit strange to call an individual tool an app.

CONTRIBUTING.md
```diff
@@ -85,6 +88,63 @@ Or, if you would like to test with Oauth, use a configuration like this:
...
+
+1. **Use `structured_output=True` and `meta`** to link the tool to a UI resource:
```
DevonFulcher: I thought structured output was already used by default: https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#structured-output

CONTRIBUTING.md
```diff
@@ -85,6 +88,63 @@ Or, if you would like to test with Oauth, use a configuration like this:
...
+       description=get_prompt("category/tool_name"),
+       title="My Visualization",
+       read_only_hint=True,
+       structured_output=True,
+       meta={"ui": {"resourceUri": "ui://dbt-mcp/my-app"}},
+   )
+   async def my_viz_tool(context: MyToolContext, param: str) -> MyResultType:
+       ...
+   ```
+   The return type must be a `TypedDict` when `structured_output=True`.
```
DevonFulcher: I think we should use Pydantic models instead if that is compatible.

DevonFulcher (commented):
Nice!

CONTRIBUTING.md
jgiuffrida: https://github.com/dbt-labs/dbt-mcp/blob/153dc9b05d94fafd443f5c240965fa83256c7471/src/dbt_mcp/tools/definitions.py#L24

This was based off of this ☝️ looks like we're defaulting internally to `False`

CONTRIBUTING.md
jgiuffrida: updated!

CONTRIBUTING.md
jgiuffrida: updated the wording - going off the naming [here](https://modelcontextprotocol.io/extensions/apps/overview)

but i'm down w/ being more specific

CONTRIBUTING.md
DevonFulcher: Oh dang, I think that is a mistake we shouldn't default false. Good catch.

CLAUDE.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Commands
+
+- `task test:unit` — run unit tests
+- `task test:integration` — run integration tests (requires dbt Platform credentials)
+- `task install` — install dependencies
+- `task dev` — run server with streamable-http transport
+- `task inspector` — run with MCP Inspector
+- `uv run pytest tests/ --ignore=tests/integration -x -q` — quick unit test run
+
```
jairus-m, line 46: Thoughts on adding `task docs:generate` for syncing new tool defs with repo docs?

jgiuffrida: yup - looks like claude just missed that

jgiuffrida: updated with a few more for completeness

src/dbt_mcp/tools/definitions.py
```diff
@@ -22,6 +22,7 @@ class GenericToolDefinition[NameEnum: Enum]:
     # We haven't strictly defined our tool contracts yet.
     # So we're setting this to False by default for now.
     structured_output: bool | None = False
+    meta: dict[str, Any] | None = None
```
b-per, line 25: Isn't there a risk to confuse ourselves/people between the MCP tool `meta` and the concept of `meta` in dbt that we already rely on in the MCP server?

PR #629, opened 2026-03-09
feat: eager warehouse auth check for browser-based authentication

### LINKED ISSUE #628: dbt CLI commands fail when warehouse token expires with browser-based authentication
## Problem

When a user's dbt profile is configured with browser-based warehouse authentication (e.g., Snowflake `authenticator: externalbrowser`) and their cached token has expired, dbt CLI commands invoked through dbt-mcp fail.

With browser-based auth, the adapter caches tokens in the OS keychain after a successful browser SSO flow. As long as the cached token is valid, dbt commands work normally — including through dbt-mcp. However, when the token expires and the adapter needs to re-authenticate via browser, dbt-mcp cannot complete this flow because:

1. `_run_dbt_command()` runs dbt via `subprocess.Popen` with `stdin=subprocess.DEVNULL`
2. The default 60-second timeout may be too short for a user to complete browser SSO
3. The user gets no feedback that browser authentication is needed

This affects dbt Core and Fusion users with local `profiles.yml` using browser-based SSO for their data warehouse. The dbt Cloud CLI is not affected since warehouse authentication is handled server-side.

### Affected adapters

Any adapter that supports browser-based authentication:
- Snowflake (`authenticator: externalbrowser`)
- BigQuery (OAuth, when not using application-default credentials)
- Databricks (browser-based SSO)

## Proposed solution

Run `dbt debug` as an eager background task during MCP server lifespan startup (following the same pattern as the LSP eager connection). This triggers the browser auth flow early — before the user invokes any CLI tool — caches the token in the OS keychain, and makes subsequent CLI commands work without re-authentication.

If the auth check times out or fails, CLI tools return a clear, actionable error message instead of a cryptic timeout.

## References

- [dbt Community Slack thread](https://getdbt.slack.com/archives/C08NGJYD563/p1772814307006949)

b-per:
## Summary

- Runs `dbt debug` as an eager background task during MCP server startup to trigger and cache browser-based warehouse auth (e.g., Snowflake `externalbrowser`)
- If the cached warehouse token has expired, the browser SSO flow completes before any CLI tool is invoked
- If auth times out or fails, CLI tools return an actionable error message instead of a cryptic timeout

Closes #628

## How it works

1. `WarehouseAuthChecker` is created during `create_dbt_mcp()` when CLI tools are configured
2. In `app_lifespan()`, `dbt debug` runs as `asyncio.create_task()` (same pattern as LSP eager start)
3. The subprocess runs without `stdin=subprocess.DEVNULL` and with a 120s timeout to allow browser SSO
4. On success, the adapter caches the token in the OS keychain; subsequent CLI commands work normally
5. On failure/timeout, `_run_dbt_command()` checks auth status and returns guidance to the user

## Test plan

- [x] Unit tests for all `AuthStatus` transitions (10 tests)
- [x] All existing 396 unit tests still pass
- [x] ruff check, ruff format, mypy all pass
- [ ] Manual test with Snowflake `externalbrowser` profile with expired token

### DIFF
--- a/docs/plans/2026-03-09-browser-auth-check-design.md
+++ b/docs/plans/2026-03-09-browser-auth-check-design.md
@@ -1,0 +1,57 @@
+# Browser-Based Warehouse Authentication Check
+
+## Problem
+
+When dbt is configured to use browser-based warehouse authentication (e.g., Snowflake `authenticator: externalbrowser`), the dbt CLI commands invoked by dbt-mcp fail because:
+
+1. `subprocess.Popen` uses `stdin=subprocess.DEVNULL`, which may block interactive auth flows
+2. The default 60-second timeout may be too short for a user to complete browser SSO
+3. The user gets no feedback that browser authentication is needed
+
+This affects dbt Core and Fusion users with local `profiles.yml` using browser-based SSO for their data warehouse. dbt Cloud CLI is not affected (warehouse auth is handled server-side).
+
+## Solution
+
+Run `dbt debug` as an eager background task during MCP server lifespan (matching the existing pattern used for LSP connection). This:
+
+1. Triggers the browser auth flow early, before the user invokes any CLI tool
+2. Caches the token in the OS keychain (via the dbt adapter)
+3. Makes subsequent CLI commands work without re-authentication
+
+### Architecture
+
+```
+app_lifespan()
+├── register proxied tools (existing)
+├── eager LSP connection (existing)
+└── NEW: eager warehouse auth check
+    └── asyncio.create_task(warehouse_auth_check())
+        └── runs `dbt debug` (no stdin=DEVNULL, longer timeout)
+            ├── success → auth_status = AUTHENTICATED
+            ├── timeout → auth_status = TIMEOUT (with guidance)
+            └── error → auth_status = FAILED (with error message)
+```
+
+### Auth Status Flow
+
+CLI tools check auth status before executing:
+- `AUTHENTICATED` or `NOT_STARTED` (no CLI config): proceed normally
+- `IN_PROGRESS`: wait briefly, then proceed (the auth may complete during the command)
+- `TIMEOUT` / `FAILED`: return actionable error message to the user
+
+### Key Design Decisions
+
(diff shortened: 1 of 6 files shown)


src/dbt_mcp/dbt_cli/auth_check.py
```diff
@@ -0,0 +1,105 @@
...
+        if self._status in (AuthStatus.AUTHENTICATED, AuthStatus.NOT_STARTED):
+            return None
+        if self._status == AuthStatus.IN_PROGRESS:
+            return None  # Let the command proceed; auth may complete during execution
+        if self._status == AuthStatus.TIMEOUT:
+            return self._error_message
+        if self._status == AuthStatus.FAILED:
+            return (
+                f"Warehouse connection check failed during startup. "
+                f"dbt debug output:\n{self._error_message}"
```
jairus-m, line 103: Non-blocker, but line 102 uses f string without interpolation

src/dbt_mcp/dbt_cli/auth_check.py
```diff
@@ -0,0 +1,105 @@
...
+                "Warehouse authentication timed out. "
+                "If your dbt profile uses browser-based authentication "
+                "(e.g., Snowflake externalbrowser), please run "
+                "'dbt debug' in your terminal to authenticate, then restart the MCP server."
+            )
+            logger.warning("Warehouse auth check timed out")
+        except Exception as e:
+            self._status = AuthStatus.FAILED
+            self._error_message = str(e)
+            logger.warning(f"Warehouse auth check error: {e}")
```
jairus-m, line 90: If dbt debug triggers external browser SSO and timeouts or there is an exception somewhere, what happens to the original subprocess? Could it be the case that with the timeout more specifically, the subprocess associated with the original dbt debug command would just linger?

I'm wondering if clean-up to explicitly kill the subprocess in a finally block would be helpful.

docs/plans/2026-03-09-browser-auth-check-design.md
jairus-m, line 1: Oh, this file may be for local docs =D

b-per: Yes, that's my buddy claude code being a bit to eager to share its amazing ideas. I will remove.

PR #631, opened 2026-03-10
Pin mcp SDK to patched fork for stateless task leak fix

wiggzz:
## Why

The upstream MCP Python SDK has a memory leak where stateless HTTP request tasks accumulate in the global task group when clients disconnect mid-request ([modelcontextprotocol/python-sdk#1764](https://github.com/modelcontextprotocol/python-sdk/issues/1764)). This causes unbounded memory growth in long-running MCP servers like our remote dbt MCP proxy.

The fix (request-scoped task groups) also introduces a regression where SSE streams don't close cleanly on shutdown — reverse proxies see connection resets instead of clean HTTP close. A graceful drain step resolves this.

## What

Points the `mcp` dependency at `dbt-labs/mcp-python-sdk@dbt-labs/patched`, which includes:
- Request-scoped task groups for stateless requests (prevents zombie task accumulation)
- Graceful SSE drain on shutdown (terminates all active transports before cancelling the task group)

Upstream PR: [modelcontextprotocol/python-sdk#2145](https://github.com/modelcontextprotocol/python-sdk/pull/2145)

## Notes

This is a temporary pin until the upstream PR is merged and released. Once upstream publishes a release containing the fix, we should revert to the published package.

Drafted by Claude Opus 4.6 under the direction of @wiggzz

### DIFF
--- a/.changes/unreleased/Under the Hood-20260310-mcp-patched-sdk.yaml
+++ b/.changes/unreleased/Under the Hood-20260310-mcp-patched-sdk.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Pin mcp SDK to dbt-labs/patched fork for stateless task leak fix
+time: 2026-03-10T14:00:00.000000-07:00


jairus-m (approved):
Nice! thank u. And awesome to see a PR fix in the actual MCP python sdk repo.

PR #636, opened 2026-03-10
Rename CLAUDE.md to AGENTS.md

DevonFulcher:
## Summary
- Renames `CLAUDE.md` to `AGENTS.md`, the standard name recognized by multiple AI coding agents

## Test plan
- [x] Verify Claude Code still picks up the file as project instructions

> [!NOTE]
> Stacked on #635

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #637, opened 2026-03-10
CLAUDE.md reference in AGENTS.md

DevonFulcher:
## Summary

In [this](https://github.com/dbt-labs/dbt-mcp/pull/636) PR, I had the misconception that Claude Code natively understood AGENTS.md. This PR fixes that to make CLAUDE.md the primary context location.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,1 +1,3 @@
-CLAUDE.md
+# Agent Instructions
+
+See [CLAUDE.md](./CLAUDE.md) for additional context and instructions.


AGENTS.md
jairus-m: Does this need more context? Like doing something like:

```md
# Agent Instructions

See [CLAUDE.md](./CLAUDE.md) for additional context and instructions.

> Also read the contents of `CLAUDE.md` in this repository before proceeding.
```

DevonFulcher: I tried to do a symlink, but I'm not sure it worked.

DevonFulcher: I'll just modify to fit your suggestion.

DevonFulcher: It is a symlink for me locally, and I researched whether git would respect this, but I'm not sure:
image

jairus-m (approved):
Ah was going to test the symlink on my end as well after checking this out but saw it was removed.

We can always test again later =D

PR #638, opened 2026-03-11
Add version awareness to product_docs toolset

mirnawong1:
## Summary

- Detect the user's installed dbt version and use it to filter `<VersionBlock>` tags in docs pages, so users only see documentation relevant to their version
- Surface `dbt_project_version` in tool responses so the LLM can give version-contextualized guidance
- Flag end-of-life (EOL) pages (v1.6 and older) with a warning annotation

## What changed

**VersionBlock filtering (`client.py`):**
- New `filter_version_blocks()` function strips `<VersionBlock>` XML tags from docs markdown based on the user's dbt version
- When version is known: blocks outside the user's version range are removed entirely; matching blocks keep their content with tags stripped
- When version is unknown (`null`): all tags are stripped but content is preserved — clean markdown, no information loss
- Uses iterative innermost-first regex substitution to handle nested `<VersionBlock>` tags correctly

**Version detection (`config.py`, `binary_type.py`):**
- Resolves the dbt version lazily via a `dbt_version_provider` closure on `Config`. The `dbt --version` subprocess runs on the first `product_docs` tool call rather than during `load_config()`, so adapter-heavy installs don't block server startup (cold `dbt --version` can take several seconds).
- The provider memoizes the result for the lifetime of the closure, thread-safe via a per-`Config` lock. Subsequent calls are free.
- Parses version output per binary type (Core `installed:` line, Fusion semver fallback)
- Excludes dbt Cloud CLI from version detection — its `--version` reports the Go binary version (e.g. `0.40.17`), not the dbt engine version, which would incorrectly strip all version-gated content

**Version-aware responses (`types.py`, `tools.py`):**
- `SearchProductDocsResponse` and `GetProductDocPagesResponse` include a `dbt_project_version` field
- `ProductDocPageResponse` has a `version_note` field for EOL flagging
- Both tool prompts include `VERSION AWARENESS` instructions for the LLM, aligned to the actual response shape of each tool

## Validation

Tested across all binary types by changing `DBT_PATH` and `DBT_PROJECT_DIR` env vars and reconnecting:

| Binary type | Detected version | VersionBlock behavior | Result |
|---|---|---|---|
| Fusion (`~/.local/bin/dbt`) | `2.0.0-preview.173` | Showed Fusion-era content, filtered out Core install steps | Pass |
| dbt Core (venv, v1.10) | `1.10.0` | Showed Core content (pip, Docker), filtered out Fusion content | Pass |
| dbt Cloud CLI (`/opt/homebrew/bin/dbt`) | `null` (after fix) | All content kept, tags stripped — graceful fallback | Pass |
| No binary / no `DBT_PROJECT_DIR` | `null` | All content kept, tags stripped — graceful fallback | Pass |

Re-validated end-to-end after the lazy-provider refactor against all three binary types via live MCP calls; behavior unchanged from the table above.

### Limitations

- **dbt Cloud CLI version is opaque:** The Cloud CLI reports its own Go binary version, not the remote dbt engine version. Version detection is intentionally skipped for this binary type.
- **EOL pages are unreachable on live docs:** Older upgrade pages (v1.5, v1.6) return HTTP 404 on the raw `.md` endpoint. The EOL annotation logic is correct but only activatable when currently-supported versions reach EOL.
- **Single-project only:** Version detection reads from `DBT_PROJECT_DIR` and `DBT_PATH` resolved at server startup. The provider memoizes the resolved version for the lifetime of the `Config` instance, so swapping the dbt binary mid-session requires a server restart to re-probe.
- **Version detection is independent of CLI toolset enablement:** The `disable_dbt_cli` setting only controls whether CLI tools are registered — version detection runs regardless (but still requires `DBT_PROJECT_DIR`).

## Checklist
- [ ] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required — No docs changes needed; this is internal MCP behavior


src/dbt_mcp/product_docs/tools.py
```diff
@@ -76,10 +79,17 @@ async def _fetch_page(client: ProductDocsClient, url: str) -> ProductDocPageResp
             error=f"Failed to fetch page: {normalized} ({e})",
         )
 
+    version_note: str | None = None
+    if detect_eol_page(normalized):
+        version_note = "This page is for an end-of-life dbt Core version (v1.6 or older)."
```
DevonFulcher: Honest question: would it be more reliable to determine EOL based on the parsed version number rather than the contents of the page?

src/dbt_mcp/config/config.py
```diff
@@ -161,6 +163,11 @@ def load_config(enable_proxied_tools: bool = True) -> Config:
             lsp_binary_info=lsp_binary_info,
         )
 
+    dbt_version: str | None = None
+    project_yml = settings.dbt_project_yml
+    if project_yml and project_yml.require_dbt_version:
+        dbt_version = parse_dbt_version_minor(project_yml.require_dbt_version)
```
DevonFulcher: Does this hold the full dbt version or just the minor version?

DevonFulcher: It looks like the full version, so I would suggest renaming the function.

src/dbt_mcp/config/dbt_project.py
```diff
@@ -9,3 +11,22 @@ class DbtProjectFlags(BaseModel):
...
+    """Extract the minimum dbt minor version (e.g. ``"1.8"``) from a ``require-dbt-version`` spec.
+
+    Handles common forms: ``">=1.8.0"``, ``"1.8"``, ``[">=1.8.0", "<2.0"]``.
+    Returns ``None`` when the version cannot be determined.
+    """
+    if require_dbt_version is None:
+        return None
+    raw = require_dbt_version if isinstance(require_dbt_version, str) else require_dbt_version[0]
+    match = _DBT_MINOR_VERSION_RE.search(raw)
+    return f"1.{match.group(1)}" if match else None
```
DevonFulcher: This would be a bit flaky if we ever release a 2.0 version. I don't think we should hard-code a `1.` prefix.

src/dbt_mcp/config/dbt_project.py
```diff
@@ -9,3 +11,22 @@ class DbtProjectFlags(BaseModel):
...
+
+
+_DBT_MINOR_VERSION_RE = re.compile(r"1\.(\d+)")
+
+
+def parse_dbt_version_minor(require_dbt_version: str | list[str] | None) -> str | None:
+    """Extract the minimum dbt minor version (e.g. ``"1.8"``) from a ``require-dbt-version`` spec.
+
+    Handles common forms: ``">=1.8.0"``, ``"1.8"``, ``[">=1.8.0", "<2.0"]``.
+    Returns ``None`` when the version cannot be determined.
```
DevonFulcher: Could we include an example of what this function returns in the docstring? Or, use a more specific type than string. For example:
```
@dataclass
class SemVer:
  major: int
  minor: int
  patch: str # pre-releases can contain non-numeric patch versions
```

jairus-m:
> DbtProjectYaml now parses the require-dbt-version field from dbt_project.yml

Just some high-level thoughts on the above! 

In the field and in my experience with past projects, `require-dbt-version` was rarely populated in the config, especially as it's an optional field. Also it does not reliability give the exact version which may be more helpful than a range that comes with a required version.

Would another approach be using the existing dbt CLI executable instead? So using the DBT_PATH and then quite literally running `$DBT_PATH --version` in a subprocess, then parsing that output? The complexity would be parsing the output of dbt core, dbt Cloud CLI, and dbt Fusion. I think there is an existing implementation in the code that already detects the binary type so maybe that could be extended or leveraged.

mirnawong1:
thanks folks! will review this soon and make updates

mirnawong1:
coming back to this next week!

mirnawong1:
changes made:

- dbt version detection: Replaced parsing require-dbt-version from dbt_project.yml (optional field, gives a
  range) with running dbt --version via subprocess to get the actual installed version. Also deduplicated a    
  redundant detect_binary_type call in config.py.                                                              
                                                                                                               
- SemVer type + regex fix: Added a SemVer dataclass (major, minor, patch) to dbt_project.py and updated        
  parse_dbt_version_minor to return it. Fixed the regex to drop the hardcoded 1. prefix so it handles dbt 2.x+.
                                                                                                               
 - EOL page detection: Switched from URL substring matching (/Older versions/) to parsing the version number     
  directly from the URL (e.g. /upgrading-to-v1.6) and comparing against the EOL boundary (≤ 1.6). Folder-based 
  detection kept as a fallback.

src/dbt_mcp/config/dbt_project.py
jairus-m, line 1: Is the related code (parsing `dbt_project.yml`)  within this module still to be used after the pivot to using dbt executables themselves to parse version?

mirnawong1:
moving to draft for now pending [outcome here](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1775724999866189?thread_ts=1770337055.663199&cid=C02NCQ9483C)

b-per:
Is running `dbt --version` blocking for starting the MCP server or not?

The issue with it is that it can be quite slow if there are many adapters installed (not a typical use case, but can happen).

```
 ❯ time dbt --version
WARNING:thrift.transport.sslcompat:using legacy validation callback
Core:
  - installed: 1.10.20
  - latest:    1.11.11 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - databricks: 1.11.6 - Update available!
  - bigquery:   1.9.2  - Update available!
  - snowflake:  1.11.2 - Update available!
  - redshift:   1.9.5  - Update available!
  - postgres:   1.9.0  - Update available!
  - duckdb:     1.9.4  - Update available!
  - spark:      1.10.1 - Up to date!

  At least one plugin is out of date with dbt-core.
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation


dbt --version  2.27s user 0.56s system 43% cpu 6.537 total
```

In this case, it took 6.5 secs to get an answer. 


I am ok with the rest of the PR

src/dbt_mcp/product_docs/client.py
```diff
@@ -86,6 +86,110 @@
...
+    ">>> VERSION NOTICE: This page describes a dbt Core version that has "
+    "reached end-of-life (EOL) and is no longer supported. The information "
+    "may be outdated or differ from current behavior. See the latest version "
+    "support info at https://docs.getdbt.com/docs/dbt-versions/core\n\n---\n\n"
+)
+
+# Matches innermost VersionBlock tags (content must not contain another <VersionBlock).
+# Processing innermost first and iterating handles nested blocks correctly.
+_INNER_VERSION_BLOCK_RE = re.compile(
+    r"<VersionBlock\s+([^>]*)>((?:(?!<VersionBlock).)*?)</VersionBlock>",
```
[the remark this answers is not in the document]
jairus-m: Grepped dbt docs site and there are 0 bare <VersionBlock> instances! Will skip this extra cautious handling for now.

b-per (approved):

PR #639, opened 2026-03-11
DI-3523: Add list_projects tool and project description field

DevonFulcher:
## Why

Multi-project support requires a way to discover accessible projects and meaningful project descriptions for agent disambiguation. This PR adds the foundational pieces to dbt-mcp.

## What

1. **`list_projects()` tool** — new Admin API tool calling `GET /api/v3/accounts/{account_id}/projects/?state=1`, returning id, name, and description for each active project. Added `LIST_PROJECTS` to `ToolName` enum and `ADMIN_API` toolset.
2. **`DbtPlatformProject.description`** — added optional `description: str | None` field to the project model so project descriptions are available downstream.

## Notes

The remaining multi-project plumbing (project_id on tool signatures, ProjectContext resolver, schema stripping/injection) are remote MCP server concerns and will be implemented in the remote MCP repo as part of Milestone 2.

## Refs

- Jira: DI-3523, DI-3526, DI-3527
- Design doc: https://www.notion.so/dbtlabs/Multi-project-dbt-MCP-Implementation-Details-31fbb38ebda7813d8447e0508b604a08

### DIFF
--- a/.changes/unreleased/Security-20260312-155012.yaml
+++ b/.changes/unreleased/Security-20260312-155012.yaml
@@ -1,0 +1,3 @@
+kind: Security
+body: Update minimatch, rollup, and ajv in ui/pnpm-lock.yaml and ai package in examples/ai_sdk_agent to fix known security vulnerabilities
+time: 2026-03-12T15:50:12.420818-05:00
--- a/.changes/unreleased/Under the Hood-20260312-114002.yaml
+++ b/.changes/unreleased/Under the Hood-20260312-114002.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: 'Harden OAuth token handling: eagerly start background refresh, validate token expiry in get_token(), and stop setting settings.dbt_token in the OAuth path'
+time: 2026-03-12T11:40:02.758624-05:00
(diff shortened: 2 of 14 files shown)


.changes/unreleased/Enhancement or New Feature-20260311-155736.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Enhancement or New Feature
+body: 'Add multi-project plumbing: project_id parameter on Discovery/SL tools, list_projects tool, ProjectContext resolver with TTL caching, single-project strip/inject support'
```
jairus-m: Should this be updated since this now only implements list_projects and the project description field?

b-per:
@DevonFulcher @jairus-m 
Would it make sense to add an `include_related` parameter to retrieve some additional info?

For example, it would be great to get the repo used for a project, no?
And maybe some information on the connections, whether SL is configured etc... 

Not too much but enough to give more context.

For example, if someone asks "is there a project for the repo xyz", we could call `list_projects` and get this info.

I can look more into it if it makes sense.

DevonFulcher:
@b-per 

Yes, I think that makes sense. We should be sure not to bloat the context window, but users will typically have far fewer results from `list_projects` than, for example, `list_models`. If we are worried about that, we could have a `get_project_details`, but I'm not sure if that is necessary.

PR #642, opened 2026-03-12
Harden OAuth token handling

DevonFulcher:
## Why

1. **Prevent stale token access** - The OAuth credential setup path was writing the initial access token into `settings.dbt_token`, creating a static copy that could be accidentally used instead of the refreshing `token_provider.get_token()`. This mutation has been removed.

2. **Eager background refresh** - The token refresh background worker was only started lazily on the first `get_token()` call. If the token expired before any tool was invoked, it would go stale silently. The background refresh is now started immediately after OAuth credential setup.

3. **Inline token validation** - `get_token()` previously returned whatever token it had without checking expiry. Now it validates the token's `expires_at` and performs a synchronous refresh if the token is expired or about to expire (within a 30-second buffer). If the refresh fails, it raises a clear error with instructions to re-authenticate.

## What

- **`OAuthTokenProvider`** (`oauth/token_provider.py`):
  - Consolidated duplicate `_refresh_token_sync()` and `async _refresh_token()` into a single `_refresh_token()` method (all operations are synchronous; the async wrapper was unnecessary)
  - `get_token()` now validates expiry and refreshes inline as a safety net
  - Removed `refresh_started` flag (no more lazy initialization)
  - Added `TOKEN_EXPIRY_BUFFER_SECONDS` constant (30s)

- **`CredentialsProvider`** (`config/settings.py`):
  - Removed the line that set `settings.dbt_token` from the OAuth access token (was marked with a TODO)
  - Now calls `token_provider.start_background_refresh()` eagerly after construction
  - Passes `has_token=True` to `validate_settings()` in the OAuth path

- **`validate_dbt_platform_settings`** / **`validate_settings`** (`config/settings.py`):
  - Added `has_token` keyword parameter to skip the `dbt_token` check when a token provider supplies the token

- **`config_providers.py`**:
  - Removed `assert settings.dbt_token` guards from all config providers
  - `SemanticLayerConfig` now holds a `token_provider` instead of a static `token` string

- **`semantic_layer/client.py`**:
  - Uses `config.token_provider.get_token()` instead of `config.token`

## Notes

### DIFF
--- a/.changes/unreleased/Security-20260312-155012.yaml
+++ b/.changes/unreleased/Security-20260312-155012.yaml
@@ -1,0 +1,3 @@
+kind: Security
+body: Update minimatch, rollup, and ajv in ui/pnpm-lock.yaml and ai package in examples/ai_sdk_agent to fix known security vulnerabilities
+time: 2026-03-12T15:50:12.420818-05:00
--- a/examples/ai_sdk_agent/package.json
+++ b/examples/ai_sdk_agent/package.json
@@ -2,7 +2,7 @@
   "dependencies": {
     "@ai-sdk/openai": "^1.0.0",
     "@modelcontextprotocol/sdk": "^1.12.0",
-    "ai": "^4.0.0"
+    "ai": "^5.0.52"
   },
   "description": "Example of using Vercel AI SDK with dbt-mcp",
   "devDependencies": {
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,20 +1,20 @@
 {
   "dependencies": {
-    "react": "^19.1.1",
-    "react-dom": "^19.1.1"
+    "react": "^19.2.4",
+    "react-dom": "^19.2.4"
   },
   "devDependencies": {
-    "@eslint/js": "^9.33.0",
-    "@types/react": "^19.1.10",
-    "@types/react-dom": "^19.1.7",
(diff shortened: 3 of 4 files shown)


src/dbt_mcp/config/settings.py
```diff
@@ -612,15 +611,27 @@ async def get_credentials(self) -> tuple[DbtMcpSettings, TokenProvider]:
...
-            self.token_provider = OAuthTokenProvider(
+            token_provider = OAuthTokenProvider(
                 access_token_response=dbt_platform_context.decoded_access_token.access_token_response,
                 dbt_platform_url=dbt_platform_url,
                 context_manager=dbt_platform_context_manager,
             )
-            validate_settings(self.settings)
+            # Eagerly start the background refresh so the token stays fresh
+            # even before the first get_token() call.
+            token_provider.start_background_refresh()
```
wiggzz: Any thoughts on moving this into the token provider __init__? That feels like it would keep the background refresh logic of the token provider encapsulated there, versus having consumers of that class need to know to start the background refresh process.

DevonFulcher: Good call. I moved this to an `OAuthTokenProvider.create()` method, which is async. `__init__()` can't be async, so using an async class method is better to ensure it is called within an async event loop.

Issue #583, opened 2026-02-04
[Feature]: Get host prefix from dbt platform when MULTICELL_ACCOUNT_PREFIX or DBT_HOST_PREFIX are unset


DevonFulcher:
### Is your feature request related to a problem?

With env var (as opposed to oauth) configuration, users need to set a `DBT_HOST_PREFIX` env var (or `MULTICELL_ACCOUNT_PREFIX` for older configurations) when their account requires one. This is so that we can correctly construct SL and Discovery URLs. However, this is often overlooked when users set up dbt MCP, and it adds another field that makes configuration more complex.

### Describe the solution you'd like

We should update env var configuration to behave similarly to oauth configuration, such that the prefix is [fetched from dbt platform](https://github.com/dbt-labs/dbt-mcp/blob/fb97ee93cbd842ddaa720f23d838a6c8626d77b0/src/dbt_mcp/oauth/fastapi_app.py#L64) and [applied to settings](https://github.com/dbt-labs/dbt-mcp/blob/fb97ee93cbd842ddaa720f23d838a6c8626d77b0/src/dbt_mcp/config/settings.py#L596). This would allow users to put their whole access URL in `DBT_HOST` rather than segment the URL across two env vars. This solution should be fully backwards compatible for configurations that use `DBT_HOST_PREFIX` or `MULTICELL_ACCOUNT_PREFIX`.

### Describe alternatives you've considered

Fetching the prefix from dbt platform is the most reliable solution as opposed to trying to parse out the prefix from `DBT_HOST`.

jairus-m:
@DevonFulcher - to continue on the work on this, should I wait for all the multi-project feature work to conclude?

DevonFulcher:
> [@DevonFulcher](https://github.com/DevonFulcher) - to continue on the work on this, should I wait for all the multi-project feature work to conclude?

@jairus-m I think the two initiatives are fairly independent. So, we can work on this at any time.

PR #650, opened 2026-03-16
Address auth issues with misconfigured env vars

### LINKED ISSUE #645: [Bug]: Silent fallback to OAuth with misconfigured env vars
### What happened?

Multicell users who set `MULTICELL_ACCOUNT_PREFIX`, `DBT_HOST`, and `DBT_TOKEN` but accidentally omit `DBT_PROD_ENV_ID` fallback to the OAuth browser flow (prod env id is necessary to use static token). 

This fallback is expected given the missing prod env id. However, the OAuth flow itself then fails with a 404 because the OAuth code path leads to a URL that does not use `MULTICELL_ACCOUNT_PREFIX` and builds a URL without the needed prefix. 

In addition to this, the fallback is silent without mentioning the configured `DBT_TOKEN` and falling back to OAuth due to a missing `DBT_PROD_ENV_ID`.

And in playing around, I identified silent breaks when we have a double prefix when we do something like
```
DBT_HOST=ab123.us1.dbt.com
MULTICELL_ACCOUNT_PREFIX=ab123
```

The goal is to fix three things:
1. If we fallback to OAuth, construct the correct URL with prefix if MULTICELL_ACCOUNT_PREFIX is set
2. Have better logs calling out fallback and the need for a prod env id for static token auth
3. Correctly deal with the double-set prefix config silent error

Then after this, address #583

### Steps to Reproduce

  1. Set `DBT_HOST`, `MULTICELL_ACCOUNT_PREFIX`, `DBT_TOKEN` (then omit `DBT_PROD_ENV_ID`)
  2. Start the MCP server
  3. OAuth browser popup opens → 404 (pre-fix)

### Deployment

Local MCP server

### Environment

_No response_

### Relevant log output

```shell

```

jairus-m:
## Summary

Addresses three bugs as detailed in the issue.
- When `DBT_TOKEN` was set but platform settings were incomplete (triggering OAuth fallback) the token was silently ignored with no warning to the user
- When `DBT_HOST=us1.dbt.com` + `MULTICELL_ACCOUNT_PREFIX=ab123` are set, the OAuth redirect URL was built from the host alone and would not have the prefix
-  `If DBT_HOST=ab123.us1.dbt.com` and `MULTICELL_ACCOUNT_PREFIX=ab123` were both set, the prefix was added again (double prefix)

This is precursor work for #583.

## Related Issues

Closes #645

## Checklist

## Additional Notes

I manually tested the different scenarios locally !

NOTE:  This PR was drafted and reviewed by Claude Code (claude-sonnet-4-6).

### DIFF
--- a/src/dbt_mcp/config/settings.py
+++ b/src/dbt_mcp/config/settings.py
@@ -197,11 +197,21 @@ class DbtMcpSettings(BaseSettings):
         Prevents double-prefixing when DBT_HOST already contains the account prefix
         (e.g. 'ab123.us1.dbt.com') and a prefix env var is also set to 'ab123'.
         Returns None if and only if actual_host is None.
+        On mismatch (host has a different embedded prefix), logs a warning and returns
+        the host unchanged to avoid silently stripping the wrong label.
         """
         host = self.actual_host
         if host is None:
             return None
-        return parse_host_prefix(host, self.actual_host_prefix).base_host
+        result = parse_host_prefix(host, self.actual_host_prefix)
+        if result.mismatched_prefix is not None:
+            logger.warning(
+                f"DBT_HOST ('{host}') appears to contain a different account prefix "
+                f"('{result.mismatched_prefix}') than the configured prefix '{self.actual_host_prefix}'. "
+                "This may result in incorrect URL construction."
+            )
+            return host
+        return result.base_host
 
     @property
     def dbt_project_yml(self) -> DbtProjectYaml | None:
@@ -505,7 +515,7 @@ async def get_dbt_platform_context(
 
 @dataclass(frozen=True)
 class HostPrefixResult:
-    base_host: str  # Host with matching prefix stripped
+    base_host: str  # Host with first label stripped; matches configured prefix when prefix_embedded=True, or suggested base host on mismatch
     prefix_embedded: bool  # True if configured prefix was at start of host
     mismatched_prefix: (
         str | None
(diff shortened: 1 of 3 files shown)


src/dbt_mcp/config/settings.py
```diff
@@ -469,20 +507,53 @@ async def get_dbt_platform_context(
...
+    Handles three cases:
+    - Prefix set, not yet in host: prepend it (e.g. 'us1.dbt.com' + 'ab123' → 'https://ab123.us1.dbt.com')
+    - Prefix set, already in host: use host as-is (no double prefix)
+    - No prefix: use host as-is
+    """
+    if actual_host_prefix:
+        if actual_host.startswith(f"{actual_host_prefix}."):
+            return f"https://{actual_host}"
+        labels = actual_host.split(".")
+        if len(labels) >= 4 and labels[0] != actual_host_prefix:
```
DevonFulcher: Traditionally, I have been hesitant to parse the URL since it can come in many forms: https://docs.getdbt.com/docs/cloud/about-cloud/access-regions-ip-addresses?version=2.0. However, I think this should be relatively safe.

jairus-m: Yes - the upstream stripping of the host URL and the fairly limited input space for multicell accounts that do adhere to prefix.region.dbt.com may be safe enough haha. but can totally revisit this

DevonFulcher: I think it is good for now!

PR #653, opened 2026-03-16
Cache JWKS clients per platform URL

DevonFulcher:
## Why

Every call to `fetch_jwks_and_verify_token` creates a new `PyJWKClient`, which fetches the JWKS endpoint over HTTP. When tokens are refreshed frequently, this causes redundant network requests to the same well-known endpoint.

## What

- Add a module-level cache (`_jwks_clients`) that stores `PyJWKClient` instances keyed by JWKS URL
- `_get_jwks_client(jwks_url)` returns the cached client or creates a new one
- `_clear_jwks_cache()` exposed for test teardown
- Update `fetch_jwks_and_verify_token` to use the cached client
- Add unit tests verifying same-URL reuse and different-URL isolation

### DIFF
--- a/src/dbt_mcp/oauth/token.py
+++ b/src/dbt_mcp/oauth/token.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from typing import Any
 
 import jwt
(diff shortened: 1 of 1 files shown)


DevonFulcher:
Dropping — the caching doesn't provide meaningful benefit since this is called ~once per hour.

PR #655, opened 2026-03-16
Consolidate OAuth token refresh into shared function

DevonFulcher:
## Why

The OAuth token refresh logic was duplicated in two places:
- `token_provider.py` (inline/background refresh in `OAuthTokenProvider._refresh_token`)
- `settings.py` (startup refresh in `_try_refresh_token`)

Both independently created an `OAuth2Session`, called `refresh_token`, and converted the response via `dbt_platform_context_from_token_response`. This duplication means any fix or change to refresh behavior must be applied in two places.

## What

- Created `oauth/refresh.py` with a single `refresh_oauth_token()` function that encapsulates the shared OAuth refresh logic (session creation, token refresh, response conversion)
- Updated `OAuthTokenProvider._refresh_token` in `token_provider.py` to call `refresh_oauth_token()` instead of managing its own `OAuth2Session`
- Updated `_try_refresh_token` in `settings.py` to call `refresh_oauth_token()` instead of duplicating the same logic
- Removed now-unused `authlib` and `OAUTH_CLIENT_ID` imports from both modules
- Updated test patch targets to mock the new shared function
- Added unit tests for `refresh_oauth_token()` covering success and exception propagation


src/dbt_mcp/oauth/refresh.py
```diff
@@ -0,0 +1,34 @@
...
+from authlib.integrations.requests_client import OAuth2Session
+
+from dbt_mcp.oauth.client_id import OAUTH_CLIENT_ID
+from dbt_mcp.oauth.dbt_platform import (
+    DbtPlatformContext,
+    dbt_platform_context_from_token_response,
+)
+
+logger = logging.getLogger(__name__)
+
```
jairus-m: Non-blocker! just unused logger

theyostalservice (approved):
seems reasonable, and it's nice to have this deduped.  Thanks, Devon!

PR #658, opened 2026-03-17
feat: add list_projects_and_environments discovery tool

DevonFulcher:
## Why

Multi-project mode (Server B) needs a discovery tool that agents call first to learn what project IDs are available before calling any other project-scoped tools.

## What

- Add `list_projects_and_environments` tool that returns all accessible projects with their prod and dev environments
- Add `ProjectToolContext` dataclass for dependency injection following existing patterns
- Add `PROJECT` toolset with enable/disable settings
- Create `server_multiproject.py` stub that registers the discovery tool on Server B
- Wire up `main.py` to use the multiproject server when `DBT_MCP_MULTI_PROJECT_ENABLED=true`
- Add `multiproject_only_tools` set so Server A tests properly account for Server B-only tools
- Update README/diagram auto-generation, tool name enum, toolset mappings, and readme mappings

## Notes

- Stacked on #657 (Phase 1: extract project/environment helpers)
- The multiproject server is minimal — it only registers the discovery tool for now. Later phases will add project-scoped versions of existing tools.


DevonFulcher:
Closing — the existing `list_projects` tool in dbt_admin already covers project discovery. Server B will register the existing tool rather than a new one.

PR #659, opened 2026-03-17
feat: add multi-project discovery tools

DevonFulcher:
## Why

Multi-project mode (Server B) needs discovery tools that can target different projects dynamically. This requires each tool to accept a `project_id` parameter and resolve project-specific config at call time, rather than using the static config from environment variables.

### New file: `src/dbt_mcp/discovery/tools_multiproject.py`
- 18 multi-project discovery tools mirroring every existing Server A discovery tool
- Each tool takes `project_id: int` as a required first parameter
- Uses `_resolve_project_environments()` to resolve per-project config
- Passes `config_override` through to all client fetcher methods
- Includes `register_multiproject_discovery_tools()` for Server B registration

### Modified: `src/dbt_mcp/discovery/client.py`
- Added `config_override: DiscoveryConfig | None = None` parameter to all fetcher methods
- When `config_override` is provided, it takes precedence over `config_provider.get_config()`

### Modified: `src/dbt_mcp/config/config_providers.py`
- Added `_resolve_project_environments()` helper for resolving prod/dev environments for a given project

### Modified: `src/dbt_mcp/config/settings.py` and `src/dbt_mcp/config/config.py`
- Added `multi_project_enabled` field (from `DBT_MCP_MULTI_PROJECT_ENABLED` env var) to `DbtMcpSettings` and `Config`

### Modified: `src/dbt_mcp/main.py`
- Removed raw env-var check with `NotImplementedError`; routing is now handled in `server.py`

### Modified: `src/dbt_mcp/mcp/server.py`
- Added `register_multi_project_dbt_mcp()` that registers multi-project discovery tools
- Extracted `register_dbt_mcp_tools()` for the single-project path
- `create_dbt_mcp()` routes to one of the two paths based on `config.multi_project_enabled`
- Proxied tools are skipped in multi-project mode

### Test fixes
- Updated mock signatures in `test_pagination.py`, `test_model_performance_fetcher.py`, and `test_resource_details_fetcher.py` to match new `config_override` parameter

## Notes

- This PR depends on Phase 1 infrastructure (PR #657) for the `_resolve_project_environments()` implementation
- Existing Server A tools are completely unchanged
- All 458 unit tests pass, all pre-commit checks (ruff, mypy) pass

Drafted by Claude Sonnet 4.6 under the direction of @DevonFulcher


src/dbt_mcp/discovery/tools_multiproject.py
```diff
@@ -0,0 +1,589 @@
...
+from dataclasses import dataclass
+
+from mcp.server.fastmcp import FastMCP
+from pydantic import Field
+
+from dbt_mcp.config.config_providers import (
+    ConfigProvider,
+    DiscoveryConfig,
+    _resolve_project_environments,
+)
```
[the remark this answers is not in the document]
DevonFulcher: Addressed — renamed `_resolve_project_environments` to `resolve_project_environments` in `config_providers.py` to make it a public function, and updated all call sites accordingly.

*(Comment created with AI assistance)*

src/dbt_mcp/discovery/client.py
```diff
@@ -436,8 +443,9 @@ async def fetch_paginated(
         self,
         query: str,
         variables: dict[str, Any],
+        config_override: DiscoveryConfig | None = None,
```
wiggzz: Instead of a config_override, it may be cleaner to instead switch the client to _not_ embed the environment_id and just pass an environment_id here. Then, in the single-project tools, pass the environment_id from the context.

src/dbt_mcp/discovery/tools_multiproject.py
```diff
@@ -0,0 +1,589 @@
...
+async def get_mart_models(
+    context: MultiProjectDiscoveryToolContext,
+    project_id: int = PROJECT_ID_FIELD,
+) -> list[dict]:
+    config = await _resolve_discovery_config_for_project(context, project_id)
+    mart_models = await context.models_fetcher.fetch_models(
+        model_filter={"modelingLayer": "marts"},
+        config_override=config,
+    )
+    return [m for m in mart_models if m["name"] != "metricflow_time_spine"]
```
wiggzz, line 80: Similarly instead of copying and pasting, can we make the other single-project tool call the multi-project tool and pass the project_id from the context?

wiggzz: And, that may allow re-use of the context. E.g. we have a DiscoveryToolContext which is used here which does not bind the environment or project id. Then for the single project tool set, we create a wrapper context like
```
class SingleProjectDiscoveryToolContext:
  discovery_tool_context: DiscoveryToolContext
  project_id: int
```
that way we don't need to have so much duplication.

But maybe there is a reason you did things this way.

DevonFulcher: So, recently, I have questioned whether or not we want to maintain both single-project & multi-project versions of the server. That is why I opted to copy-paste without much code sharing with the single-project version. My intent is to just get it working to test out the e2e integration with ChatGPT as quickly as possible.

I think there is an opportunity for code consolidation, but I don't think we want the single-project tool to call the multi-project tool. That would result in an unnecessary network hop to get the prod env ID for the project, something that is already configured for the single-project server. It could be beneficial for the multi-project server to call the single-project server, though.

I think there are other code consolidation efforts we could explore at the decorator or adapter level, so that we don't have to redefine tools at all. However, I think I want to take this on later, after we have the multi-project MCP working end-to-end through remote MCP.

src/dbt_mcp/discovery/client.py
DevonFulcher: I think that would require a lot of refactoring. The `environment_id` isn't a field on `DiscoveryToolContext`. In the single-project version, the env ID is fetched from `DefaultDiscoveryConfigProvider`, which ultimately pulls the env ID based on the local Oauth credential selection. So, we could decouple the env ID from the `DiscoveryConfig` and place it on `DiscoveryToolContext`, but that would be a change that impacts a large part of the local Oauth code and may not be the behavior we want for local Oauth anyway. So, let's defer that.

I think a simpler change, and maybe this is what you meant, would be to call `config_provider.get_config()` higher up the call stack, at the tool definition level, and pass in the config (including the env ID). In this way, we would always pass the config here rather than conditionally, making this value non-nullable. I will experiment with that.

DevonFulcher: Okay, I addressed this in my latest commits.

src/dbt_mcp/discovery/tools_multiproject.py
wiggzz: > That would result in an unnecessary network hop to get the prod env ID for the project, something that is already configured for the single-project server.

Yea. Maybe we just need a shared method under the hood then - one that takes environment_id. The single-project tool uses it directly, the multi-project tool resolves the environment_id before calling it.

wiggzz: But maybe this is fine, the fetchers are basically this. I'm just confused by all the configs / contexts / config_provider etc...

src/dbt_mcp/discovery/client.py
```diff
@@ -354,11 +354,13 @@ class GraphQLQueries:
 
 
 class MetadataAPIClient:
```
wiggzz: You can probably get rid of this client now since it doesn't store anything - the method can either become static or just a top level function.

src/dbt_mcp/discovery/tools_multiproject.py
wiggzz: Ok yea in it's current state this is good I think.

src/dbt_mcp/discovery/tools_multiproject.py
```diff
@@ -0,0 +1,589 @@
...
+    exposures_fetcher: ExposuresFetcher
+    sources_fetcher: SourcesFetcher
+    macros_fetcher: MacrosFetcher
+    resource_details_fetcher: ResourceDetailsFetcher
+    lineage_fetcher: LineageFetcher
+    model_performance_fetcher: ModelPerformanceFetcher
+
+    def __init__(
+        self,
+        config_provider: ConfigProvider[DiscoveryConfig],
```
wiggzz: This looks unused now

src/dbt_mcp/discovery/tools_multiproject.py
```diff
@@ -0,0 +1,589 @@
...
+
+PROJECT_ID_FIELD = Field(
+    description="The dbt Cloud project ID to query. "
+    "Use list_projects_and_environments to discover available project IDs.",
+)
+
+
+@dataclass
+class MultiProjectDiscoveryToolContext:
+    credentials_provider: CredentialsProvider
```
wiggzz: Given the only difference here is the addition of the credentials_provider, should we just add the credentials provider in the DiscoveryToolContext and share the same context between the single project and multi-project?

DevonFulcher: Done!

PR #660, opened 2026-03-17
Add multi-project semantic layer tools

DevonFulcher:
## Why

To support multi-project scenarios where an AI assistant needs to query metrics across different dbt projects, we need semantic layer tools that accept a `project_id` parameter to dynamically resolve the correct environment configuration.

## What

Adds 6 multi-project semantic layer tools that mirror the existing single-project tools but accept `project_id` as a required first parameter:

- `list_metrics`
- `list_saved_queries`
- `get_dimensions`
- `get_entities`
- `query_metrics`
- `get_metrics_compiled_sql`

Each tool resolves the project's production environment via `get_config_for_project()` (from Phase 1) and passes the resulting config as a `config_override` to the existing `SemanticLayerFetcher` methods.

Key changes:
- **`client.py`**: Added `config_override` parameter to all `SemanticLayerFetcher` methods and a `_resolve_client()` helper for SDK-based operations
- **`tools_multiproject.py`**: New file with the 6 multi-project tool definitions and registration function
- **`tool_names.py` / `toolsets.py`**: New `ToolName` entries and `MULTI_PROJECT_SEMANTIC_LAYER` toolset
- **`config.py` / `settings.py`**: Enable/disable configuration for the new toolset (disabled by default)
- **Tests**: Updated existing tests to account for the new toolset

## Notes

- These tools are intended for Server B (multi-project server) only and are not registered on the default Server A
- The existing single-project tools are unchanged
- Stacked on #657 (Phase 1: core infrastructure)

### DIFF
--- a/src/dbt_mcp/mcp/server.py
+++ b/src/dbt_mcp/mcp/server.py
@@ -1,6 +1,5 @@
 import asyncio
 import logging
-import os
 import time
 import uuid
 from collections.abc import AsyncIterator, Callable, Sequence
@@ -103,14 +102,6 @@ class DbtMCP(FastMCP):
         return result
 
 
-def _multi_project_enabled() -> bool:
-    return os.environ.get("DBT_MCP_MULTI_PROJECT_ENABLED", "").lower() in (
-        "true",
-        "1",
-        "yes",
-    )
-
-
 @asynccontextmanager
 async def app_lifespan(server: FastMCP[Any]) -> AsyncIterator[bool | None]:
     if not isinstance(server, DbtMCP):
@@ -120,7 +111,10 @@ async def app_lifespan(server: FastMCP[Any]) -> AsyncIterator[bool | None]:
         # register proxied tools inside the app lifespan to ensure the StreamableHTTP client (specific
         # to dbt Platform connection) lives on the same event loop as the running server
         # this avoids anyio cancel scope violations (see issue #498)
-        if server.config.proxied_tool_config_provider and not _multi_project_enabled():
+        if (
+            server.config.proxied_tool_config_provider
+            and not server.config.multi_project_enabled
+        ):
             logger.info("Registering proxied tools")
             await register_proxied_tools(
                 dbt_mcp=server,
@@ -193,9 +187,7 @@ async def create_dbt_mcp(config: Config) -> DbtMCP:
         lifespan=app_lifespan,
     )
 
-    multi_project_enabled = _multi_project_enabled()
-
-    if multi_project_enabled:
+    if config.multi_project_enabled:
         logger.info("DBT_MCP_MULTI_PROJECT_ENABLED=true -> Multi-project mode")
         await register_multi_project_dbt_mcp(dbt_mcp, config)
     else:
--- a/src/dbt_mcp/semantic_layer/tools_multiproject.py
+++ b/src/dbt_mcp/semantic_layer/tools_multiproject.py
@@ -1,4 +1,3 @@
-import logging
 from dataclasses import dataclass
 
 from dbtsl.api.shared.query_params import GroupByParam
@@ -25,8 +24,6 @@ from dbt_mcp.tools.definitions import GenericToolDefinition, dbt_mcp_tool
 from dbt_mcp.tools.register import register_tools
 from dbt_mcp.tools.tool_names import ToolName
 from dbt_mcp.tools.toolsets import Toolset
-
-logger = logging.getLogger(__name__)
 
 
 @dataclass
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -9,8 +9,8 @@ from dbt_mcp.config.config_providers import (
 )
 from dbt_mcp.config.settings import (
     CredentialsProvider,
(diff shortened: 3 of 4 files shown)


src/dbt_mcp/main.py
jairus-m, line 1: nice! this in src/dbt_mcp/mcp/server.py instead 🙏

src/dbt_mcp/semantic_layer/tools_multiproject.py
```diff
@@ -0,0 +1,245 @@
...
+async def list_metrics(
+    context: MultiProjectSemanticLayerToolContext,
+    project_id: int,
+    search: str | None = None,
+) -> list[MetricToolResponse]:
+    config = await context.semantic_layer_config_provider.get_config_for_project(
+        project_id
+    )
+    fetcher = context.fetcher_for_config(config)
+    return await fetcher.list_metrics(search=search)
```
jairus-m, line 66: Not sure if this is worth doing, but since these lines are repeated in all the tools, could it there be a async method in `MultiProjectSemanticLayerToolContext` to get the fetcher just to consolidate a bit?

```python

class MultiProjectSemanticLayerToolContext:
    ...
    get_fetcher(project_id: int) -> SemanticLayerFetcher:
       config = await context.semantic_layer_config_provider.get_config_for_project(
            project_id
        )
        return context.fetcher_for_config(config)

# later on
async def list_metrics(
    context: MultiProjectSemanticLayerToolContext,
    project_id: int,
    search: str | None = None,
) -> list[MetricToolResponse]:
    # call async method
    fetcher = context.get_fetcher(project_id)
    return await fetcher.list_metrics(search=search)
```

src/dbt_mcp/main.py
jairus-m, line 1: nice! this in src/dbt_mcp/mcp/server.py instead 🙏

src/dbt_mcp/semantic_layer/tools_multiproject.py
```diff
@@ -0,0 +1,245 @@
...
+
+@generic_dbt_mcp_tool(
+    description=get_prompt("semantic_layer/list_metrics"),
+    name_enum=ToolName,
+    name="list_metrics",
+    title="List Metrics",
+    read_only_hint=True,
+    destructive_hint=False,
+    idempotent_hint=True,
+)
```
jairus-m: if these multi proj SL tools share the same exact names as the single proj SL tools (and are registered in the same path, just replaced when opted into), is the `@generic_dbt_mcp_tool` decorator still necessary? Trying to understand the difference for usage with `@dbt_mcp_tool`

src/dbt_mcp/mcp/server.py
```diff
@@ -101,6 +103,14 @@ async def call_tool(
         return result
 
 
+def _multi_project_enabled() -> bool:
+    return os.environ.get("DBT_MCP_MULTI_PROJECT_ENABLED", "").lower() in (
```
b-per: Shouldn't we use pydantic-settings like we do in other parts of the code?

It already contains the logic for converting strings to bools (incl things like `y` and `t` as well

src/dbt_mcp/semantic_layer/tools_multiproject.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def fetcher_for_config(self, config: SemanticLayerConfig) -> SemanticLayerFetcher:
+        return SemanticLayerFetcher(
+            config_provider=self.semantic_layer_config_provider,
+            client_provider=self._client_provider,
+            config=config,
+        )
+
+
+@generic_dbt_mcp_tool(
```
b-per: I might have missed something, but why don't we update the existing tools instead of creating new ones?

Is it to avoid breaking changes?

src/dbt_mcp/mcp/server.py
DevonFulcher: Yes, I'll make that update. I didn't do that initially because I view the intention of this env var to be undocumented, experimental, and temporary. Regardless, it can still live in pydantic-settings.

src/dbt_mcp/semantic_layer/tools_multiproject.py
DevonFulcher: That could result in breaking changes, yes. However, my intent with a separate server was mainly to keep this experimental to start with. Once we are satisfied, we can merge the two servers into one.

src/dbt_mcp/semantic_layer/tools_multiproject.py
DevonFulcher: Oh good catch, ya I think we can use `@dbt_mcp_tool` here instead.

src/dbt_mcp/semantic_layer/tools_multiproject.py
DevonFulcher: Fixed!

src/dbt_mcp/semantic_layer/tools_multiproject.py
DevonFulcher: `generic_dbt_mcp_tool` is mainly just for our internal usage.

PR #684, opened 2026-03-31
Using Admin API Client for dynamic multi-project server config

DevonFulcher:
## Summary

Previously, when we fetched environment information for the multi-project version of dbt MCP, we used a separate httpx client from the one provided by the `DbtAdminApiClient`. This won't work for how we handle internal traffic for remote MCP. Additionally, this PR updates how configuration is fetched to make it pluggable for different scenarios in the remote MCP case.

I've tested the discovery tools for both the single-project and multi-project servers with `task client`.

## Checklist

### DIFF
--- a/src/dbt_mcp/dbt_admin/client.py
+++ b/src/dbt_mcp/dbt_admin/client.py
@@ -261,16 +261,6 @@ class DbtAdminAPIClient:
             for p in data
         ]
 
-    async def get_project_details(
-        self, account_id: int, project_id: int
-    ) -> dict[str, Any]:
-        """Get details for a specific project."""
-        result = await self._make_request(
-            "GET",
-            f"/api/v3/accounts/{account_id}/projects/{project_id}/",
-        )
-        return result.get("data", {})
-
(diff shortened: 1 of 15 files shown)


src/dbt_mcp/dbt_admin/client.py
```diff
@@ -52,8 +52,103 @@ async def _make_request(
...
+            if (
+                environment.deployment_type
+                and environment.deployment_type.lower() == "development"
+            ):
+                dev_environment = DbtPlatformEnvironment(
+                    id=environment.id,
+                    name=environment.name,
+                    deployment_type=environment.deployment_type,
+                )
+                break
```
alan-andrade, line 104: Can there be more than one development environment so this just returns the first one?

DevonFulcher: Good question, but there can only be one development environment: https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/snippets/_cloud-environments-info.md?plain=1#L11

PR #698, opened 2026-04-07
ci: add HOL skill-publish validate workflow

internet-dot:
This adds a small validate-only CI check for the existing skill metadata in this repo.

It only adds one workflow file, does not publish anything, and does not need secrets.

Why it looked like a fit here:
- A MCP (Model Context Protocol) server for interacting with dbt
- Tools for executing and generating SQL on dbt Platform infrastructure

It runs schema and trust validation for `.claude/skills/release-dbt-mcp` and leaves runtime code, release flow, and publish credentials alone.
Permissions: read-only. No secrets. No publish. No runtime changes.

If you'd like it folded into an existing workflow or pointed at a different path, I can adjust the branch.

### DIFF
--- a/.github/workflows/hol-skill-validate.yml
+++ b/.github/workflows/hol-skill-validate.yml
@@ -1,0 +1,30 @@
+name: HOL Skill Validate
+
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    branches:
+      - main
+      - master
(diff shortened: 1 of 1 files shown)


jairus-m:
Thanks for opening this! We don't publish the skill in this repo or nned a dedicated CI check for this purpose. Will close in light of that 👍

PR #701, opened 2026-04-07
feat: populate dbt_cloud_account_identifier in telemetry events

eknoll:
## Summary

Populate the `dbt_cloud_account_identifier` field in telemetry events, which was previously hardcoded to an empty string.

## What Changed

- Added `account_identifier` field to `DbtMcpSettings`
- Added `get_account()` method to `DbtAdminAPIClient` (`GET /api/v2/accounts/{account_id}/`)
- Resolve account identifier during credential setup:
  - **OAuth path**: Extract from JWT claim `https://dbt.com/account_identifier`, falling back to Admin API
  - **PAT path**: Fetch from Admin API
  - Both paths degrade gracefully — if resolution fails, the field remains empty
- Use the resolved value in the `VortexTelemetryDbtCloudContext` telemetry event

## Why

We populate the `account_id` field rather than hardcoding an empty string to conform to telemetry collector expectations.

## Related Issues

N/A

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- No documentation changes needed; this is an internal telemetry change

## Additional Notes

**Testing evidence:**
- `task test:unit`: 469 tests passed
- `ruff check`, `ruff format`, `mypy`: all clean

Tests added:
- `test_get_account` — verifies the new Admin API client method
- `test_emit_tool_called_event_account_identifier_none` — verifies empty string fallback when identifier is unresolved
- `test_oauth_extracts_identifier_from_jwt_claims` — verifies JWT claim extraction
- `test_oauth_falls_back_to_admin_api` — verifies API fallback when JWT claim is absent
- `test_pat_fetches_identifier_from_admin_api` — verifies PAT path resolution
- `test_api_failure_does_not_break_credentials` — verifies graceful degradation

### DIFF
--- a/src/dbt_mcp/config/credentials.py
+++ b/src/dbt_mcp/config/credentials.py
@@ -177,6 +177,28 @@ class CredentialsProvider:
         self.settings = settings
         self.token_provider: TokenProvider | None = None
         self.authentication_method: AuthenticationMethod | None = None
+        self.account_identifier: str | None = None
+
+    async def _resolve_account_identifier(self) -> None:
+        """Fetch and store the account identifier from the Admin API.
+
+        Fails silently — account_identifier remains None on error.
+        """
+        if not self.settings.dbt_account_id or not self.settings.actual_host:
+            return
+        try:
+            # Imported here to avoid circular import:
+            # credentials → config_providers.admin_api → credentials
+            from dbt_mcp.config.config_providers.admin_api import (
+                DefaultAdminApiConfigProvider,
+            )
+            from dbt_mcp.dbt_admin.client import DbtAdminAPIClient
+
+            admin_client = DbtAdminAPIClient(DefaultAdminApiConfigProvider(self))
+            account_data = await admin_client.get_account(self.settings.dbt_account_id)
+            self.account_identifier = account_data.get("identifier")
+        except Exception as e:
+            logger.warning(f"Failed to fetch account identifier: {e}")
 
     def _log_settings(self) -> None:
         settings = self.settings.model_dump()
@@ -246,6 +268,8 @@ class CredentialsProvider:
             )
             self.token_provider = token_provider
 
+            await self._resolve_account_identifier()
+
             # Only validate CLI settings here — platform settings were already
             # checked at the top of get_credentials() and the OAuth flow has
             # populated the remaining fields (host, env ids, account id).
@@ -280,6 +304,7 @@ class CredentialsProvider:
                 self.settings.host_prefix = fetched_prefix
                 self.settings.dbt_host = self.settings.base_host
                 logger.info(f"Fetched prefix {fetched_prefix} from dbt Platform.")
+        await self._resolve_account_identifier()
         validate_settings(self.settings)
         self.authentication_method = AuthenticationMethod.ENV_VAR
         self._log_settings()
--- a/src/dbt_mcp/config/credentials.py
+++ b/src/dbt_mcp/config/credentials.py
@@ -6,8 +6,10 @@ from pathlib import Path
 
 from filelock import FileLock
 
+from dbt_mcp.config.config_providers.admin_api import DefaultAdminApiConfigProvider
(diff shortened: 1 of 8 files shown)


src/dbt_mcp/config/credentials.py
```diff
@@ -153,6 +153,24 @@ def __init__(self, settings: "DbtMcpSettings") -> None:
...
+        """Fetch account identifier from the Admin API. Fails silently."""
+        if not self.settings.dbt_account_id or not self.settings.actual_host:
+            return
+        try:
+            # Imported here to avoid circular import:
+            # credentials → config_providers.admin_api → credentials
+            from dbt_mcp.config.config_providers.admin_api import (
+                DefaultAdminApiConfigProvider,
+            )
+            from dbt_mcp.dbt_admin.client import DbtAdminAPIClient
```
DevonFulcher: It is unlikely that there is actually a circular import here. AI often makes this mistake. Can we move these imports to the top?

src/dbt_mcp/config/credentials.py
```diff
@@ -221,6 +239,15 @@ async def get_credentials(self) -> "tuple[DbtMcpSettings, TokenProvider]":
             )
             self.token_provider = token_provider
 
+            # Try to resolve account identifier from JWT claims
+            self.settings.account_identifier = (
+                dbt_platform_context.decoded_access_token.decoded_claims.get(
+                    "https://dbt.com/account_identifier"
+                )
+            )
+            if not self.settings.account_identifier:
```
DevonFulcher: This part may also be a hallucination. Let's just rely on `_fetch_account_identifier` instead.

src/dbt_mcp/config/settings.py
```diff
@@ -55,6 +55,7 @@ class DbtMcpSettings(BaseSettings):
         None, alias="MULTICELL_ACCOUNT_PREFIX"
     )  # legacy support
     host_prefix: str | None = Field(None, alias="DBT_HOST_PREFIX")
+    account_identifier: str | None = None
```
DevonFulcher: This seems a bit out of place here because it is the only field here that doesn't have an alias and presumably can't be set with an environment variable. If you can come up with a way to decouple this field from this class, that would be preferable.

DevonFulcher (commented):
Thanks for contributing this! I had a few small comments.

src/dbt_mcp/config/credentials.py
```diff
@@ -153,6 +153,24 @@ def __init__(self, settings: "DbtMcpSettings") -> None:
         self.token_provider: TokenProvider | None = None
         self.authentication_method: AuthenticationMethod | None = None
 
+    async def _fetch_account_identifier(self) -> None:
+        """Fetch account identifier from the Admin API. Fails silently."""
```
jairus-m: Just a small thing i noticed! When i saw this fetch method return None, was curious at what it was doing. 

Given that this helper "gets" the account identifier but really sets/mutates `DbtMcpSettings.account_identifier` should the docstring and function be named be more aligned to that?

src/dbt_mcp/config/credentials.py
eknoll: I believe this was an actual circular import, but worked around it in https://github.com/dbt-labs/dbt-mcp/pull/701/commits/bf87eae97dc70a3cf4b23fd558fb63e6cd58a854

src/dbt_mcp/config/credentials.py
eknoll: Thanks, agree and addressed in https://github.com/dbt-labs/dbt-mcp/pull/701/commits/3c1154a4c8dae02fd22bbe582001928e9022d30e

DevonFulcher (approved):
Looks good thanks!

jairus-m:
Dependabot issue should be resolved if you merge latest changes into main. Thanks!

PR #700, opened 2026-04-07
Remove mcp sdk version pin

DevonFulcher:
## Summary

Since we are now depending on a fork of the MCP SDK, we no longer need to pin a specific version. This pin was causing uv dependency resolution issues when importing dbt MCP into our internal server.

## Checklist
- [x] I have performed a self-review of my code
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #702, opened 2026-04-08
Remove forked MCP SDK source

DevonFulcher:
## Summary

Upon further investigation, this fork isn't being used in the wheel we are distributing to pypi. For instance, this command shows that we still depend on the official SDK:
```
uvx --with dbt-mcp python -c "from importlib.metadata import version; print(version('mcp'), __import__('mcp').__file__)"
```
So, it is best to remove this dependency. We plan on continuing to depend on the fork internally for remote MCP due to the memory leak issues it fixes.

This supersedes this other PR: https://github.com/dbt-labs/dbt-mcp/pull/700

## Checklist


wiggzz (approved):
looks good. we can pin in ai-codegen-api anyway. so we never needed to override here.

PR #715, opened 2026-04-12
fix: resolve symlinks before creating mcp.lock FileLock

### LINKED ISSUE #533: [Bug]: `.dbt/mcp.lock` cannot be a symbolic link of depth 2+
### What happened?

I like to source control my dotfiles (i.e. [dataders/dotfiles](https://github.com/dataders/dotfiles)). however to not share sensitive information, sensitive files in that repo are also symlinked to another directory.

With this setup, I'm getting this error message.

```
OSError: [Errno 62] Too many levels of symbolic links: '/Users/dataders/.dbt/mcp.lock'
```

AFAICT, this file is just an empty text file.

Also this problem does not happen with the `mcp.yml` which actually has information in it

### Steps to Reproduce

1. set up dbt-mcp
2. move `.dbt/mcp.lock` to `dotfiles_private/`
3. symlink it to `dotfiles_public/`
    ```sh
    ln -sf ~/Developer/dotfiles_private/.dbt/mcp.lock  ~/Developer/dotfiles/.dbt/mcp.lock
    ```
4. symlink to `~/`
    ```sh
    ln -sf ~Developer/dotfiles/.dbt/mcp.lock  ~/.dbt/mcp.lock  
    ```
5. start mcp server `dbt-mcp`

### Deployment

Remote MCP server

### Environment

- mac os: `15.6.1 (24G90)`
- python `3.13`
- dbt-mcp: `v1.7.0`

### Relevant log output

```shell

```

psaikaushik:
## Summary

When `~/.dbt/mcp.lock` is a multi-level symbolic link (symlink → symlink → actual file), Python's `filelock.FileLock` raises `OSError: [Errno 62] Too many levels of symbolic links`. This PR fixes it by calling `Path.resolve()` on the lock file path before passing it to `FileLock`.

Closes #533

### `src/dbt_mcp/config/credentials.py`
- In `get_dbt_platform_context()`, the lock file path is now resolved via `(dbt_user_dir / "mcp.lock").resolve()` before being passed to `FileLock`
- This follows all symlink levels to the actual target file, regardless of symlink depth
- Added a code comment referencing the issue for context

### `tests/unit/config/test_mcp_lock_symlink.py`
- Test that verifies multi-level symlink chains resolve correctly
- Test that creates a symlink chain, passes the symlink as `dbt_user_dir`, and asserts `FileLock` receives the exact resolved real path

## Why

Users who source-control their dotfiles often use symlink chains:
```
~/.dbt/mcp.lock → ~/dotfiles/.dbt/mcp.lock → ~/dotfiles_private/.dbt/mcp.lock
```

The existing code passes the unresolved path directly to `FileLock`, which fails when the symlink depth exceeds the OS limit. The `mcp.yml` file doesn't have this problem because it's read differently.

## Test plan

- [x] Unit tests verify symlink resolution behavior
- [x] No changes to existing test behavior
- [x] Manual verification with multi-level symlink setup

## Checklist
- [x] I have added tests that prove my fix is effective
- [x] Changelog entry added


.changes/unreleased/Fixed-20260412-533.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: Resolve symbolic links for mcp.lock file path before creating FileLock, preventing OSError when ~/.dbt/mcp.lock is a multi-level symbolic link.
+time: 2026-04-12T21:00:00.000000-07:00
```
jairus-m: `kind: Fixed` is not one of the options for changie and will probably break in the release process. Can you use `changie new` locally and selecting "Bug Fix" in the options menu. You should not have to edit the changie files themselves. Thanks!

psaikaushik: Good catch, thanks! Fixed — deleted the `Fixed-20260412-533.yaml` file and replaced it with `Bug Fix-20260412-210000.yaml` using `kind: Bug Fix` as defined in `.changie.yaml`.

psaikaushik:
@jairus-m , @jasnonaz , @b-per  : Please take a look. Addessed the comments. thanks!!

tests/unit/config/test_mcp_lock_symlink.py
```diff
@@ -0,0 +1,81 @@
...
+        mock_context_manager.read_context.return_value = mock_ctx
+
+        with patch("dbt_mcp.config.credentials.FileLock", MockFileLock):
+            from dbt_mcp.config.credentials import get_dbt_platform_context
+
+            await get_dbt_platform_context(
+                dbt_user_dir=tmp_path,
+                dbt_platform_url="https://cloud.getdbt.com",
+                dbt_platform_context_manager=mock_context_manager,
+            )
```
jairus-m, line 89: This is a real directory path that pytest passes? Should we be passing in a symlink to test explicitly?

tests/unit/config/test_mcp_lock_symlink.py
```diff
@@ -0,0 +1,81 @@
...
+
+        assert len(captured_lock_paths) == 1
+        # The path passed to FileLock should be fully resolved (no symlinks)
+        assert not captured_lock_paths[0].is_symlink() or not os.path.islink(
+            str(captured_lock_paths[0])
+        )
```
jairus-m: This current assertion only tells you the result wasn't a symlink. It may be better to explicitly assert that the result is what's actually expected(true path and not a symlink).

```python
assert captured_lock_paths[0] == (real_dbt_dir / "mcp.lock").resolve()
```

^^ the above assertion would go hand-in-hand with creating a symlink based off some `real_dbt_dir` from the earlier comment

tests/unit/config/test_mcp_lock_symlink.py
psaikaushik: You're absolutely right,  passing a real directory completely bypasses the symlink resolution path we're trying to test. Fixed: the test now creates `link2 -> link1 -> real_dbt_dir` and passes `link2` (the symlink) as `dbt_user_dir`, so `get_dbt_platform_context` actually has to resolve through the symlink chain. Sorry about that.

tests/unit/config/test_mcp_lock_symlink.py
psaikaushik: Makes sense. Updated to:

```python
assert captured_lock_paths[0] == (real_dbt_dir / "mcp.lock").resolve()
```

This asserts the exact resolved path matches the real directory, not just "it's not a symlink". thanks for the feedback!

psaikaushik:
@jairus-m : Thanks for the feedback. Addressed the comments. Please take a look. thanks!

jairus-m:
Thanks @psaikaushik ! Updates look great - will approve.

Last couple of things - just realized your commits arent signed. Can you sign them? More info in the [CONTRIBUTING.md](https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#signed-commits)!

In your test plan, I see your last step to manually verify your changes is still unchecked, any chance you were able to test this locally?

psaikaushik:
> Thanks @psaikaushik ! Updates look great - will approve.
> 
> Last couple of things - just realized your commits arent signed. Can you sign them? More info in the [CONTRIBUTING.md](https://github.com/dbt-labs/dbt-mcp/blob/main/CONTRIBUTING.md#signed-commits)!
> 
> In your test plan, I see your last step to manually verify your changes is still unchecked, any chance you were able to test this locally?

@jairus-m  : Didn't know about the signing. Just signed all the commits. Thanks for pointing it out. Regarding the test plan, yes, I tested it locally by creating a multi level symlink chain, run the server briefly to confirm it doesn't crash. It was successful. I forgot to check the box later. thanks for that. I fixed it.

jairus-m (approved):
Thank you! 🚀

psaikaushik:
@jairus-m  : I updated the branch. I am not entirely sure what else is required to merge this. Please let me know. Thanks!

PR #723, opened 2026-04-16
Add DBT_HOST_PREFIX env var to remote session initialization

jairus-m:
## Summary

Small change to add the other (primary) environment variable (`DBT_HOST_PREFIX`) for adding the account prefix to the host url for remote session initialization.

~~Also reuses `_build_dbt_platform_url` since it has some existing protections from user misconfigs.~~

^^ reverted the above:
> _build_dbt_platform_url has a heuristic that treats any 4-label host's first label as a potential account prefix. The CI environment uses DBT_HOST w/ 4 labels (us is a regional subdomainwith) while MULTICELL_ACCOUNT_PREFIX differs. This triggers a false-positive mismatch error. The function was designed to receive pre-validated values from DbtMcpSettings, not raw env var strings. session.py does not have the validation, so the assumed heuristic of  _build_dbt_platform_url does not hold.

## Related Issues

Closes #
Related to #

## Checklist

### DIFF
--- a/.changes/unreleased/Under the Hood-20260415-145943.yaml
+++ b/.changes/unreleased/Under the Hood-20260415-145943.yaml
@@ -1,0 +1,3 @@
+kind: Under the Hood
+body: Add DBT_HOST_PREFIX support to remote session initialization
+time: 2026-04-15T14:59:43.695388-07:00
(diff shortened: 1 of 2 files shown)


DevonFulcher (approved):
🚀

PR #728, opened 2026-04-18
feat: add MCP elicitation for interactive DBT_HOST configuration 

### LINKED ISSUE #488: [Feature]: Configuration with Elicitation

With the latest oauth enhancements, the configuration burden on users is greatly reduced. However, there are still a few required configurations that users need to modify in order to use this server. This means that users need to edit a JSON file to get started. For some business users, that can be too complicated. Currently, users need to configure `DBT_HOST`, both `DBT_PROJECT_DIR` & `DBT_PATH`, or all three.

When not set through environment variables, these options can be set through MCP elicitation when the server starts up and before the oauth flow is triggered. This allows the user to set these variables without needing to edit a JSON file.

MCP has a new elicitation spec for [URL-based elicitation](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887). However, this is quite new and likely not widely supported. Since we are only configuring three variables, we can likely just utilize the pre-existing elicitation method.

### Additional context

Elicitation isn't widely supported. We need to ensure we don't require the client to provide elicitation to configure the server. One solution is to check the [client's capabilities](https://modelcontextprotocol.io/specification/2025-06-18/architecture/index#capability-negotiation) before requesting elicitation. I'm not sure how to do this with the Python SDK.

Muizzkolapo:
## Summary

Add MCP elicitation support so users can configure `DBT_HOST` interactively when it's not set, instead of getting an error.

## What Changed

- New `elicitation.py` with reusable elicitation infrastructure (`get_elicitation_session`, `elicit_or_raise`) that any future elicitation can build on
- First consumer: `ElicitingCredentialsProvider` wraps `CredentialsProvider` to prompt for `DBT_HOST` when missing, then feeds it into the existing OAuth flow
- Elicited host persists to `~/.dbt/mcp.yml` via `DbtPlatformContext.dbt_host` — reuses the existing context file and `override()` merge logic (no separate config file)
- `config.py`/`settings.py`/`server.py`: wire the wrapper, stop auto-disabling platform features, remove conditional guards on tool registration

## Why

Users currently need to edit JSON config files to set `DBT_HOST` before the server works. For non-technical users, this is a barrier. With elicitation, the server prompts for the host interactively on first use, persists it, and never asks again.

Clients that don't support elicitation (e.g., Claude Desktop) get the original error message — no regression.

## Related Issues

Closes #488

###  MCP INSPECTOR:
image
image
image

### CC
image

image
image

## Checklist
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required ->  https://github.com/dbt-labs/docs.getdbt.com/pull/9001

## Additional Notes

- Tested with MCP Inspector (elicitation form works), Claude Code (elicitation works inline), and Claude Desktop (graceful degradation — `elicitation=None` in capabilities)
- Only `DBT_HOST` is elicited; `DBT_PROJECT_DIR` and `DBT_PATH` remain env-var-only for CLI users
- Elicited host persists to `~/.dbt/mcp.yml` as `dbt_host` field on `DbtPlatformContext`; env vars always take precedence
- `host_prefix` is also restored from persisted context on restart for multi-cell account URL construction


src/dbt_mcp/config/elicitation.py
```diff
@@ -0,0 +1,178 @@
...
+            except MissingHostError as e:
+                data = await elicit_or_raise(
+                    e,
+                    DbtHostSchema,
+                    "Let's set up dbt-mcp. What's your dbt Cloud host?",
+                )
+                self._inner.settings.dbt_host = data.dbt_host
+                result = await self._inner.get_credentials()
+                self._persistence.write("dbt_host", data.dbt_host)
+                return result
```
jairus-m: In this async function, does this `_persistence.write` method need to `await` the result? Otherwise that line will hold up the whole async function.

src/dbt_mcp/config/elicitation.py
```diff
@@ -0,0 +1,178 @@
...
+    """Read/write elicited config values to ~/.dbt/mcp-config.yml."""
+
+    def __init__(self, config_path: Path | None = None) -> None:
+        if config_path is None:
+            from dbt_mcp.config.credentials import get_dbt_profiles_path
+
+            config_path = get_dbt_profiles_path() / "mcp-config.yml"
+        self._path = config_path
+        self._lock_path = config_path.with_suffix(".lock")
+
```
jairus-m: There was a recent update (https://github.com/dbt-labs/dbt-mcp/pull/715) that had to do with issues arising with symlinks and lock files. 

I'm not 100% sure if this applies in this scenario but wonder if adding a resolve() method can be a safeguard from future bugs

```python
self._lock_path = config_path.with_suffix(".lock").resolve()
```

src/dbt_mcp/config/elicitation.py
```diff
@@ -0,0 +1,178 @@
...
+
+class ConfigPersistence:
+    """Read/write elicited config values to ~/.dbt/mcp-config.yml."""
+
+    def __init__(self, config_path: Path | None = None) -> None:
+        if config_path is None:
+            from dbt_mcp.config.credentials import get_dbt_profiles_path
+
+            config_path = get_dbt_profiles_path() / "mcp-config.yml"
```
jairus-m: So there already exists a `~/.dbt/mcp.yml` file and an associated lock file - is there a path to merging the write of this elicitation to that existing file? So we don't have to persist two similar config/lock files?

Also any extra context too for the approach here will be of great help!

src/dbt_mcp/config/elicitation.py
Muizzkolapo: Wrapped it in `asyncio.to_thread()` so it doesn't block the event loop.

src/dbt_mcp/config/elicitation.py
Muizzkolapo: done

src/dbt_mcp/config/elicitation.py
Muizzkolapo: having a look

Muizzkolapo: image

Muizzkolapo:
Updated tests post pr review.
image

image
image

image
image

src/dbt_mcp/config/credentials.py
```diff
@@ -226,13 +229,20 @@ async def get_credentials(self) -> "tuple[DbtMcpSettings, TokenProvider]":
...
             )
             config_location = dbt_user_dir / "mcp.yml"
+            dbt_platform_context_manager = DbtPlatformContextManager(config_location)
+            existing_context = dbt_platform_context_manager.read_context()
+            if (
+                existing_context
+                and existing_context.dbt_host
+                and not self.settings.actual_host
+            ):
+                self.settings.dbt_host = existing_context.dbt_host
```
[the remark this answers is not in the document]
Muizzkolapo: Fixed now restoring both dbt_host and host_prefix from persisted context when env vars are not set.

src/dbt_mcp/config/credentials.py
```diff
@@ -226,13 +229,20 @@ async def get_credentials(self) -> "tuple[DbtMcpSettings, TokenProvider]":
                 dbt_profiles_dir=self.settings.dbt_profiles_dir
             )
             config_location = dbt_user_dir / "mcp.yml"
+            dbt_platform_context_manager = DbtPlatformContextManager(config_location)
+            existing_context = dbt_platform_context_manager.read_context()
```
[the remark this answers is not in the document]
Muizzkolapo: Updated

src/dbt_mcp/mcp/server.py
```diff
@@ -125,19 +142,25 @@ async def app_lifespan(server: FastMCP[Any]) -> AsyncIterator[bool | None]:
...
+                        else None
+                    ),
+                    enabled_toolsets=server.config.enabled_toolsets,
+                    disabled_toolsets=server.config.disabled_toolsets,
+                )
+            except MissingHostError as e:
+                logger.warning(
+                    "Could not register proxied tools — credentials not yet available: %s",
+                    e,
+                )
```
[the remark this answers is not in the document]
Muizzkolapo: This was by design.

src/dbt_mcp/config/elicitation.py
```diff
@@ -0,0 +1,125 @@
+"""MCP elicitation primitives."""
+
+from __future__ import annotations
+
+import asyncio
```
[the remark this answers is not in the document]
Muizzkolapo: Updated pr.

src/dbt_mcp/mcp/server.py
```diff
@@ -56,7 +57,17 @@ def __init__(
...
+            ) = await self.config.eliciting_credentials_provider.get_credentials()
+        except MissingHostError as e:
+            logger.warning(
+                "Could not resolve credentials — defaulting to single-project mode: %s",
+                e,
+            )
+            return False
         return bool(
             settings.dbt_project_ids is not None and len(settings.dbt_project_ids) > 0
         )
```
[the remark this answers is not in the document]
Muizzkolapo: _is_multi_project() calls get_credentials(), which returns cached credentials on subsequent calls. Elicitation only triggers on the first call when no DBT_HOST is set and OAuth is needed. The CLI-only tools don't hit this path.

src/dbt_mcp/errors/common.py
```diff
@@ -22,3 +22,9 @@ class NotFoundError(ToolCallError):
     """Exception raised when a resource is not found."""
 
     pass
+
+
+class MissingHostError(ValueError):
```
jairus-m:
```suggestion
class MissingHostError(ConfigurationError):
```
not a big deal - but ConfigurationError may be more accurate 👍

src/dbt_mcp/config/config.py
```diff
@@ -87,14 +88,15 @@ class Config:
...
     multi_project_semantic_layer_config_provider: (
-        MultiProjectSemanticLayerConfigProvider | None
+        MultiProjectSemanticLayerConfigProvider
     )
-    semantic_layer_config_provider: DefaultSemanticLayerConfigProvider | None
-    admin_api_config_provider: DefaultAdminApiConfigProvider | None
+    semantic_layer_config_provider: DefaultSemanticLayerConfigProvider
+    admin_api_config_provider: DefaultAdminApiConfigProvider
     credentials_provider: CredentialsProvider
+    eliciting_credentials_provider: ElicitingCredentialsProvider
```
jairus-m: Now that we have both of these credential providers in `Config`, will there ever be confusion on using the correct provider? Is it correct that we should always be using `eliciting_credentials_provider` from now on? Can the naming be clarified to reflect this somehow? Not sure about this best way about this.

Muizzkolapo: My approach would be to demote the inner credentials_provider to a local variable in load_config() (it's only used there to construct the other providers) and rename eliciting_credentials_provider →   credentials_provider on Config. I am happy to raise this as a follow up PR or we can add it here if agreed.

jairus-m: > Keep the original raw/inner `CredentialProvider` in `load_config()` as it's the only place it's used to instantiate the different tool set config providers and then return only a single `ElicitingCredentialProvider` in the `Config` class.

I do think this makes sense! And I also personally think the changes can also go into this PR. I asked @DevonFulcher to weigh in here, but this PR looks great! Thanks Muizz 🔥

Muizzkolapo: No worries man, I've added it to this branch now.

jairus-m (changes requested):
Overall, looking great! Been doing some testing now and came across an issue. So with setting `DBT_PROJECT_DIR` and `DBT_PATH`, elicitation does correctly pop up for `DBT_HOST`. All good there!

However, if I put in an erroneous host value, the elicitation accepts it and then tries to continue on with the tool execution but ends up indefinitely hanging. I think some type of validation <> re-elicit loop/check would be useful! And/or a fallback to trigger a timeout if for some reasons, things do end up hanging. 

Another UX thing, is when elicitation _is_ triggered, would it be helpful to suggest to user to also consider setting a `DBT_HOST` in their env vars in the text? I ask this as elicitation is a fallback from user misconfig, yeah? So the context that it would come up assumes that env vars are configurable. Thoughts?

jairus-m:
@Muizzkolapo I re-tested after your latest commits. Thanks for addressing! There are a few things that I think still need to be improved with the UX wrt to invalid inputs:
1. No re-elicitation loop / no pre-validation
2. Invalid host persisted in `mcp.yml`
3. Would like a user hint on invalid host

I made some changes and this is what the UX looks like:

### Input invalid host: `invalid-host.us1.dbt.com`:
Screenshot 2026-04-24 at 3 39 03 PM

### Validation fails, re-elicits, and gives the user a hint:
Screenshot 2026-04-24 at 3 39 17 PM

- In the above, do note that this is configured to re-elicit at maximum 3 tries. In addition, after every failure, the `mcp.yml` dbt_host entry is cleared of invalid hosts to prevent bugs when re-trying in the future.

### After 3 tries, fails and outputs config help
Screenshot 2026-04-24 at 3 40 52 PM

- Do note, that even after failing 3 times, i asked to try again, was able to re-elciit succesfully (due to cleaning of `mcp.yml` at every failure), then put in the correct `DBT_HOST` which then succeeded.


The previous state didn't re-elicit on invalid host URLs (so no hints either) or clean the `mcp.yml`. Hope this makes sense and is a valuable addition! I'm going to push my commit but feel free to test it out, provide feedback, and make adjustments (can totally revert this as well).

Muizzkolapo:
@jairus-m just finished testing the changes on my side as well. Looks good to me and Thanks for the UX improvements!

jairus-m (approved):
Awesome!! Great work @Muizzkolapo

jairus-m:
@Muizzkolapo I see that PR is blocked due to early commits not being signed. Given previous contributions/collaboration, verification/signage of later commits (majority of them), and our use of sqaush commits, will merge! But pls remember to get those signed from the start! Appreciate you bro 🙏

PR #738, opened 2026-04-22
feat: populate dbt_cloud_user_id in telemetry events

DevonFulcher:
# Why

The `dbt_cloud_user_id` field in telemetry events was only populated when the `DBT_USER_ID` environment variable was explicitly set (PAT/env_var auth path). In the OAuth path it worked correctly via the JWT `sub` claim, but PAT users without `DBT_USER_ID` sent empty user IDs in all events. This PR follows the same pattern as #701 (which resolved `account_identifier`).

## Checklist
- [x] Self-review performed
- [x] No documentation changes needed (internal telemetry change)
- [x] Tests added and passing (551 unit tests pass)
- [x] \`pre-commit run --all-files\`: ruff check, ruff format, mypy all clean


src/dbt_mcp/dbt_admin/client.py
```diff
@@ -70,6 +70,11 @@ async def get_account(self, account_id: int) -> dict[str, Any]:
         )
         return result.get("data", {})
 
+    async def get_current_user(self) -> dict[str, Any]:
+        """Get details for the current authenticated user."""
+        result = await self._make_request("GET", "/api/v2/users/profile/")
```
b-per: This endpoint doesn't work when I tested it. Neither with a PAT or a Service Token.

`/api/v2/whoami/` works for PATs and returns user data ; but it doesn't work for Service Tokens

tests/unit/dbt_admin/test_client.py
```diff
@@ -696,6 +696,26 @@ async def test_get_account(client):
     )
 
 
+async def test_get_current_user(client):
+    mock_response = MagicMock()
+    mock_response.json.return_value = {"data": {"id": 789, "email": "user@example.com"}}
```
b-per: See other comment. I think that this mock is incorrect.

Should we have a mechanism to make sure that all our mocks are actually OK? Maybe a way to create them from "real" API calls? The issue here is that the test is passing but it is because of the incorrect mock.

b-per (changes requested):
From my tests, the endpoint doesn't work

b-per:
After checking a bit more, there might not be any other option than calling `/api/private/identify/` to get the token id

src/dbt_mcp/dbt_admin/client.py
DevonFulcher: Oh dang, sorry! AI hallucinated that and I should have validated this more.

tests/unit/dbt_admin/test_client.py
DevonFulcher: I'll fix this mock! I think the best way to test real API calls is to use our integration test suite. I'll add a test to this PR for this API call.

src/dbt_mcp/config/credentials.py
```diff
@@ -203,6 +205,25 @@ async def _resolve_account_identifier(self) -> None:
...
+            self.user_id = self.settings.dbt_user_id
+            return
+        if not self.settings.actual_host:
+            return
+        try:
+            admin_client = DbtAdminAPIClient(DefaultAdminApiConfigProvider(self))
+            user_data = await admin_client.get_current_user()
+            self.user_id = user_data.get("id")
+        except Exception as e:
+            logger.warning(f"Failed to fetch user id: {e}")
```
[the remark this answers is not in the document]
DevonFulcher: Addressed in fcebd605: added an early return in `_resolve_user_id` that attempts `token_provider.get_token()` and skips the Admin API call when no token is available, so we no longer log a warning on every startup in this configuration.

(reply from AI)

tests/integration/dbt_admin/test_admin_client.py
```diff
@@ -0,0 +1,24 @@
...
+from dbt_mcp.config.credentials import CredentialsProvider
+from dbt_mcp.config.settings import DbtMcpSettings
+from dbt_mcp.dbt_admin.client import DbtAdminAPIClient
+
+
+@pytest.fixture
+def admin_client() -> DbtAdminAPIClient:
+    settings = DbtMcpSettings()  # type: ignore
+    credentials_provider = CredentialsProvider(settings)
+    return DbtAdminAPIClient(DefaultAdminApiConfigProvider(credentials_provider))
```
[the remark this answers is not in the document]
DevonFulcher: Addressed in fcebd605: the fixture now skips when `DBT_HOST`, `DBT_TOKEN`, or `DBT_ACCOUNT_ID` is missing, matching the pattern in `tests/integration/discovery/conftest.py`.

(reply from AI)

b-per (approved):

PR #741, opened 2026-04-26
feat: register admin API tools in multiproject MCP server

DevonFulcher:
## Summary

- Registers admin API tools in the multiproject MCP server so they are available when `DBT_PROJECT_IDS` is configured
- Admin tools are account-scoped (not project-scoped), so they work in both modes without requiring multiproject variants
- Updates `list_jobs` tool description to clarify that results are auto-scoped to the prod environment when configured, and account-wide otherwise
- Adds a test asserting admin tools appear in the multiproject MCP when `admin_api_config_provider` is set

## Behavior in multiproject mode

All 10 admin tools (`list_projects`, `list_jobs`, `trigger_job_run`, etc.) are now available when `DBT_PROJECT_IDS` is set. The `list_jobs` tool returns account-wide results in this mode since no `prod_environment_id` is configured — filtering by `project_id` is noted as future work in the existing TODO comment.

## Test plan

- [x] Unit tests pass: `task test:unit`
- [x] Checks pass: `task check`
- [x] Verify admin tools appear in tool list when running with `DBT_PROJECT_IDS` set


src/dbt_mcp/prompts/admin_api/list_jobs.md
```diff
@@ -4,14 +4,14 @@ This tool retrieves jobs from the dbt Admin API. Jobs are the configuration for
 
 ## Parameters
 
-- **project_id** (optional): Filter jobs by specific project ID
-- **environment_id** (optional): Filter jobs by specific environment ID
 - **limit** (optional): Maximum number of results to return
 - **offset** (optional): Number of results to skip for pagination
 
+When a single production environment is configured, results are automatically scoped to that environment. Otherwise, all jobs in the account are returned — use `limit` to constrain large result sets.
+
```
jairus-m, line 11: > When a single production environment is configured

Does `list_jobs` return production jobs only? So even for multiproject, only returns prod jobs?

Or is that supposed to reference:

> When a single ~~production environment~~ project is configured

Meaning that `list_jobs` returns all jobs, regardless of the environment?

EDIT: Confusion point was since `project_id` and `environment_id` are removed and line 10 now calls out prod env configs only, was not sure if I understood the baseline behavior of `list_jobs` correctly

DevonFulcher: Ya, I think `list_jobs` will return non-production jobs. I couldn't verify that, but there doesn't seem to be any filtering on the endpoint or tool for prod-only jobs.

This statement refers more to how the user configures dbt MCP than how the agent should interact with it, so it might not be the best phrasing. I think it is just trying to convey to the agent: "the returned jobs may belong to multiple environments". I think your suggestion makes sense, though. I'll update.

I don't think the removal of `project_id and `environment_id` is related. Removing those is just about correctness. They weren't parameters to the tool before this PR either.

jairus-m (approved):
Thank you - your update and response makes sense!

PR #750, opened 2026-05-04
fix: don't merge stderr into stdout in subprocess calls

### LINKED ISSUE #749: [Bug]: `--quiet` flag returns stderr noise instead of "OK" on successful build
### What happened?

When using `externalbrowser` auth with Snowflake, the `build`, `run`, and `test` tools return a urllib3 warning string instead of `"OK"` on success.

The root cause is in `_run_dbt_command` — `stderr=subprocess.STDOUT` merges Python warnings (urllib3, deprecation notices, etc.) into stdout. With `--quiet`, dbt produces no stdout on success, so the `return output or "OK"` fallback should return `"OK"`. But the leaked stderr makes `output` truthy, so the tool returns a useless warning string instead.

On failure, the behavior is correct — errors survive `--quiet` and are returned as expected.

### Steps to Reproduce

1. Configure dbt with `authenticator: externalbrowser` (Snowflake SSO)
2. Run any build/run/test via the MCP tool (e.g., `build` with `node_selection: "my_model"`)
3. All models and tests pass
4. Instead of `"OK"`, the tool returns something like:
   /path/to/urllib3/init.py:35: NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+...

### Deployment

Local MCP server

### Environment

- dbt-mcp v1.16.0
- macOS 
- Claude Code (MCP client)

### Relevant log output

```shell

```

b-per:
## Summary

Closes #749

- Changed `stderr=subprocess.STDOUT` to `stderr=subprocess.PIPE` in `_run_dbt_command` (dbt CLI tools) and `_run_codegen_operation` (dbt codegen tools)
- This prevents Python-level warnings (e.g. `urllib3` `NotOpenSSLWarning` from the Snowflake `externalbrowser` authenticator) from being merged into stdout
- With `--quiet`, dbt produces no stdout on a successful run, so `output or "OK"` now correctly returns `"OK"` instead of the leaked warning string

## Test plan

- [ ] Unit tests added for both `dbt_cli` and `dbt_codegen`: simulate empty stdout + stderr warning and assert the tool returns `"OK"` and that `Popen` is called with `stderr=subprocess.PIPE`
- [ ] Run `task test:unit` — all 601 tests pass
- [ ] Run `task check` — linting and type checking pass

### DIFF
--- a/.changes/unreleased/Bug Fix-20260504-111117.yaml
+++ b/.changes/unreleased/Bug Fix-20260504-111117.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Fix build/run/test tools returning urllib3 warnings instead of 'OK' on success when using externalbrowser auth
+time: 2026-05-04T11:11:17.317214+02:00
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -104,7 +104,7 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
                 args=args,
                 cwd=cwd_path,
                 stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
+                stderr=subprocess.PIPE,
                 stdin=subprocess.DEVNULL,
                 text=True,
             )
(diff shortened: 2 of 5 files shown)


DevonFulcher (commented):
I'm unsure about this. Adding stderr to the tool output was an intentional decision. I think there may be useful information that is sent to stderr.

b-per:
Alt approach in #772

PR #753, opened 2026-05-05
fix: prevent CLI-only server crash introduced in v1.17.0

b-per:
## Summary

- **Root cause:** In v1.17.0, `proxied_tool_config_provider` became unconditionally non-`None`. `register_proxied_tools` called `get_config()` _before_ checking whether any proxied tools were enabled. For CLI-only users (`DBT_HOST` unset, all platform toolsets explicitly disabled), `get_config()` successfully retrieved credentials via the static-token path, then hit a bare `assert settings.actual_host` — raising `AssertionError("")` (empty message). That exception is not a `MissingHostError`, so the `except MissingHostError` guard in `app_lifespan` didn't catch it, and the server crashed with an uninformative `ERROR Error in MCP server:` log line.
- **Workaround (until this lands):** Add `DBT_HOST=localhost` (any non-empty string) to the env file. Since all platform features are already disabled, the value is never used.

## Changes

- **`proxy/tools.py`** — Move the `configured_proxied_tools` filter check _before_ `get_config()`. CLI-only users with all proxied toolsets disabled now return early without touching credentials at all.
- **`config_providers/proxied_tool.py`** — Replace `assert settings.actual_host` with an explicit `MissingHostError` raise so the exception type is consistent and is caught by the existing handler in `app_lifespan`.
- **`mcp/server.py`** — Switch `logger.error(...)` to `logger.exception(...)` in `app_lifespan` so any unexpected crash always includes the full traceback, not just the (possibly empty) exception message.

## Test plan

- [ ] `test_register_proxied_tools_skips_get_config_when_all_proxied_toolsets_disabled` — asserts `get_config()` is never called when all proxied toolsets are disabled
- [ ] `test_proxied_tool_config_raises_missing_host_error_not_assertion_error` — asserts `get_config()` raises `MissingHostError` (not `AssertionError`) when `actual_host` is `None`
- [ ] `test_lifespan_logs_exception_with_traceback` — asserts `app_lifespan` logs with `exc_info` set so tracebacks are always visible
- [ ] Full unit suite: `uv run pytest tests/ --ignore=tests/integration -x -q` — 604 passed


jairus-m (approved):
TDIL about logger.exception vs logger.error - thanks for this ! Root cause and changes make sense

DevonFulcher (approved):

PR #761, opened 2026-05-07
Add MCP server instructions

DevonFulcher:
## Summary

I've been testing with a new MCP client. It requires a lot of explicit instructions to use the server. I'm hoping these instructions will allow the agent to use dbt MCP more readily.

## Checklist

### DIFF
--- a/src/dbt_mcp/prompts/mcp/server_instructions.md
+++ b/src/dbt_mcp/prompts/mcp/server_instructions.md
@@ -3,7 +3,7 @@ Use these tools to interact with dbt resources (typically via dbt Platform):
 - Understand how the project is structured: what exists in the environment, how objects depend on one another, and where to look when something looks wrong or slow.
 - Reason over governed business meaning—named measures and breakdowns the project maintains—and answer questions with validated aggregates or the warehouse logic behind them when that helps.
 - Work with platform automation when available: see what runs on a schedule, inspect past outcomes, act on failures or reruns, and dig into logs and outputs.
-- Assist with engineering work on dbt projects: take action on the project, generate or modify project code, and reason about the underlying SQL.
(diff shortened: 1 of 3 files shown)


src/dbt_mcp/prompts/mcp/server_instructions.md
jairus-m, line 1: These instructions are specific to SL tool set usage? What about other uses like admin api, cli, sql, etc?

DevonFulcher: I tried to make the instructions as generally applicable as possible. I basically wanted this to read like generic analytics engineering tasks, not anything specific to a toolset. I did try to focus it a bit on the three main use cases: Discovery, SL, and admin while being generic enough to apply more broadly.

How would you change the wording?

jairus-m: > I tried to make the instructions as generally applicable as possible.

Re-reading with this in mind, I understand the motivation much better. It did feel skewed towards general SL/Discovery/Admin since there was no wording about building models, but that may be too specific tbh and something that would be better delegated to the dbt CLI itself in the future?

If I were to change any wording of this, would be to just add after line 6
```
- Execute and validate dbt project changes: build, compile, test, or clone models when investigating or making changes.
```

but now, this may be too specific (what you have already is great with me, was more curious about the skew)

DevonFulcher: Is it better now? I think the intent should be to just prime the agent on the server's overarching theme. The real details are in the tool descriptions.

jairus-m (approved):
> Re: intent should be to just prime the agent on the server's overarching theme + tool descripts

Heard! Thank you 🙌

PR #773, opened 2026-05-13
test(lsp): silence asyncio DeprecationWarning in sync tests

DevonFulcher:
## Summary
- Four tests in `TestMessageHandling` (`tests/unit/lsp/test_lsp_connection.py`) construct an `asyncio.Future()` synchronously without a current event loop, which emits `DeprecationWarning: There is no current event loop` under Python 3.12+.
- Adds a small `_sync_test_event_loop` context manager that binds a fresh loop for the duration of each affected test, and switches to `loop.create_future()`.
- No behavior change; warning is gone and the tests stay future-proof against Python removing implicit loop creation.

## Test plan
- [x] `task test:unit` — 692 passed
- [x] `task check` — lint, ruff, mypy clean
- [x] Verified the same five `TestMessageHandling` tests still pass with no warnings


jairus-m (approved):
Ran tests locally - no deprecations!

PR #774, opened 2026-05-13
refactor(tools): default structured_output to True, require title

DevonFulcher:
## Summary
- `GenericToolDefinition.structured_output` now defaults to `bool = True` (was `bool | None = False`), so new tools get structured output without explicit opt-in.
- `generic_dbt_mcp_tool` now requires `title` and accepts only keyword arguments, keeping call sites explicit.

## Test plan
- [x] `task test:unit` — 692 passed
- [x] `task check` — lint, ruff, mypy clean

### DIFF
--- a/src/dbt_mcp/tools/definitions.py
+++ b/src/dbt_mcp/tools/definitions.py
@@ -14,14 +14,12 @@ from dbt_mcp.tools.tool_names import ToolName
 @dataclass
 class GenericToolDefinition[NameEnum: Enum]:
     fn: Callable[..., Any]
+    title: str
     description: str
     name_enum: type[NameEnum]
     name: str | None = None
-    title: str | None = None
     annotations: ToolAnnotations | None = None
-    # We haven't strictly defined our tool contracts yet.
-    # So we're setting this to False by default for now.
-    structured_output: bool | None = False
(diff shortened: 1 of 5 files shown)


src/dbt_mcp/tools/definitions.py
```diff
@@ -14,14 +14,12 @@
 @dataclass
 class GenericToolDefinition[NameEnum: Enum]:
     fn: Callable[..., Any]
+    title: str
     description: str
     name_enum: type[NameEnum]
     name: str | None = None
```
jairus-m: Can we can document `title` vs `name` in `GenericToolDefinition` ? Like a comment or docstring?

Sorry, the difference may be obvious based on the patterns in the CLI tools where `name` is function identifer that an LLM calls via MCP (falls back to function name if `name` not set) and `title` is the human readable name - but just had a lapse of confusion when looking.

jairus-m (approved):
Looks great! With one small non-blocking comment.

src/dbt_mcp/tools/definitions.py
DevonFulcher: For sure. I'll add a comment.

PR #781, opened 2026-05-15
Fix unstructured tool errors being masked as 'outputSchema defined but no structured output returned'

jairus-m:
## Summary

When a tool raised an exception, `DbtMCP.call_tool()` caught it and returned `[TextContent(...)]` which is unstructured. After #774 changed `structured_output` to default to `True`, all tools gained an `outputSchema`. Because of this, the MCP server rejects the unstructured error response with "Output validation error: outputSchema defined but no structured output returned". The actual error is not returned and hidden to the user/LLM:

Screenshot 2026-05-15 at 8 46 38 AM

To address this, ~~we instead return `CallToolResult(isError=True)` on exception so the MCP server "short circuits"~~ raise an error which gets caught by FastMCP's handler _before_ the `outputSchema` check which then properly outputs the failure:

Screenshot 2026-05-15 at 8 46 16 AM

## Checklist

## Additional Notes

~~One thing Im unsure about about this implementation here is that we have to override the return type established by the parent `FastMCP.call_tool` method and add `CallToolResult` to `DbtMCP.call_tool`'s annotations. Because of this, mypy gets a bit mad and we have to tell it to "shhh". No behavioral changes with regards to the [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle), just type annotations! Hope this is fine.~~

^^ addressed in comments

### DIFF
--- a/src/dbt_mcp/mcp/server.py
+++ b/src/dbt_mcp/mcp/server.py
@@ -9,7 +9,7 @@ from typing import Any
 from dbtlabs_vortex.producer import shutdown
 from mcp.server.fastmcp import FastMCP
 from mcp.server.lowlevel.server import LifespanResultT
-from mcp.types import CallToolResult, ContentBlock, TextContent, Tool
+from mcp.types import ContentBlock, Tool
 
 from dbt_mcp.config.config import Config
 from dbt_mcp.dbt_admin.tools import register_admin_api_tools
@@ -79,11 +79,9 @@ class DbtMCP(FastMCP):
             settings.dbt_project_ids is not None and len(settings.dbt_project_ids) > 0
         )
 
-    async def call_tool(  # type: ignore[override]
+    async def call_tool(
         self, name: str, arguments: dict[str, Any]
-    ) -> Sequence[ContentBlock] | dict[str, Any] | CallToolResult:
-        # Return type includes CallToolResult on error to satisfy outputSchema check
-        # Parent class, FastMCP.call_tool does not include this type
+    ) -> Sequence[ContentBlock] | dict[str, Any]:
         logger.info(f"Calling tool: {name} with arguments: {_safe_args(arguments)}")
         result = None
         start_time = int(time.time() * 1000)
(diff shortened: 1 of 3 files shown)


src/dbt_mcp/mcp/server.py
```diff
@@ -79,9 +79,11 @@ async def _is_multi_project(self) -> bool:
             settings.dbt_project_ids is not None and len(settings.dbt_project_ids) > 0
         )
 
-    async def call_tool(
+    async def call_tool(  # type: ignore[override]
         self, name: str, arguments: dict[str, Any]
-    ) -> Sequence[ContentBlock] | dict[str, Any]:
+    ) -> Sequence[ContentBlock] | dict[str, Any] | CallToolResult:
```
DevonFulcher: Changing this type makes me feel a bit nervous. I'll look into this more.

jairus-m: Yes - agreed about the concern! Had a blurb in the PR description calling it out and wanted to make sure I got your help considering this more critically.

src/dbt_mcp/mcp/server.py
```diff
@@ -108,12 +110,10 @@ async def call_tool(
...
-            return [
-                TextContent(
-                    type="text",
-                    text=str(e),
-                )
-            ]
+            return CallToolResult(
+                content=[TextContent(type="text", text=str(e))],
+                isError=True,
+            )
```
DevonFulcher: I did some testing. I think all we need is this and we don't have to change the types.
```suggestion
            raise
```

jairus-m: Oh i see - just made this change locally and mypy is happy. I was a bit confused about how this propagates properly, but i think it makes sense to me:

1. Raise is not a return, so types are fine
2. When we raise, the error gets propagated to FastMCP's handler and the exception is caught

I did test the behavior and it works the same as expected. Thank you @DevonFulcher  for the help here.

b-per (approved):

PR #782, opened 2026-05-19
feat(semantic-layer): add get_dimension_values tool

b-per:
## Summary

- Adds a `get_dimension_values` tool to the Semantic Layer toolset
- Returns distinct values for a named dimension, optionally scoped to specific metrics
- Results are capped at a configurable `limit` (default 100); a `truncated` flag signals when more values exist beyond the limit
- Intended to be called before `query_metrics` to discover valid filter values for `where` clauses

## Implementation notes

- Uses `SyncSemanticLayerClient.dimension_values` (ADBC / Arrow Flight SQL), which is lightweight — no metric aggregation
- Return type is `DimensionValuesResponse(values: list[str], truncated: bool)`; all column values are stringified for consistent handling regardless of the underlying dimension type (dates, integers, etc.)
- `limit` is validated as `ge=1` to prevent empty-but-non-truncated responses
- Error handling follows the same pattern as other fetcher methods (`_format_semantic_layer_error`)

## Test plan

- [x] Unit tests: `uv run pytest tests/unit/semantic_layer/test_client.py -k "dimension_values" -v`
- [x] Full unit suite: `uv run pytest tests/unit/ -x -q`
- [x] Linting and type checks: `task check`
- [ ] Smoke test against a real dbt platform environment with the Semantic Layer enabled


src/dbt_mcp/semantic_layer/client.py
```diff
@@ -350,6 +357,30 @@ async def get_entities(
...
+                    sl_client.dimension_values,
+                    metrics=metrics or [],
+                    group_by=dimension,
+                )
+            raw: list[str] = [
+                str(v) if v is not None else ""
+                for v in raw_table.column(dimension).to_pylist()
+            ]
+            truncated = len(raw) > limit
+            return DimensionValuesResponse(values=raw[:limit], truncated=truncated)
```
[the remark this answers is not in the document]
b-per: The dbtsl SDK's `dimension_values` method doesn't expose a server-side limit parameter, so client-side truncation is the best we can do today. Added a comment in the code to document this. A server-side limit would require an SDK change.

docs/diagram.mmd
jairus-m, line 1: Nearly all docs today are automated via existing scripts in `scripts/` and the use of `task docs:generate`. This mermaid diagram is not included in that and will drift.

Should we (1) include this mermaid generation in docs automation in this PR or (2) exclude the file from here and open another PR for this addition in particular?

src/dbt_mcp/semantic_layer/client.py
```diff
@@ -350,6 +357,30 @@ async def get_entities(
...
+            with sl_client.session():
+                raw_table: Any = await asyncio.to_thread(
+                    sl_client.dimension_values,
+                    metrics=metrics or [],
+                    group_by=dimension,
+                )
+            # SDK doesn't support server-side limiting; truncation is applied client-side.
+            raw: list[str] = [
+                str(v) for v in raw_table.column(dimension).to_pylist() if v is not None
+            ]
```
jairus-m, line 386: I tested this locally and a bug came up - the SDK looks to return column names in uppercase and if the dimension is passed as lowercase, there is a mismatch and we get `Error: Error executing tool get_dimension_values: 'Field "<field_name>" does not exist in schema'`

I was able to fix it by adding a change locally to implement a case-insensitive lookup

tests/unit/semantic_layer/test_client.py
jairus-m, line 1: Non-blocker! 

`mock_sl_client` setup and `SemanticLayerConfig` construction are copy-pasted across all the tests here - may be worth to extract them into class level fixtures for sake of being DRY

src/dbt_mcp/semantic_layer/tools.py
```diff
@@ -28,6 +30,7 @@
 )
 from dbt_mcp.semantic_layer.types import (
     DimensionToolResponse,
+    DimensionValuesResponse,
     EntityToolResponse,
     GetMetricsCompiledSqlSuccess,
     ListMetricsResponse,
```
jairus-m, line 36: Do we also need to add these and register the new tool in `src/dbt_mcp/semantic_layer/tools_multiproject.py` as well? SL tools are registered in both the single and multiproject modules.

src/dbt_mcp/semantic_layer/client.py
```diff
@@ -350,6 +357,30 @@ async def get_entities(
...
+            sl_client = await self.client_provider.get_client(config=config)
+            with sl_client.session():
+                raw_table: Any = await asyncio.to_thread(
+                    sl_client.dimension_values,
+                    metrics=metrics or [],
+                    group_by=dimension,
+                )
+            raw: list[str] = [
+                str(v) if v is not None else ""
+                for v in raw_table.column(dimension).to_pylist()
```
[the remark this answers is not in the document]
b-per: Fixed in commit b685479 — null values are now omitted from the returned list rather than coerced to empty strings.

.changes/unreleased/Bug Fix-20260519-142146.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Bug Fix
+body: 'Fix get_dimension_values: widen values type to list[Any], add error handling, add limit >= 1 validation'
```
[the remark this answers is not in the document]
b-per: The changelog entry was already updated in commit 81f58a7 to reflect the actual behavior (values are stringified to `list[str]`, nulls omitted). The current body accurately describes what the code does.

src/dbt_mcp/tools/toolsets.py
```diff
@@ -53,6 +53,7 @@ class Toolset(Enum):
         ToolName.LIST_SAVED_QUERIES,
         ToolName.GET_DIMENSIONS,
         ToolName.GET_ENTITIES,
+        ToolName.GET_DIMENSION_VALUES,
```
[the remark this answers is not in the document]
b-per: Fixed in commit 7eebbb7 — added `get_dimension_values` to `tools_multiproject.py` and `MULTIPROJECT_SEMANTIC_LAYER_TOOLS`.

docs/diagram.mmd
b-per: Good catch — removed the file in commit 94959e2. We'll handle diagram automation in a separate PR.

src/dbt_mcp/semantic_layer/client.py
b-per: Fixed in commit 7eebbb7 — the column name is now looked up case-insensitively against the Arrow schema before calling `.column()`.

tests/unit/semantic_layer/test_client.py
b-per: Done in commit 94959e2 — extracted the five tests into a `TestGetDimensionValues` class with shared `mock_sl_client`, `sl_config`, and `dim_fetcher` fixtures.

src/dbt_mcp/semantic_layer/tools.py
b-per: Yes, fixed in commit 7eebbb7 — added `get_dimension_values` to `tools_multiproject.py` with the `project_id` parameter and registered it in `MULTIPROJECT_SEMANTIC_LAYER_TOOLS`.

src/dbt_mcp/semantic_layer/client.py
```diff
@@ -112,6 +113,12 @@ def compile_sql(
         read_cache: bool = True,
     ) -> str: ...
 
+    def dimension_values(
+        self,
+        metrics: list[str],
+        group_by: str,
+    ) -> Any: ...
```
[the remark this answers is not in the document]
b-per: Good point on `raw_table`. Tightened it to `pa.Table` in commit fe36483. For the Protocol, the SDK stubs type `dimension_values` as `-> list[str]` (even though it returns an Arrow table at runtime), so the Protocol must keep `Any` to avoid a mypy incompatibility with `SyncSemanticLayerClient`. The tighter local annotation still restores mypy coverage for all downstream attribute accesses.

src/dbt_mcp/semantic_layer/client.py
```diff
@@ -350,6 +357,38 @@ async def get_entities(
...
+        self,
+        config: SemanticLayerConfig,
+        dimension: str,
+        metrics: list[str] | None = None,
+        limit: int = 100,
+    ) -> DimensionValuesResponse:
+        try:
+            sl_client = await self.client_provider.get_client(config=config)
+            with sl_client.session():
+                raw_table: Any = await asyncio.to_thread(
```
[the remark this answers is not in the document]
b-per: Fixed in commit fe36483 — `raw_table` is now annotated as `pa.Table`.

jairus-m (approved):

Issue #759, opened 2026-05-07
[Bug]: Cursor - v1.17.1 No tools, prompts, or resources


pauln-relay:
### What happened?

Recent versions of dbt mcp, when using dbt core, are broken and display "No tools, prompts, or resources" in the Cursor when the MCP is enabled. This is new behavior as of 1.17.

Originally, we used the opt-out approach for features, but I discovered this is still the case for opt-in (See reproduction below)

The [documentation](https://docs.getdbt.com/docs/dbt-ai/mcp-environment-variables?version=2.0#disable-mode-default) could also be clearer about which MCP functions are tied to dbt Cloud and/or core and what "SQL Tools" means.

### Steps to Reproduce

**Working (1.15.0):**

<img width="1468" height="92" alt="Image" src="https://github.com/user-attachments/assets/e0aa09f5-a517-4451-b612-3358284a3b03" />

```json
        "dbt Labs": {
            "command": "uvx",
            "args": [
                "dbt-mcp@1.15.0"
            ],
            "env": {
                "DBT_PROJECT_DIR": "dbt/analytics",
                "DBT_PROFILES_DIR": "dbt/analytics",
                "DBT_PATH": ".venv/bin/dbt",
                "DBT_CLI_TIMEOUT": "900",
                "DBT_MCP_ENABLE_DBT_CLI": "true"
            },
        }
```

**Not Working (1.17.1):**

<img width="1470" height="101" alt="Image" src="https://github.com/user-attachments/assets/1a21d58e-6395-44ec-a35b-cc9e290f3646" />

```json
        "dbt Labs": {
            "command": "uvx",
            "args": [
                "dbt-mcp@1.17.1"
            ],
            "env": {
                "DBT_PROJECT_DIR": "dbt/analytics",
                "DBT_PROFILES_DIR": "dbt/analytics",
                "DBT_PATH": ".venv/bin/dbt",
                "DBT_CLI_TIMEOUT": "900",
                "DBT_MCP_ENABLE_DBT_CLI": "true"
            },
        }
```

**See log output attached for v1.17.1 output**

### Deployment

Local MCP server

### Relevant log output

```shell
2026-05-07 09:59:10.851 [error] INFO [dbt_mcp.mcp.server] Starting MCP server
 undefined
2026-05-07 09:59:10.851 [error] WARNING [dbt_mcp.mcp.server] Could not resolve credentials — defaulting to single-project mode: DBT_HOST is a required environment variable
 undefined
[...]

PR #787, opened 2026-05-23
fix: wire elicitation into platform config providers

### LINKED ISSUE #762: fix: elicitation fires for wrong users — DBT_HOST=None can't serve as both 'no platform' and 'please elicit host'
## Background

PR #728 added MCP elicitation so users without `DBT_HOST` could be prompted to enter it interactively instead of getting a config error. In doing so it also removed the old conditional guards that disabled platform tools when `DBT_HOST` was absent.

PR #760 had to bypass `ElicitingCredentialsProvider` in `_is_multi_project()` to fix #759 (CLI users seeing "No tools" in Cursor). As a result, **elicitation is now effectively disabled** — `ElicitingCredentialsProvider` is wired only to `Config.credentials_provider`, which is no longer called from any hot path.

## The core problem

`DBT_HOST=None` means two incompatible things at once:

1. **"I am a CLI-only user — I don't need platform tools"** (e.g. `DBT_MCP_ENABLE_DBT_CLI=true`, no host needed)
2. **"I am a platform user who hasn't configured my host yet"** (the target audience for elicitation)

There is no way to distinguish these two users from `DBT_HOST=None` alone. Any elicitation logic placed in the routing layer (`_is_multi_project`, `list_tools`, `call_tool`) will fire for both.

Additionally, `validate_dbt_platform_settings` checks legacy `DISABLE_*` flags and is blind to the newer `DBT_MCP_ENABLE_*` allowlist mode — so it incorrectly treats CLI-only users as needing platform credentials.

### Option 1 — New explicit opt-in flag

Add a new env var such as `DBT_MCP_ENABLE_ELICITATION=true`.

Elicitation only fires when:
- The flag is set, AND
- `DBT_HOST` is missing

**Pros:** Completely unambiguous — the user explicitly asked for interactive setup.  
**Cons:** One more env var to document; users need to know it exists.

---

### Option 2 — Reuse existing `DBT_MCP_ENABLE_*` toolset flags

Elicitation fires when:
- Any **platform** toolset is active (`SEMANTIC_LAYER`, `DISCOVERY`, `ADMIN_API`, `SQL`), AND
- `DBT_HOST` is missing

"Platform toolsets" need a host. "Local toolsets" (`DBT_CLI`, `DBT_CODEGEN`, `DBT_LSP`, `PRODUCT_DOCS`, `MCP_SERVER_METADATA`) never do. In default mode (no `ENABLE_*` flags set, all tools registered) platform toolsets are active, so elicitation fires — which is the right default for users who haven't configured anything yet.

**Pros:** No new env var; reuses existing, well-understood configuration.  
**Cons:** Requires `validate_dbt_platform_settings` (or the elicitation trigger) to understand allowlist mode; slightly more logic.

---

### Option 3 — Remove elicitation entirely

Revert to pre-#728 behaviour: `DBT_HOST=None` means local/CLI tools only; platform tools require `DBT_HOST` to be set explicitly.

**Pros:** Simplest; no ambiguity; power users are comfortable setting env vars.  
**Cons:** Loses the interactive setup UX for non-technical users; the MCP elicitation infrastructure built in #728 goes unused.

---

## Implementation note (applies to options 1 and 2)

Wherever elicitation is triggered, it must fire from **inside the platform config providers** (semantic layer, discovery, admin API), not from the routing layer. `ElicitingCredentialsProvider` should be passed to those providers so elicitation fires only when a platform tool actually needs a host — never for CLI-only tool calls.

Muizzkolapo:
## Summary

The elicitation wizard from #728 was wired up but never actually reachable. Every platform config provider was handed the raw `CredentialsProvider`, so the `ElicitingCredentialsProvider` wrapper sat there doing nothing on any hot path.

This PR fixes that by passing the eliciting wrapper to platform providers when platform toolsets are active. CLI-only users never see the wizard because they don't need a host.

It also root-causes the Cursor timeout from #759. The old `_is_multi_project()` was calling `get_credentials()` on every `list_tools()`, which could block on elicitation or OAuth. That's what caused the 17-second hang and the "No tools" state. #760 worked around it by bypassing the eliciting provider, but the real problem was the I/O itself. Now `_is_multi_project()` just reads `settings.dbt_project_ids` directly — sync, no I/O, no credential fetch.

## What changed

- **`CredentialsProviderProtocol`** — structural typing so provider constructors accept either the raw or eliciting credentials. No class hierarchy change needed.
- **`load_config()` picks the right wrapper** — checks `settings.any_platform_toolset_active`. If any platform toolset (semantic layer, discovery, admin API, SQL) is on, platform providers get the eliciting wrapper. If you're CLI-only, they get raw credentials. The `any_platform_toolset_active` property understands both the `ENABLE_*` allowlist mode and the legacy `DISABLE_*` flags, which is what `validate_dbt_platform_settings` was blind to.
- **`_is_multi_project()` is sync now** — reads settings directly instead of fetching credentials. This kills the Cursor timeout at the root rather than working around it.
- **Proxied tools stay on raw credentials** — they register at lifespan time where there's no MCP request context, so elicitation can't run there anyway.

One trade-off worth calling out: users who get `dbt_project_ids` from OAuth (not env var) won't see multi-project mode until the first platform tool call triggers credential resolution. Setting `DBT_PROJECT_IDS` env var avoids this.

## Testing

**list_tools returns instantly with no hang or credential fetch:**
image

**Calling a platform tool triggers elicitation:**
image
image

**Non-platform tools unaffected:**
image

- [x] `list_tools()` returns instantly
- [x] Platform tool without `DBT_HOST` triggers elicitation
- [x] CLI-only mode — no wizard, no platform errors
- [x] Non-elicitation clients get clean error message
- [x] 700 unit tests pass

Fixes #762


jairus-m (commented):
> One trade-off worth calling out: users who get dbt_project_ids from OAuth (not env var) won't see multi-project mode until the first platform tool call triggers credential resolution. Setting DBT_PROJECT_IDS env var avoids this.
 
Appreciate the work here!

I want to flag a "product" concern about the trade-off in the PR description. The tradeoff hits remote MCP users using OAuth that we just [released for private beta](https://docs.getdbt.com/docs/platform/manage-access/connect-apps-oauth?version=2.0&name=Fusion#use-with-remote-mcp). A big win of OAuth for remote MCP users is not having to set any config/env vars or having to do work arounds. We've learned from the local MCP that these configurations can be a _rocky_ experience for users and remote + OAuth is supposed to be the antidote!

I'm afraid the current implementation undercuts that. 

I'm still trying to understand the approach of the PR and what's changed but this was the first thing that stood out to me.

Muizzkolapo:
> > One trade-off worth calling out: users who get dbt_project_ids from OAuth (not env var) won't see multi-project mode until the first platform tool call triggers credential resolution. Setting DBT_PROJECT_IDS env var avoids this.
> 
> Appreciate the work here!
> 
> I want to flag a "product" concern about the trade-off in the PR description. The tradeoff hits remote MCP users using OAuth that we just [released for private beta](https://docs.getdbt.com/docs/platform/manage-access/connect-apps-oauth?version=2.0&name=Fusion#use-with-remote-mcp). A big win of OAuth for remote MCP users is not having to set any config/env vars or having to do work arounds. We've learned from the local MCP that these configurations can be a _rocky_ experience for users and remote + OAuth is supposed to be the antidote!
> 
> I'm afraid the current implementation undercuts that.
> 
> I'm still trying to understand the approach of the PR and what's changed but this was the first thing that stood out to me.



> > One trade-off worth calling out: users who get dbt_project_ids from OAuth (not env var) won't see multi-project mode until the first platform tool call triggers credential resolution. Setting DBT_PROJECT_IDS env var avoids this.
> 
> Appreciate the work here!
> 
> I want to flag a "product" concern about the trade-off in the PR description. The tradeoff hits remote MCP users using OAuth that we just [released for private beta](https://docs.getdbt.com/docs/platform/manage-access/connect-apps-oauth?version=2.0&name=Fusion#use-with-remote-mcp). A big win of OAuth for remote MCP users is not having to set any config/env vars or having to do work arounds. We've learned from the local MCP that these configurations can be a _rocky_ experience for users and remote + OAuth is supposed to be the antidote!
> 
> I'm afraid the current implementation undercuts that.
> 
> I'm still trying to understand the approach of the PR and what's changed but this was the first thing that stood out to me.

Drafting the follow-up issue as discussed and awaiting feedback from the team. I believe the best approach is to remove elicitation, as the value it provides does not justify the number of changes required to make it work seamlessly.

jairus-m:
Will close - refer to https://github.com/dbt-labs/dbt-mcp/pull/802 - thanks Muizz for your work as always! Glad we could align on this.

PR #788, opened 2026-05-29
fix: change list_metrics search type to list[str] | None

### LINKED ISSUE #784: list_metrics incompatible with Google ADK (Vertex AI backend) due to union type schema
## Summary

Since v1.19, `list_metrics` fails when used via Google ADK with a Vertex AI backend:

```
list_metrics functionDeclaration parameters.search schema didn't specify the schema type field
```

## Root cause

PR #768 widened the `search` parameter from `str | None` to `str | list[str] | None`. This 3-way union generates a JSON Schema with `anyOf` and no top-level `type` field:

```json
{
  "anyOf": [
    {"type": "string"},
    {"type": "array", "items": {"type": "string"}},
    {"type": "null"}
  ]
}
```

The Vertex AI backend of `google-genai` (`Schema.from_json_schema()`) requires a top-level `type` field and rejects this. The Gemini API backend handles it correctly.

## Affected versions

v1.19+ (introduced in #768)

## Upstream bug

This is tracked as a P2 bug in the `google-genai` library: [googleapis/python-genai#1807](https://github.com/googleapis/python-genai/issues/1807). The ADK issue [google/adk-python#3424](https://github.com/google/adk-python/issues/3424) was closed pointing there.

## Workaround (client-side)

Pin to ADK v1.17.0 and `google-genai==1.44.0` until the upstream fix lands.

## Potential fix (dbt-mcp side)

Change `search` to `list[str] | None` (drop the `str` shorthand) — a 2-way union that google-genai handles correctly. LLMs pass a single-element list for single searches. This is a breaking change for callers passing a plain string, so pending feedback.

arsenelupin14:
Fixes #784.

Addresses the JSON schema parsing issue encountered with the Google ADK (Vertex AI backend) when generating the schema for `list_metrics`.

Fix:
- Removed `str` from the `search` parameter type union across `client.py`, `tools.py`, and `tools_multiproject.py`. The parameter is now strictly `list[str] | None`.
- Updated `client.py` logic to process only list and None inputs, removing the fallback handling for bare strings.
- Updated the unit tests in `test_client.py` to pass `list[str]` instead of bare strings, aligning with the new signature and preventing unintended character iteration.

### DIFF
--- a/.changes/unreleased/fix-list-metrics-schema.yaml
+++ b/.changes/unreleased/fix-list-metrics-schema.yaml
@@ -1,0 +1,2 @@
+kind: Bug Fix
+body: Change `list_metrics` search parameter to `list[str] | None` to resolve Vertex AI schema union error.
--- a/src/dbt_mcp/semantic_layer/client.py
+++ b/src/dbt_mcp/semantic_layer/client.py
@@ -145,12 +145,12 @@ class SemanticLayerFetcher:
     async def list_metrics(
         self,
         config: SemanticLayerConfig,
-        search: str | list[str] | None = None,
+        search: list[str] | None = None,
     ) -> ListMetricsResponse:
-        # `search` may be a single substring or a list of substrings; for a list
-        # we fan out one GraphQL call per substring, then merge & dedupe by name.
+        # `search` is a list of substrings; we fan out one GraphQL call
+        # per substring, then merge & dedupe by name.
         search_terms: list[str | None]
-        if isinstance(search, list):
+        if search is not None:
             # Strip whitespace, drop empty values, and dedupe identical terms
             # (preserving first-seen order) so a whitespace-only or duplicated
             # term can't broaden the fan-out into redundant or no-filter calls.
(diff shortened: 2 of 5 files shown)


jairus-m:
Hello thanks for the contribution! Can you read the `CONTRIBUTING.md` file and make sure you hit all the asks? The main ones being (1) verified commits and (2) having a change log entry.

Update: Also re-reading the issue, I see that there are two options to address the issue. This option you implemented would be a breaking change so curious to hear your thought process for choosing one over the other as the issue noted "pending feedback" before implementing but no discussion was had.

arsenelupin14:
@jairus-m I initially dropped the string union entirely as proposed in the original issue ("This is a breaking change for callers passing a plain string, so pending feedback."), since a 3-way union blocks Vertex AI users from calling it altogether.

To resolve this without introducing a breaking change for existing callers, I've just pushed a follow-up commit that retains the stricter `list[str] | None` type hint (which generates a 2-way schema that Vertex AI supports natively) but uses Pydantic's `BeforeValidator` to automatically coerce any incoming bare strings into a single-element list.

b-per:
I am not convinced that we need to update this MCP server implementation because of bugs in other systems/libraries.

Why impact all MCP clients for a bug in one of them.

jairus-m:
I'm aligned with Benoit here. I appreciate the fix, but using Pydantic's BeforeValidator is a bit hacky since the code schema and runtime schema will be different. This is not to say the iteration is not valuable , but we feel it's best to keep the workaround on the client-side and document the pin to unblock folks. If this issue grows in noise or the upstream fix drags, happy to revisit!

PR #790, opened 2026-06-03
Propagate operational errors from semantic layer query_metrics

wiggzz:
# Why

`SemanticLayerFetcher.query_metrics` caught every exception and flattened it into a returned `QueryMetricsError` string. That made two very different situations indistinguishable to callers: a query the semantic layer processed and rejected (user/agent-fixable) versus an operational failure such as an authentication or connection error (the semantic layer being unreachable). A caller that wants to surface query errors to a model for self-correction while still detecting operational failures had no way to tell them apart.

# What

`query_metrics` now returns a `QueryMetricsError` only for genuine query-level failures (`QueryFailedError`). Operational errors (auth, connection, transport, etc.) propagate as exceptions instead of being swallowed.

Timeout behavior is unchanged: COMPILED-status timeouts still raise `SemanticLayerQueryTimeoutError`, and RUNNING/unknown-status timeouts still return a `QueryMetricsError`.

Added unit tests covering the three cases: a query failure is returned as data, and an execution-time operational error and a client-acquisition error both propagate.

---
Drafted by Claude Opus 4.8 under the direction of @wiggzz.

### DIFF
--- a/.changes/unreleased/Bug Fix-20260603-090806.yaml
+++ b/.changes/unreleased/Bug Fix-20260603-090806.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Semantic layer query_metrics now lets operational errors (authentication, connection) propagate instead of flattening them into a returned query-error string, so callers can distinguish an invalid query from an unreachable semantic layer.
+time: 2026-06-03T09:08:06.019505-05:00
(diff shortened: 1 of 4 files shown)


jairus-m (approved):

wiggzz:
I think the integration test needs to be updated.

PR #796, opened 2026-06-05
Return dimension values query errors as tool results

wiggzz:
# Why

`get_dimension_values` can fail because the model supplied an invalid dimension or metric. Those query-level failures should be returned to the model for self-correction instead of being raised as generic runtime errors.

# What

This adds a typed dimension-values error result from the Semantic Layer fetcher and has the public single-project and multi-project tools return either the success response or typed error response. Operational Semantic Layer failures still propagate as real exceptions.

# Notes

Drafted by GPT-5 Codex under the direction of @wiggzz.


jairus-m (approved):
Nice ty

PR #802, opened 2026-06-10
Revert host elicitation (pre-v1.17.0 behaviour)

### LINKED ISSUE #762: fix: elicitation fires for wrong users — DBT_HOST=None can't serve as both 'no platform' and 'please elicit host'

jairus-m:
## Summary

Removing host elicitation - based off talks with @Muizzkolapo , the complexity and scope of implementation touches too many things for the value this feature brings. In addition, host elicitation has brought up some bugs that have had to be squashed since released. 

## Related Issues

Closes #762 

## Checklist


b-per (approved):
Thanks @Muizzkolapo and @jairus-m for checking if elicitation could work in the MCP server and deep dive into ways it could integrate with the rest of the config!

PR #806, opened 2026-06-15
fix: wrap get_model_health result in a list to match list[dict] schema

theyostalservice:
## Why

`get_model_health` was failing response validation on every call with:

```
Input should be a valid list [type=list_type, input_type=dict]
```

The tool is declared `-> list[dict]` with `structured_output=True`, so FastMCP generates a Pydantic output model that expects a JSON array. But `fetch_model_health` returned a bare dict (`edges[0]["node"]`), causing Pydantic to reject the response on every invocation. The underlying data was present — this was purely a response-schema mismatch.

## What

Wraps the returned node in a list so the result matches the declared `list[dict]` contract — consistent with how `fetch_model_parents` and `fetch_model_children` return their respective list fields. The empty-model case (`return []`) was already correct.

Adds a regression test covering both the found-model case (result is a `list` of length 1) and the empty-model case.

Drafted by claude-opus-4-8 under the direction of @theyostalservice

### DIFF
--- a/src/dbt_mcp/config/config.py
+++ b/src/dbt_mcp/config/config.py
@@ -104,7 +104,7 @@ class Config:
     # product_docs tool actually needs the version, then cached for the
     # lifetime of the closure.
     dbt_version_provider: Callable[[], str | None] = field(
-        default_factory=lambda: (lambda: None)
+        default_factory=lambda: lambda: None
     )
 
 
--- a/src/dbt_mcp/discovery/client.py
+++ b/src/dbt_mcp/discovery/client.py
@@ -577,7 +577,7 @@ class ModelsFetcher:
         edges = result["data"]["environment"]["applied"]["models"]["edges"]
         if not edges:
             return []
-        return edges[0]["node"]
+        return [edges[0]["node"]]
 
 
 class ExposuresFetcher:
--- a/tests/unit/discovery/test_models_fetcher.py
+++ b/tests/unit/discovery/test_models_fetcher.py
@@ -1,0 +1,66 @@
+from unittest.mock import Mock
+
+import pytest
+
+from dbt_mcp.discovery.client import ModelsFetcher
+
+
+@pytest.fixture
+def models_fetcher():
+    return ModelsFetcher(paginator=Mock())
+
+
+async def test_fetch_model_health_wraps_single_node_in_list(
+    models_fetcher, mock_api_client, unit_discovery_config
+):
+    """fetch_model_health must return a list[dict], not a bare dict."""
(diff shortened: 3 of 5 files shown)


tests/unit/lsp/test_lsp_connection.py
```diff
@@ -950,8 +950,8 @@ async def mock_recv_wrapper(size):
         ):
             mock_loop = MagicMock()
             mock_get_loop.return_value = mock_loop
-            mock_loop.run_in_executor.side_effect = (
-                lambda _, func, *args: mock_recv_wrapper(*args)
+            mock_loop.run_in_executor.side_effect = lambda _, func, *args: (
+                mock_recv_wrapper(*args)
```
theyostalservice, line 954: these are just linting changes

src/dbt_mcp/config/config.py
```diff
@@ -104,7 +104,7 @@ class Config:
     # product_docs tool actually needs the version, then cached for the
     # lifetime of the closure.
     dbt_version_provider: Callable[[], str | None] = field(
-        default_factory=lambda: (lambda: None)
+        default_factory=lambda: lambda: None
```
theyostalservice, line 107: just style/linting

src/dbt_mcp/discovery/client.py
```diff
@@ -577,7 +577,7 @@ async def fetch_model_health(
         edges = result["data"]["environment"]["applied"]["models"]["edges"]
         if not edges:
             return []
-        return edges[0]["node"]
+        return [edges[0]["node"]]
```
theyostalservice, line 580: this is the only real change, plus the tests.

tests/unit/discovery/test_models_fetcher.py
```diff
@@ -0,0 +1,66 @@
...
+
+    result = await models_fetcher.fetch_model_health(
+        unique_id="model.project.my_model", config=unit_discovery_config
+    )
+
+    assert isinstance(result, list), (
+        "fetch_model_health must return a list, not a bare dict"
+    )
+    assert len(result) == 1
+    assert result[0] == node
```
theyostalservice, line 35: this failed previously

jairus-m (approved):
Thank you!

b-per:
It doesn't matter too much, but was this the correct fix? Instead of returning a `list(dict)` with 0 or 1 item, we should just return a `Optional[dict]`, no?

(but i also know that now that the MCP server will be published to some new marketplaces it will become much more difficult to change the "shape" or our tools)

PR #815, opened 2026-06-18
Make the --quiet flag configurable on dbt CLI tools

### LINKED ISSUE #808: --quiet flag suppresses all CLI output for test, compile, and other CLI-based MCP tools when using Cloud CLI
**Summary**
The dbt MCP server hardcodes a --quiet flag when executing CLI commands (e.g., test, compile, build, run, docs, parse, list). When using the dbt Cloud CLI binary, this causes all meaningful output to be suppressed — only the invocation ID is returned to the LLM client.

**Steps to Reproduce**
1. Configure dbt-mcp with DBT_PATH pointing to the dbt Cloud CLI binary.
2. Use an MCP-compatible client (e.g., Claude Code) to invoke the test tool with a valid node_selection (e.g., model_name,test_type:unit).
3. Observe that the only response returned is:
`{"result": "Created invocation id <uuid>\n"}`
4. No test results, pass/fail status, or compilation output is returned.

**Expected Behavior**
The tool should return meaningful output (test results, compilation details, etc.) so the LLM can interpret and relay results to the user without requiring them to manually check the platform.

**Actual Behavior**
The --quiet flag strips all output, and the LLM interprets the invocation ID as an async job reference — leading to user confusion about whether the command completed and what the results were.

**Based on my testing these seems to be the affected tools (please double check):**
- test
- compile
- build
- run
- docs
- parse
- list

**Potential Root Cause:**
The --quiet flag is hardcoded here with no override mechanism: https://github.com/dbt-labs/dbt-mcp/blob/main/src/dbt_mcp/dbt_cli/tools.py
(around lines 107–112)

**Workaround**
I've disabled the affected MCP tools and instruct the LLM to run dbt commands directly in the terminal instead. Full output is returned as expected when bypassing the MCP server.

jairus-m:
## Summary

Adds an `is_quiet` parameter to all dbt CLI tools that previously hardcoded `--quiet`.

## What Changed

- Added `is_quiet: bool = True` parameter to `build`, `compile`, `run`, `test`, `docs`, `parse`, `list`, and `clone`
- Added prompt description for the new parameter
- Added unit tests covering explicit opt-out and opt-in

## Why

The `--quiet` flag was intentionally hardcoded to reduce token noise. This PR preserves that as the default while letting users set `is_quiet=False` when they need full output — for example, when using the dbt Cloud CLI where `--quiet` suppresses all meaningful output.

## Related Issues

Closes #808

## Checklist

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20260617-185651.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20260617-185651.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Make the --quiet flag configurable on dbt CLI tools
+time: 2026-06-17T18:56:51.545347-07:00
--- a/src/dbt_mcp/dbt_cli/tools.py
+++ b/src/dbt_mcp/dbt_cli/tools.py
@@ -2,7 +2,7 @@ import json
 import os
 import subprocess
 from collections.abc import Iterable
-from typing import Any
+from typing import Annotated, Any
 import logging
 
 from mcp.server.fastmcp import FastMCP
@@ -59,6 +59,7 @@ def create_dbt_cli_tool_definitions(config: DbtCliConfig) -> list[ToolDefinition
         sample: str | None = None,
         yml_selector: str | None = None,
         state_path: str | None = None,
+        is_quiet: bool = True,
     ) -> str:
         try:
             # Commands that should always be quiet to reduce output verbosity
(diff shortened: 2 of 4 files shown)


b-per:
We can't really do that easily now that we published the MCP server on OpenAI and Claude marketplaces, can we?

Also, shouldn't the fix be done in the Cloud CLI directly so that its behavior is aligned with the other CLIs?

jairus-m:
> Also, shouldn't the fix be done in the Cloud CLI directly so that its behavior is aligned with the other CLIs?

Def didnt register that Cloud CLI is the only one that has the issue - you are right! Working on an upstream fix.

PR #817, opened 2026-06-22
feat(discovery): enrich get_lineage with description; remove get_model_parents/children

theyostalservice:
# Why

`get_model_parents` and `get_model_children` are redundant with `get_lineage(depth=1)`. Every registered tool name + description ships as context overhead on every turn; removing two tools reduces token cost for any harness using the Discovery toolset.

`get_lineage` lacked the `description` field that the parents/children tools returned per node, so we enrich it first to preserve information parity.

# What

- Add `description` to the lineage node selection in `get_full_lineage.gql` so callers can read node descriptions without a separate details call
- Update `get_lineage.md` prompt to document the new field and note that `depth=1` replaces the removed tools
- Remove `get_model_parents` and `get_model_children` tools, prompt files, GraphQL queries, and client methods
- Update enum/toolset/README mappings in lockstep

# Refs

DI-4002

---
Drafted by Claude Opus 4.8 under the direction of @theyostalservice


src/dbt_mcp/prompts/discovery/get_lineage.md
```diff
@@ -7,27 +7,33 @@ A list of all nodes in the connected subgraph, where each node contains:
 - `uniqueId`: The resource's unique identifier
 - `name`: The resource name
 - `resourceType`: The type of resource (Model, Source, etc.)
+- `description`: A description of the resource
 - `parentIds`: List of unique IDs that this resource directly depends on
 
+**Getting immediate parents or children:** use `depth=1`. This returns the target node plus its direct upstream dependencies and direct downstream dependents.
```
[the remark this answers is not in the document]
jairus-m: Agree with the above! `get_lineage` does not return parent or child but does both . Many references may need to be updated to make "or" --> "and". Or maybe we can update `get_lineage` to return either parents or children to maintain the "or".

.changes/unreleased/Enhancement or New Feature-20260622-103011.yaml
jairus-m, line 1: I feel this would be considered a breaking change? One of the examples actually (if you Ctrl + F for the deleted tools), actually refers to both `get_model_parents` and `get_model_children` in `examples/ag2_agent/main_multiagent.py` which is a solid case of potentially existing references to these tools that would break

cc @b-per thoughts here?

b-per: Yes. Same as the other PR. This should be a breaking change. We need an approach to
- potentially move dbt-mcp to 2.x
- plan the steps required to keep the openai and anthropic apps working

theyostalservice:
Closing in favor of a new approach. The original implementation removes tools immediately, but the dbt MCP server is now a formally submitted app in the Anthropic and OpenAI app stores — which requires a deprecate-then-remove lifecycle before permanent removal. A replacement PR for this work is being opened as part of DI-4651 (lineage enrichment + deprecation approach, additive only).

PR #820, opened 2026-06-25
Fix list_saved_queries tool which was calling savedQueriesPaginated instead of savedQueries

jairus-m:
## Summary
`list_saved_queries` was always failing because it called `savedQueriesPaginated`, a field that does not exist in the dbt Semantic Layer API. The fix switches to the non-paginated `savedQueries` query, moves search filtering client-side, and adds a fallback for `queryParams` subfields that can time out server-side.

## What Changed
- `gql.py`: Replace the `savedQueriesPaginated` query with two queries — `saved_queries` (lightweight, always succeeds) and `saved_queries_with_params` (adds `queryParams`; attempted with a 5s timeout, falls back on failure).
- `client.py`: Rewrite `list_saved_queries` to call the simple query first, then optimistically attempt the with_params query; fall back to simple results on any exception. Move `search` filtering client-side. Fix `queryParams` null guard (`or {}` instead of `.get("queryParams", {})`).
- `types.py`: Fix `SavedQueryToolResponse.where` annotation from `list[str] | None` to `str | None` to match the actual API shape.
- `tests/`: Update mock response shape and add coverage for client-side search, null handling, the with_params happy path, and the timeout fallback.

## Why
`savedQueriesPaginated` was never added to the dbt Semantic Layer API. The original implementation copied the pattern from `list_metrics`/`list_dimensions` which had already migrated to `*Paginated` variants — but no paginated variant exists for saved queries. Every call to `list_saved_queries` resulted in a GQL error (`Cannot query field 'savedQueriesPaginated' on type 'Query'`).

## Related Issues
Closes #
Related to #

## Checklist

## Additional Notes
Bug reported in the dbt community Slack: https://getdbt.slack.com/archives/C08NGJYD563/p1782324792705029

### DIFF
--- a/src/dbt_mcp/semantic_layer/types.py
+++ b/src/dbt_mcp/semantic_layer/types.py
@@ -53,7 +53,7 @@ class SavedQueryToolResponse:
     description: str | None = None
     metrics: list[str] | None = None
     group_by: list[str] | None = None
-    where: list[str] | None = None
+    where: str | None = None
 
 
 @dataclass
--- a/.changes/unreleased/Bug Fix-20260624-205734.yaml
+++ b/.changes/unreleased/Bug Fix-20260624-205734.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Fix list_saved_queries tool which was calling savedQueriesPaginated (a non-existent API field) instead of savedQueries
+time: 2026-06-24T20:57:34.35722-07:00
--- a/src/dbt_mcp/semantic_layer/client.py
+++ b/src/dbt_mcp/semantic_layer/client.py
@@ -319,6 +319,8 @@ class SemanticLayerFetcher:
         except Exception as e:
             logger.warning(f"Error fetching saved queries with params: {e}")
             results = simple_results
+        search = search.strip() if isinstance(search, str) else search
+        search = search if search else None
         if search:
             search_lower = search.lower()
             results = [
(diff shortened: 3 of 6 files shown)


src/dbt_mcp/semantic_layer/types.py
```diff
@@ -53,7 +53,7 @@ class SavedQueryToolResponse:
     description: str | None = None
     metrics: list[str] | None = None
     group_by: list[str] | None = None
-    where: list[str] | None = None
+    where: str | None = None
```
jairus-m, line 56: @b-per what do you think about this case? where the return type was incorrect and is now being corrected but is being flagged in the ChatGPT contract check?

b-per: You'll need to generate a new contract doc.

```
✗ contract snapshot is out of date.
The ChatGPT-published contract changed. Run: task contract:generate
```

Then, normally we'd need to do a lot of steps etc... (see Notion) to publish such a change. But in that case, as it was never working, I think that we are OK. (we might still need to push a new App version when the remote MCP points to the version of dbt-mpc with this fix)

b-per (approved):

src/dbt_mcp/semantic_layer/types.py
jairus-m: OK that makes sense - for these types of fixes the new process doesnt apply mostly bc the tool itself was already broken. 

But for future, we have to coordinate and resubmit an app request even if a small bug fix caused a contract change for a working tool.

PR #826, opened 2026-06-29
docs: add deprecate-then-remove process guide for MCP tools

theyostalservice:
# Why

Tools in dbt-mcp can't be removed immediately — the server is a published app in the
Anthropic and OpenAI app stores, and removing a tool is a breaking change for all
installed clients. Without a written process, it's easy to accidentally delete a tool
outright. This adds a canonical reference so every future deprecation follows the same
pattern.

# What

- **`docs/deprecating-tools.md`** — new standalone process guide covering:
  - Why direct removal breaks published clients (links to the existing
    `Published-app contract` section and `contract/snapshot.py`)
  - The 4-step deprecate-then-remove lifecycle
  - Mechanical deprecation checklist using the helpers in
    `src/dbt_mcp/tools/deprecation.py` (`deprecated_description` + `deprecation_meta`),
    with `get_model_parents` as a worked example from the stacked PR
  - Why the deprecated description should be short and blunt rather than the
    original prompt with a banner prepended — it makes a model less likely to
    pick the tool, which speeds the usage soak — plus an optional note on
    forwarding a deprecated tool to its replacement internally when the
    replacement is a true drop-in
  - What NOT to delete (ToolName, toolsets, readme_mappings, `*_TOOLS` lists, prompt
    file, client code)
  - A **Release & rollout** section: merging a dbt-mcp PR doesn't reach the
    published apps on its own — it has to be released, rolled into the hosted
    MCP service, and resubmitted to each app store afterward
  - The `ToolCalled` telemetry signal and the ~0 usage bar for safe removal,
    measured against the deployed service rather than the merged PR
  - Phase B removal checklist
- **`CONTRIBUTING.md`** — a `## Deprecating a tool` section (after `## Adding Tools`)
  that summarizes the lifecycle in two sentences and links to the new doc

Drafted by Claude Sonnet 4.6 under the direction of @theyostalservice

### DIFF
--- a/docs/deprecating-tools.md
+++ b/docs/deprecating-tools.md
@@ -1,0 +1,120 @@
+# Deprecating a tool
+
+Tools in dbt-mcp follow a **deprecate-then-remove** lifecycle rather than being
+deleted outright. This document explains why and how.
+
+## Why tools can't just be deleted
+
+The dbt MCP server is a **published app** in the Anthropic and OpenAI app stores.
+When an app is submitted, the host caches the server's tool surface — names, titles,
+descriptions, input/output schemas, annotations, and `_meta` — as a versioned
+contract at submission time. Removing or renaming a tool is a breaking change for
+all installed clients that depend on that tool string.
+
+A deploying the server immediately removes a tool from the live server, but the
+published contract (the snapshot submitted to the app store) still lists it. Clients
+using the published version see a mismatch; the tool disappears without warning.
+See `CONTRIBUTING.md` (§ Published-app contract) and
+`src/dbt_mcp/contract/snapshot.py` for technical details.
+
+## The deprecate-then-remove lifecycle
+
+1. **Add the replacement tool.** Ship the new, consolidated tool. Its contract
+   snapshot is submitted with the next app-store update.
+2. **Deprecate the old tool.** Keep it registered and callable. Apply a deprecation
+   banner to its description and a `deprecated` / `replacement` signal to its `meta`.
+   Publish a compatible app update.
+3. **Monitor usage.** Track calls to the deprecated tool name via the `ToolCalled`
+   telemetry event (see [Monitoring](#monitoring-usage-before-removal)). Wait until
+   usage falls to approximately zero over a trailing 30-day window.
+4. **Remove the tool.** Delete it in a separate PR, bump the major version, and
+   submit a breaking-change app update.
+
+## How to deprecate a tool
+
+Use the helpers in `src/dbt_mcp/tools/deprecation.py`:
+
+```python
+from dbt_mcp.tools.deprecation import deprecated_description, deprecation_meta
+
+@dbt_mcp_tool(
+    description=deprecated_description(
+        get_prompt("discovery/get_model_parents"), replacement="get_lineage"
+    ),
+    meta=deprecation_meta(replacement="get_lineage"),
+    title="Get Model Parents",
+    read_only_hint=True,
+    destructive_hint=False,
+    idempotent_hint=True,
+)
+async def get_model_parents(...):
+    ...
+```
+
+`deprecated_description` prepends a standard banner:
+
+> **DEPRECATED — use \`get_lineage\` instead.** This tool will be removed in a
+> future release.
+
+`deprecation_meta` returns `{"deprecated": True, "replacement": "<name>"}`, which
+surfaces as `Tool._meta` in the MCP protocol so clients can inspect it
+programmatically.
+
+### What to change
+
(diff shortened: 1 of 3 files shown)


b-per (commented):
The in-repo steps look accurate. The doc stops at the `dbt-mcp` boundary though: it treats "submit a new app version after merge" as the end, but merging a dbt-mcp PR doesn't change the live app. A "Release & rollout" section is missing.

The published apps connect to the hosted MCP service, which depends on a released `dbt-mcp` version. The chain has steps the doc doesn't mention:

1. Land the dbt-mcp PR (code + snapshot + changelog) — covered.
2. Cut a dbt-mcp release. Merging is not releasing.
3. Bump the dbt-mcp dependency in the hosted service and deploy. This is what changes the surface the apps see.
4. Resubmit the app version in each store (OpenAI and Claude). Each store is a separate submission.

Step 4 has to come after step 3: the store captures the live server's tool list at submission, so submitting before the hosted service runs the new release captures the old surface.

Removal phase: the tool disappears from the live server when the hosted service deploys the removal, but the cached contract still lists it until the new app version is approved. So usage needs to be ~0 before that deploy, not just before the removal PR merges. One sentence in Phase B would cover it.

Suggested section:

> ## Release & rollout
> A merged dbt-mcp PR doesn't reach the published apps on its own. For both the deprecate and remove phases:
> 1. Merge the dbt-mcp PR (code + snapshot + changelog).
> 2. Cut a dbt-mcp release.
> 3. Bump the dbt-mcp dependency in the hosted MCP service and deploy. A major bump (a removal) needs the version pin updated there, not just a lockfile refresh.
> 4. Resubmit the app version in each store (OpenAI and Claude), after step 3 — the store captures the live server's tool list at submission.

theyostalservice:
@b-per Good catch — pushed `7a152c4` with a **Release & rollout** section (using your 4-step chain, worded generically since this doc is public/OSS) plus pointers to it from:

- the deprecation checklist's last item (was "submit a new app version after merge")
- the Phase B removal checklist's last item, with a sentence tying the ~0-usage bar to the *deployed* hosted service rather than the merged PR

_Comment posted by an AI assistant on behalf of @theyostalservice._

docs/deprecating-tools.md
```diff
@@ -0,0 +1,146 @@
...
+See `CONTRIBUTING.md` (§ Published-app contract) and
+`src/dbt_mcp/contract/snapshot.py` for technical details.
+
+## The deprecate-then-remove lifecycle
+
+1. **Add the replacement tool.** Ship the new, consolidated tool. Its contract
+   snapshot is submitted with the next app-store update.
+2. **Deprecate the old tool.** Keep it registered and callable. Apply a deprecation
+   banner to its description and a `deprecated` / `replacement` signal to its `meta`.
+   Publish a compatible app update.
```
jairus-m, line 26: From [MCP app update brainstorm](https://app.notion.com/p/dbtlabs/MCP-app-update-brainstorm-389bb38ebda78050872ff0401955cb12?source=copy_link)
Image

is it also worth implementing the above? That descriptions can be trimmed to reduce the likelihood of being used? In addition to the banner itself.

I also like the idea of forwarding the `{deprecated_tool}` as an alias to the `{new_tool}` given that the input/output schema of the tool can remain the same.

This comment also applies to #824 and #825 i believe.

theyostalservice: Good ideas — took both, with one adjustment.

**Trim the description:** done. `deprecated_description()` now *replaces* the description with a short, blunt line instead of prepending a banner to the full original prompt — matches the brainstorm doc's "lead with the deprecation and the replacement, and trim the rest." Applied it here and rippled it to both stacked PRs: #824 (`get_model_parents`/`get_model_children`) and #825 (the 8 `get_*_details` tools).

**Forward the deprecated tool as an alias:** captured as an *optional* step in the doc rather than something we force everywhere, because it isn't safe as a blanket rule:
- #825's 8 detail tools already call the same `ResourceDetailsFetcher.fetch_details` dispatcher as `get_details` — they're effectively aliased already, nothing to change.
- #824's `get_model_parents`/`get_model_children` aren't a drop-in for `get_lineage`: different fetcher, `get_lineage` requires `unique_id` where parents/children accept `name` alone, and the output is graph/depth-shaped rather than a flat list. Forwarding those would mean resolving name→unique_id and reshaping output during the soak window — a real behavior change for tools that are supposed to stay byte-for-byte stable until removal. Left those as direct implementations and documented the tradeoff.

Pushed `b18a114` on #826 with the doc updates, plus the helper + call-site changes on #824 (`1188a9e`) and #825 (`1da4b3b`).

_Comment posted by an AI assistant on behalf of @theyostalservice._

jairus-m (approved):

PR #828, opened 2026-06-30
fix: return DimensionValuesError when dimension column absent from Arrow result

### LINKED ISSUE #827: get_dimension_values raises unhandled KeyError when dimension column absent from Arrow result
## Bug

`get_dimension_values` in the Semantic Layer fetcher crashes with an unhandled `KeyError` when the Arrow table returned by the SDK does not contain a column matching the requested dimension name.

## Root cause

`client.py` lines 421–428:

```python
schema_names = raw_table.schema.names
column_name = next(
    (n for n in schema_names if n.lower() == dimension.lower()),
    dimension,   # fallback when no match found
)
raw: list[str] = [
    str(v)
    for v in raw_table.column(column_name).to_pylist()   # KeyError here
```

When no column matches case-insensitively, `next()` falls back to the raw `dimension` string. If that name is not in the schema, `raw_table.column(column_name)` raises `KeyError`. The `except` block only catches `QueryFailedError`, so the error propagates unhandled.

## Observed error (prod-us, 2026-06-29)

```
builtins.KeyError: 'Field "project_activity__date_day" does not exist in schema'
Location: pyarrow/table.pxi
Service: ai-codegen-api
```

## Expected behavior

`get_dimension_values` should return a `DimensionValuesError` with a clear message instead of raising.

## Fix

Pre-check whether the resolved column name is actually present in the schema and return `DimensionValuesError` if not, rather than falling back to a value that will cause `KeyError`.

b-per:
Closes #827

## Summary

- `get_dimension_values` crashed with an unhandled `KeyError` when the Arrow table returned by the Semantic Layer SDK contained no column matching the requested dimension name (not even case-insensitively).
- The `next()` fallback in `client.py` returned the raw dimension string, and `raw_table.column(that_string)` raised `KeyError`. The `except` block only caught `QueryFailedError`, so the error propagated unhandled to prod.
- Fix: change the `next()` fallback to `None` and return `DimensionValuesError` with a clear message when no matching column is found.

## Test plan

- [ ] New test `TestGetDimensionValues::test_dimension_column_absent_from_result_returns_error` reproduces the exact prod error and passes after the fix
- [ ] Full unit suite passes (`744 passed`)
- [ ] `task check` clean (ruff, mypy, pre-commit)

### DIFF
--- a/.changes/unreleased/Bug Fix-20260630-104909.yaml
+++ b/.changes/unreleased/Bug Fix-20260630-104909.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Fix unhandled KeyError in get_dimension_values when the Arrow result schema does not contain the requested dimension column
+time: 2026-06-30T10:49:09.977022+02:00
--- a/src/dbt_mcp/semantic_layer/client.py
+++ b/src/dbt_mcp/semantic_layer/client.py
@@ -425,7 +425,8 @@ class SemanticLayerFetcher:
             )
             if column_name is None:
                 return DimensionValuesError(
-                    error=f"Dimension '{dimension}' not found in result schema"
+                    error=f"Dimension '{dimension}' not found in result schema. "
+                    f"Available columns: {schema_names}"
                 )
             raw: list[str] = [
                 str(v)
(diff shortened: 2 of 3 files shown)


b-per:
Fixed in 5fdaa84 — error message now includes the available schema columns (addresses Copilot's inline suggestion).

wiggzz (approved):

PR #830, opened 2026-06-30
fix: add hatch-vcs fallback-version for mcpb installs without git history

### LINKED ISSUE #829: fix: mcpb install fails when no git history or tags present
## Problem

When installing dbt-mcp via a local `.mcpb` extension, the build fails with:

```
LookupError: Error getting the version from source `vcs`: setuptools-scm was unable to detect version
```

The `local.mcpb` installer copies source files to `Claude Extensions/` without the `.git` folder. Since `pyproject.toml` uses `hatch-vcs` (`source = "vcs"`) to derive the package version from git tags, the build fails in any environment without git history — including clean mcpb installs with no prior cached build artifacts.

Users who had a previous install (with leftover `egg-info` or `_version.py` artifacts) don't hit this because those files satisfy the version lookup without git.

## Fix

Add `fallback-version` to the hatch version config:

```toml
[tool.hatch.version]
source = "vcs"
fallback-version = "0.0.0"
```

This allows the build to succeed when VCS version detection fails, without affecting normal builds from a tagged git repo.

b-per:
## Summary

Closes #829.

Local `.mcpb` installs copy source files to `Claude Extensions/` without the `.git` folder. Since the project uses `hatch-vcs` to derive its version from git tags, the build fails in any environment where git history is absent — including clean mcpb installs with no prior cached build artifacts.

Adding `fallback-version = "0.0.0"` to `[tool.hatch.version]` lets the build succeed when VCS detection fails. Normal builds from a tagged git repo are unaffected.

### DIFF
--- a/.changes/unreleased/Bug Fix-20260630-183628.yaml
+++ b/.changes/unreleased/Bug Fix-20260630-183628.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Add hatch-vcs fallback-version to fix mcpb install failures in environments without git history
+time: 2026-06-30T18:36:28.827834+02:00
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,6 +81,7 @@ include = ["src/dbt_mcp/**/*", "README.md", "LICENSE"]
 
 [tool.hatch.version]
 source = "vcs"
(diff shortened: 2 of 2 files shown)


jairus-m (approved):
🙌

PR #835, opened 2026-07-07
Reduce semantic layer GraphQL request timeout to 30s

alan-andrade:
# Why

The semantic layer GraphQL request timeout is currently 60s. When the semantic layer server is slow or unresponsive, a 60s read timeout keeps each request (`list_metrics`, `query_metrics`, `get_dimensions`, etc.) hanging for a full minute before failing. Under sustained high latency this holds connections open far longer than is useful and adds load without returning results.

# What

Lower `SEMANTIC_LAYER_GQL_TIMEOUT` from 60s to 30s so unresponsive requests fail faster. 30s is the value #588 introduced to fix cold-cache `query_metrics` ReadTimeouts, so this preserves that fix while dropping the extra 30s of hanging on stuck requests.

---
Drafted by Claude Opus 4.8 under the direction of @alan-andrade


jairus-m (approved):
🙏

wiggzz (approved):

PR #849, opened 2026-07-20
[mcp] DI-4795: Capture tool-call token/char estimates in local MCP telemetry

zmandell:
## Why

Product wants to quantify the tokens processed by dbt-mcp. both per call and per session.

Today the `ToolCalled` event captures tool arguments but nothing about payload size or the resulting response. This blocks cost attribution and usage analysis across hosted and local MCP surfaces.

## What

Adds token/char estimates for tool-call inputs and outputs on the local MCP emitter, using the six new `ToolCalled` proto fields (proto#667).

- New `token_estimation.py` with a dependency-free char/4 heuristic (no tokenizer added to the OSS package). Exact char counts are the durable primitive; token estimates are directional.
- `llm_model` left blank — local MCP does not know the client model.
- Estimation is fully defensive: never raises, so telemetry cannot break a tool call.

## Stack

| PR | Status |
|----|--------|
| proto#667 | Merged |
| ai-codegen-api#1264 | Merged |
| **This PR** | Ready for review |
| vortex-event-analytics#113 | Approved |
| internal-analytics#4615 | Ready for review |

### DIFF
--- a/src/dbt_mcp/tracking/token_estimation.py
+++ b/src/dbt_mcp/tracking/token_estimation.py
@@ -1,0 +1,79 @@
+"""Lightweight token/char estimation for MCP tool-call telemetry.
+
+Local dbt-mcp deliberately avoids a tokenizer dependency: it estimates tokens
+with a simple ~4-characters-per-token heuristic. The exact character counts are
+the durable, tokenizer-independent primitive; the token estimate is directional.
+
+Estimation must never break a tool call, so serialization is defensive and falls
+back to a zeroed measurement on any error.
+"""
+
+import json
+import logging
+from collections.abc import Mapping, Sequence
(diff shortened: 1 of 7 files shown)


src/dbt_mcp/tracking/token_estimation.py
```diff
@@ -0,0 +1,79 @@
+"""Lightweight token/char estimation for MCP tool-call telemetry.
+
+Local dbt-mcp deliberately avoids a tokenizer dependency: it estimates tokens
```
b-per, line 3: I don't think that there is a hard rule to not depend on a tokenizer but to my knowledge there just isn't any existing library that supports estimating tokens for LLMs across provider and models.

b-per (approved):

Issue #413, opened 2025-10-28
[Feature]: Search for token-inefficient tools


DevonFulcher:
### Is your feature request related to a problem?

Currently, there are many tools that produce results that are very large and can exceed what is acceptable to pass to an LLM. We should implement a common search method for these tools so the LLM can search over the results rather than retrieve the full list.

### Describe the solution you'd like

Implement search over token-heavy tools with DuckDB. These tools are especially token-inefficient. There may be others:
- `dbt ls`
- `dbt run`
- `dbt test`
- `dbt build`
- `get_mart_models`
- `get_all_models`
- `get_exposures`
- `list_metrics`
- `get_dimensions`
- `get_entities`
- `get_job_run_artifacts`

### Describe alternatives you've considered

I looked into a few options for implementing this:
- Python libraries: there doesn't seem to be one library that fully fits our needs, so we would need multiple, making the solution a bit more complex.
- ChromaDB: Doesn't have great batch insertion functionality: https://cookbook.chromadb.dev/strategies/batching/#creating-batches. Doesn't have great hybrid search functionality: https://github.com/chroma-core/chroma/issues/1330.
- LanceDB: Has in-memory functionality, but it isn't well supported: https://github.com/lancedb/lancedb/issues/2247. May have slower insertions than a purpose-built in-memory db.
- DuckDB: Has good hybrid search capabilities: https://motherduck.com/blog/search-using-duckdb-part-3/

b-per:
Should this be implemented in the MCP server or on top of our platform APIs (leaving the CLI tools aside).

I don't think that `dbt run`, `dbt test` and `dbt build` are token-inefficient. From what I recall, we just return an OK or the error message.

DevonFulcher:
This could be implemented within dbt Platform. However, the tools span many APIs and teams. The pragmatic solution may be to do it in this project, especially for CLI tools (it seems that is just `dbt ls`), as you mentioned.

jairus-m:
Super random and not sure how appropriate "meta-convos" are in this technical thread (can move this elsewhere!) but just thought to share an experience (and also want to follow this issue in general).

[...]

PR #852, opened 2026-07-30
Re-add `get_job_run_artifacts` with jq-first large-artifact handling

jairus-m:
## Summary

Adds `get_job_run_artifacts` — fetches a specific artifact file from a dbt job run. Output mode depends on size and whether a `jq_filter` is provided:

- **Small, no filter** — returned inline
- **Large, no filter** — returns a guidance string with suggested `jq_filter` starting points; no file written
- **`jq_filter` set** — filter applied server-side; compact JSON array returned regardless of size

### Why jq over a filesystem path

Server-side paths aren't transport-portable (SSE/HTTP deploys point at the server's filesystem, not the agent's) and aren't useful even locally — a multi-hundred MB manifest can't be loaded into context regardless. jq targets and shrinks. A `resource_link`-based approach is the right long-term successor but doesn't solve the payload problem; left as deferred future work.

## What Changed

- `get_job_run_artifacts` with params `artifact_path`, `step`, and `jq_filter`
- Large unfiltered artifacts return a guidance string (suggested filters: `keys`, `.metadata`, `.nodes | keys`, `.results[] | select(...)`) — no disk writes
- `jq` runs in a `spawn`ed subprocess: required for env isolation and to avoid blocking the event loop on large JSON loads
- `env`/`$ENV` builtins neutralized by clearing `os.environ` in the worker — replaces the previous regex block that falsely rejected `.env` field access (e.g. `.metadata.env`)
- Filter capped at 120s; filtered output >500 KB raises instead of truncating
- Annotations corrected: `read_only_hint=True`, `destructive_hint=False`, `idempotent_hint=True`
- `jq` added as a dependency (pre-built wheels; no system install)
- Prompt notes performance distinction: `keys`/`.metadata` are O(1); filters iterating `.nodes` values scan the full manifest

## Why

`get_job_run_artifact` was [removed in #725](https://github.com/dbt-labs/dbt-mcp/pull/725) for token-inefficiency. This reinstates it with a jq-first design — large payloads stay off the wire unless the caller explicitly scopes them.

## Validation

End-to-end validated against a real dbt Platform job:

| What | Result |
|---|---|
| Small artifact, no filter | Inline content ✓ |
| Large `manifest.json`, no filter | Guidance string; no file written ✓ |
| `jq_filter=".metadata"` | `[{...}]` inline ✓ |
| `jq_filter=".metadata.env"` (previously false-blocked) | Returns manifest's captured env object ✓ |
| `jq_filter="keys"` | Top-level keys array inline ✓ |
| Aggregate filter (`... \| length`) | Small result inline ✓ |
| `jq_filter="env"` | `[{}]` — no secret leak ✓ |
| `jq_filter='$ENV.SECRET // "gone"'` | `["gone"]` ✓ |
| Oversized filtered output | `ValueError: Filtered output exceeds 500 KB` ✓ |
| Invalid jq syntax | `"Invalid jq filter: …"` ✓ |
| Nonexistent run ID | `NotFoundError`, no URL leaked ✓ |
| Step targeting | Correct artifact / `NotFoundError` / `NotFoundError` ✓ |

> Concurrent jq calls against the same large manifest can cause memory pressure (artifact pickled per call). Sequential calls are safe.

## Related Issues

Closes #
Related to #413

## Checklist
- [ ] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required

This PR was drafted with AI-assistance 🤖 (Claude Code) !


b-per (commented):
Nice to see `get_job_run_artifacts` back with jq filtering and size-based offloading. A few things worth addressing before merge, roughly in priority order:

## Security

**A. `jq_filter` can read the server's environment variables.** jq's `env`/`$ENV` builtins aren't filtered out, so a filter like `"$ENV"` or `"$ENV.SOME_VAR"` returns the dbt-mcp server process's actual environment — independent of the artifact being fetched. I verified this directly against the library call this PR uses:
```python
>>> jq.compile("$ENV.MY_SECRET_TOKEN").input(None).first()
'super-secret-value'
```
Since dbt-mcp's own credentials (API tokens, host config) are typically env-configured, this is a real exfiltration path, not theoretical. Suggest stripping/blocking `env`/`$ENV` if the library supports a restricted mode, or running the filter against a sanitized environment, plus a regression test asserting `$ENV` isn't reachable.

## Correctness / reliability

**B. No size cap on `jq_filter` output.** The no-filter path enforces `INLINE_CONTENT_LIMIT` (500 KB), but `jq_filter` returns `json.dumps(results)` uncapped. The PR's own validation table shows an enumeration filter on a 9,027-node manifest produced 543K chars — only caught because the Claude Code harness intercepted the oversized tool result and wrote its own temp file. On an MCP host without that safety net, this has no graceful fallback. Suggest applying the same 500 KB check to filtered output, erroring with guidance to narrow the filter (aggregate instead of enumerate).

**C. No timeout on jq execution; blocks the event loop.** `jq.compile(jq_filter).input(parsed).all()` runs synchronously inside an `async def` with no `asyncio.to_thread` and no timeout, unlike the `output_path` write a few lines below it, which is thread-offloaded. A pathological filter (unbounded `reduce`/`while`) can hang the event loop indefinitely, affecting concurrent requests beyond just this call. Suggest wrapping in `asyncio.to_thread` with a timeout.

**D. Auto-temp-file / `output_path` don't account for remote transport.** Both the size-triggered auto-temp-file and the explicit `output_path` write happen unconditionally regardless of `MCP_TRANSPORT`. For `stdio` this is safe (guaranteed local subprocess), but for `sse`/`streamable-http` the resulting path may be unreachable by the client — the tool just returns "Artifact written to {path}" as if it succeeded, with no indication anything is wrong beyond a prose warning in the `output_path` param description (which the auto-temp-file default path doesn't even carry). Suggest checking transport server-side and raising a clear error directing the caller to `jq_filter` instead, rather than relying on the model having read the docs.

## Performance

**H. No caching — every jq_filter iteration re-downloads the full artifact.** `get_job_run_artifact` in `client.py` does a fresh `httpx` GET on every call, with no in-memory or disk cache, and no conditional-request handling (ETag/If-None-Match). Since a completed run's artifacts are immutable, and this tool is explicitly designed for iterative jq_filter refinement (per the prompt doc's own guidance), each refinement re-downloads the entire artifact — costly for a multi-hundred-MB manifest. Suggest caching by `(account_id, run_id, artifact_path, step)`, either in-memory (bounded, since content can be large) for the life of the server process, or reusing the auto-written temp file as the cache when one already exists for that key.

## Cleanup

**E. Auto-written temp files are never deleted.** `_write_to_temp_file` uses `NamedTemporaryFile(delete=False, ...)`, so every size-triggered auto-write leaves a file behind permanently. Repeated large-artifact fetches (e.g. iterating on filters against a large manifest) will accumulate uncapped disk usage on the server host. Consider a TTL/cleanup sweep, or at least documenting that callers own cleanup.

**F. Stale orphaned prompt doc left in place.** `src/dbt_mcp/prompts/admin_api/get_job_run_artifact.md` (singular) is a leftover from the pre-#725-removal tool and isn't wired to anything anymore — it now sits right next to the new `get_job_run_artifacts.md` (plural), which is confusing. Suggest deleting it in this PR.

## Token efficiency

**G. Trim or relocate the jq filter reference.** The new prompt doc is ~1,473 tokens (cl100k_base), loaded into context whenever this tool is enabled, regardless of whether JSON filtering is ever used. Consider trimming the worked examples down to artifact names + top-level key pointers, or moving the detailed filter reference to an MCP resource the model reads on demand instead of shipping it inline in every tool description.

jairus-m:
@b-per as per [1ddc9f0](https://github.com/dbt-labs/dbt-mcp/pull/852/commits/1ddc9f06fd01a531ba943808068fdf3c686425e9), some comments were addressed and others deferred:
- Implemented: env filtration, cap jq output, jq blocking event loop / no timeout, remove stale tool markdown, trim prompt/description
- Deferred: deletion of temp files, remote transport behavior, no caching of jq results

The deferred can be for future work/PRs, still pending on alignment with team on this tool in the meantime.

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -273,8 +273,8 @@ async def list_job_run_artifacts(
 @dbt_mcp_tool(
     description=get_prompt("admin_api/get_job_run_artifacts"),
     title="Get Job Run Artifacts",
-    read_only_hint=True,
-    destructive_hint=False,
+    read_only_hint=False,
+    destructive_hint=True,
```
b-per: Why did we change those?

jairus-m: The [idea](https://github.com/dbt-labs/dbt-mcp/pull/852#discussion_r3684709814) was that this tool interacts with a user's file system and can write files and potentially overwrite files. It's not destructive to objects in dbt Platform tho depending on how we define these things.

... realizing that this gets a bit complicated as for remote MCP, only using `jq` filtering to return pointed results without writing a tmp file can be confusing. 

In this case, I think it makes sense to have this as the umbrella default?

b-per: Can we not allow people to define a path/filename and save it to a [temp](https://docs.python.org/3/library/tempfile.html) one instead?

If there is no real need for a tool to be destructive and not read-only, I think that we should avoid it.

jairus-m: I do think that's possible. I only think: what about the case where someone wants to write and keep the artifact for specific usage downstream? This may be an edge case, but I do think it's valuable to be able to have that flexibility to write it to a specific location and keep it there.

jairus-m: @b-per i made updates in the last few commits that addresses the above:
1. Write deterministic files names to `~/.dbt/artifacts/` 
    - No user option to control output path
3. Include tmp file cleanup
4. Destructive annotation updated:
    - read_only = false 
    - destructive = false <-- this is updated
    - idempotent = true

Read only is still technically true - thoughts?

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -248,6 +275,82 @@ async def list_job_run_artifacts(
...
+        )
+    admin_api_config = await context.admin_api_config_provider.get_config()
+    content = await context.admin_client.get_job_run_artifact(
+        admin_api_config.account_id, run_id, artifact_path, step=step
+    )
+    if output_path is not None:
+        try:
+            await asyncio.to_thread(
+                Path(output_path).write_text, content, encoding="utf-8"
+            )
```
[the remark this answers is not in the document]
jairus-m: this was addressed ! Files written to a `~/.dbt/artifacts` w/ deterministic names

b-per (approved):

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -254,6 +279,70 @@ async def list_job_run_artifacts(
...
+                        "artifacts are slow to traverse — prefer a structural or "
+                        "aggregation filter ('keys', '.metadata', '... | length') over "
+                        "one that enumerates every node, or target a specific step."
+                    )
+                except json.JSONDecodeError:
+                    raise ValueError(
+                        "jq_filter requires a JSON artifact; this artifact is not valid JSON"
+                    )
+                except Exception as e:
+                    raise ValueError(f"Invalid jq filter: {e}") from e
```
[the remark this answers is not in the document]
b-per: Checked this against CPython's `multiprocessing/pool.py` source directly rather than just the docs — I don't think there's a leak here.

`Pool.terminate()` calls `self._terminate()`, which is the registered `_terminate_pool` static method. That function already terminates any live workers *and* explicitly `p.join()`s every worker process as its last step before returning — it's synchronous and blocks until all workers are reaped. There's no missing join to add.

The `with ctx.Pool(...) as pool:` context manager's `__exit__` also just calls `self.terminate()`. On the timeout path we call `pool.terminate()` explicitly and then the `with` block calls it again on the way out — that's safe because `terminate()` is backed by a `multiprocessing.util.Finalize` object, which guards against double invocation (it deletes itself from an internal registry on first call and no-ops on any subsequent call).

Since each pool here is single-use (`processes=1`, scoped to one call) and by the time we reach cleanup `future.get()` has already returned or raised, there's no in-flight work that would benefit from a graceful `close()` instead of `terminate()` either way.

Happy to revisit if there's a concrete repro of a lingering process, but based on the source I don't think `close()/join()` changes the actual behavior here.

PR #854, opened 2026-07-30
fix: correct cancel_job_run tool annotations (destructive, idempotent)

alan-andrade:
# Why

`cancel_job_run` was annotated with `destructiveHint: false` and `idempotentHint: false`, underreporting its effects. Per the MCP spec, cancelling a run is destructive (it stops something running, as opposed to additive), and idempotent (re-cancelling is a no-op).

`trigger_job_run` and `retry_job_run` are additive, not destructive, so they remain `destructiveHint: false`.

# What

- Mark `cancel_job_run` as destructive and idempotent.
- Update test to assert per-tool annotation differences.
- Regenerate contract snapshot.

Drafted by gpt-5.6-terra and Claude Opus 4.6 under the direction of @alan-andrade.

### DIFF
--- a/src/dbt_mcp/dbt_admin/tools.py
+++ b/src/dbt_mcp/dbt_admin/tools.py
@@ -110,7 +110,7 @@ async def get_job_details(
     description=get_prompt("admin_api/trigger_job_run"),
     title="Trigger Job Run",
     read_only_hint=False,
-    destructive_hint=False,
+    destructive_hint=True,
     idempotent_hint=False,
 )
 async def trigger_job_run(
@@ -200,7 +200,7 @@ async def get_job_run_details(
     description=get_prompt("admin_api/cancel_job_run"),
     title="Cancel Job Run",
     read_only_hint=False,
-    destructive_hint=False,
+    destructive_hint=True,
     idempotent_hint=False,
 )
 async def cancel_job_run(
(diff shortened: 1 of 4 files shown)


src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -200,7 +200,7 @@ async def get_job_run_details(
     description=get_prompt("admin_api/cancel_job_run"),
     title="Cancel Job Run",
     read_only_hint=False,
-    destructive_hint=False,
+    destructive_hint=True,
     idempotent_hint=False,
```
b-per: If we are looking at hints, shouldn't `idempotent_hint` be `True` for `cancel_job_run`? If a run is cancelled, cancelling it again won't trigger any side effect.

src/dbt_mcp/dbt_admin/tools.py
```diff
@@ -110,7 +110,7 @@ async def get_job_details(
     description=get_prompt("admin_api/trigger_job_run"),
     title="Trigger Job Run",
     read_only_hint=False,
-    destructive_hint=False,
+    destructive_hint=True,
```
b-per: What is the "real" definition of destructive?

Checking the MCP docs it says
> destructiveHint: If it does modify things, is the change destructive (as opposed to additive)?

To me, triggering a job run is not destructive then, it is additive
Same for `retry_job_run`
But `cancel_job_run` could somehow be considered destructive, yes (even if it is debatable). In the end we don't "destroy" anything (not in the dbt platform or in the user data warehouse), but yes we "stop"/"cancel" something running.

b-per (approved):

PR #870, opened 2026-08-28
Fix QueryFailedError message formatting and classify warehouse credential errors

theyostalservice:
# Why

The semantic layer can now classify certain warehouse query failures (e.g. bad credentials vs. a permission/object-not-found error) and mark that classification in the error message it returns. Today, `query_metrics` and related tools surface these messages to the calling agent/model without distinguishing "your query was wrong" from "your warehouse credentials or access are the problem" — a distinction that matters a lot for what the agent should try next.

Separately, the existing error-cleanup path had a latent formatting bug: `QueryFailedError.__str__` (from the `dbt-sl-sdk` dependency) wraps its message in a `message="...", status=...` artifact, and the cleanup chain that's supposed to strip noise from these messages was operating on that mangled `str()` output instead of the exception's clean `.message` attribute. This produced garbled error text for callers, and it also stood in the way of reliably detecting a warehouse classification marker at the start of the message.

# What

- `SemanticLayerFetcher._format_semantic_layer_error` now reads `QueryFailedError.message` directly instead of `str(error)`, so the existing cleanup chain operates on the real underlying message rather than `__str__`'s wrapper artifact.
- Added a small classifier (`dbt_mcp.errors.hints.classify_warehouse_error`) that detects a warehouse authentication or permission marker embedded in an error message, strips it, and reports a generic category. This runs before the existing cosmetic cleanup so the marker can't be corrupted by cleanup steps that assume no leading bracket.
- When a message is classified, an actionable hint is appended: one pointing the caller at reconfiguring credentials for authentication failures, and one noting that a permission failure can also mean the referenced object doesn't exist or is misspelled (since the warehouse reports both cases identically) for permission failures.
- Messages without a marker — the overwhelmingly common case — are unaffected.

No changes to `QueryMetricsError`'s shape were needed; this ships as a pure message-text change.

# Notes

`ui/` lint (`pnpm run lint`) wasn't run locally — this environment's asdf-pinned Node version didn't match what's installed — but no files under `ui/` were touched by this change. `ruff check`, `ruff format`, and `mypy` all pass via `pre-commit run --all-files`, and `task test:unit` passes (807 passed, plus 3 new tests; the 10 pre-existing failures are unrelated to this change and reproduce identically on a clean `origin/main` checkout).

---

Drafted by Claude Sonnet 5 under the direction of @theyostalservice

### DIFF
--- a/.changes/unreleased/Bug Fix-20260828-120252.yaml
+++ b/.changes/unreleased/Bug Fix-20260828-120252.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: Fix semantic layer query error formatting to use QueryFailedError's clean underlying message instead of its str() representation, which was leaking a message="...", status= wrapper artifact into the error shown to callers
+time: 2026-08-28T12:02:52.820897-07:00
--- a/.changes/unreleased/Enhancement or New Feature-20260828-120257.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20260828-120257.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: Classify warehouse authentication and permission errors from semantic layer query failures, stripping the internal classification marker and appending an actionable hint pointing the caller toward the likely fix
+time: 2026-08-28T12:02:57.19693-07:00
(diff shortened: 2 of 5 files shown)


jairus-m (approved):

theyostalservice:
I just updated the whole stack of commits to be signed commits to meet repo requirements.

PR #871, opened 2026-08-31
feat: expose dbt LSP capabilities through MCP

dataders:
## Summary
Adds a persistent local dbt LSP connection and MCP facade for compiler-backed navigation, diagnostics, lineage, SQL previews, safe refactor previews, and incremental in-memory document updates.

## What Changed
- Owns the local dbt LSP as a stdio child process with lazy startup, request correlation, server-request handling, and graceful shutdown.
- Adds MCP tools for project status, diagnostics, nodes, model lineage, definitions, references, compilation, SQL previews, code-action previews, rename previews, and incremental document updates.
- Automatically compiles before compiler-backed queries and tracks document versions across didOpen/didChange notifications.
- Keeps code actions and rename operations preview-only; edits are returned for review and never applied by MCP.
- Adds tool prompts, generated documentation, changelog coverage, and focused LSP tests.

## Why
A persistent compiler-backed workspace lets an agent inspect and refine a dbt project incrementally instead of reaching for a full dbt CLI invocation for every question or edit.

## Testing
- `uv run pytest tests/ --ignore=tests/integration -x -q` (825 passed)
- `uv run pytest tests/unit/lsp -q` (123 passed)
- `uv run pre-commit run --all-files`
- `uv run mypy --show-error-codes --namespace-packages . --exclude examples/`
- Live stdio smoke test with the fs `codex/lsp-code-actions` dbt binary covered initialization, compilation, incremental edits, diagnostics, and formatter code-action previews.

## Notes
No shared semantic service is introduced in this change. The existing column-lineage path remains available; this PR focuses on making the LSP a reusable agent-facing compiler and workspace interface.


dataders (commented):
Adversarial review found three issues:

1. **[P1] Constrain paths to `DBT_PROJECT_DIR`.** `LSPClient._path()` resolves `../...` and absolute paths without enforcing containment. Every new path-based MCP tool therefore accepts files outside the configured project, despite the schema describing project-relative paths; `compile_file()` can additionally return file contents. Reject any resolved path where `candidate.relative_to(self.project_dir)` fails.

2. **[P2] Mark read-only tools accurately.** `make_definition()` defaults `read_only` to false, and project status, diagnostics, node, lineage, definition, references, compile preview, code actions, and rename preview never override it. MCP hosts will treat these non-mutating operations as writes, potentially adding approval prompts or excluding them in read-only sessions. Default the helper to true and override only `dbt_lsp_update_document` and any operation intentionally considered stateful.

3. **[P2] Preserve legacy socket transport compatibility.** Binary discovery still supports custom and editor-installed legacy `dbt-lsp` binaries, but the provider now unconditionally launches every detected binary over stdio. Previously those binaries received `--socket`; no version/capability gate establishes stdio support. Older supported installations can start but fail initialization. Select transport by detected binary/version or retain a socket fallback.

CI snapshot during review: styling, unit, docs, app contract, manifest, changelog, and CODEOWNERS passed; integration was still pending.

dataders:
Addressed the review feedback in commit `944850b`:\n\n- Path-based LSP operations now resolve and enforce containment inside `DBT_PROJECT_DIR`, including symlink-resolved paths.\n- LSP MCP annotations now mark read-only operations accurately; project compilation and in-memory document updates remain non-read-only.\n- Binary detection now records transport explicitly: Fusion `dbt lsp` uses stdio, while legacy/editor-installed `dbt-lsp` binaries retain the socket transport.\n\nValidation: 828 non-integration tests passed, LSP tests passed, generated docs are current, and pre-commit plus mypy passed.

b-per:
Let's keep this as an experiment for now and see if/how we could merge it at some point

PR #876, opened 2026-09-08
Filter out dbt docs v2 Parquet artifacts from artifact tools

jairus-m:
## Summary
Filter dbt docs v2 Parquet artifacts from `list_job_run_artifacts` and add an
early guard in `get_job_run_artifacts` for these paths.

## What Changed
- `list_job_run_artifacts` now filters `index/` and `metadata/` in addition to
  `compiled/` and `run/`
- `get_job_run_artifacts` raises `ArtifactRetrievalError` before making any HTTP
  request when the path starts with `index/` or `metadata/`
- Added a PAR1 magic-byte backstop for unexpected binary Parquet outside those
  prefixes
- Updated comments to correctly describe these as dbt docs v2 / dbt Information
  Schema artifacts rather than internal Fusion artifacts

## Why
`index/` and `metadata/` paths contain dbt docs v2 Parquet files generated by
`--write-index` and `--write-metadata`. They are binary and cannot be returned
as text. Listing them leads to failed fetch attempts with no useful output. No
MCP tooling exists yet to query them, so they are excluded until that support
is added.

## Related Issues
Closes #
Related to #

## Checklist


src/dbt_mcp/dbt_admin/client.py
```diff
@@ -373,12 +373,18 @@ async def list_job_run_artifacts(self, account_id: int, run_id: int) -> list[str
...
+        # and cannot be returned as text.
         filtered_data = [
             artifact
             for artifact in data
             if (
-                not artifact.startswith("compiled/") and not artifact.startswith("run/")
+                not artifact.startswith("compiled/")
+                and not artifact.startswith("run/")
+                and not artifact.startswith("index/")
+                and not artifact.startswith("metadata/")
```
b-per, line 387: I was not even aware that we were already filtering out artifacts...

OK with this but we should revise at some point and/or make sure that the tool description mention that we return a filtered list and what is filtered out.

b-per (approved):
OK for now for me, but we should create follow-up issues/tickets to make sure that descriptions are aligned with what the tools do

src/dbt_mcp/dbt_admin/client.py
jairus-m: Updated - also, hopefully soon (when dbt docs v2 is closer or at GA), we bring back `/index` and `/metadata` so that `get_job_run_artifacts` can be extended with tools to query parquet artifacts

PR #878, opened 2026-09-08
Bound get_lineage result size

wiggzz:
## Why

Large lineage graphs can overwhelm MCP clients and consume excessive model context.

## What

Add an optional `limit` to `get_lineage`, defaulting to 100 nodes. Bounded results keep the closest nodes in breadth-first order, retain only valid edges, and include a structured `truncation` summary with the exact omitted node count, counts by resource type, and omitted immediate parent/child counts for bidirectional calls.

---
Drafted by GPT-5 Codex under the direction of @wiggzz.


src/dbt_mcp/discovery/tools.py
```diff
@@ -344,16 +349,22 @@ async def get_lineage(
...
         types=types,
         depth=depth,
         direction=direction,
+        limit=limit + 1,
         config=config,
     )
-    return build_lineage_graph(root_id=unique_id, nodes=nodes)
+    return build_lineage_graph(
+        root_id=unique_id,
+        nodes=nodes[:limit],
```
DevonFulcher: It seems this may be superfluous, and `_filter_connected_nodes()` maintains the limit, right? I think that is fine. I'm just trying to understand the code better.

src/dbt_mcp/prompts/discovery/get_lineage.md
```diff
@@ -20,6 +20,10 @@ Use `direction` to narrow the response to one side of the graph and reduce paylo
 - `direction="downstream"`: descendants only — excludes the target node and ancestors
 - `direction="both"` (default): the target node plus both ancestors and descendants
 
+Use `limit` to bound the number of returned nodes (default `100`). When the
+connected graph is larger, the closest nodes are returned first and `truncated`
+is `true` in the response.
+
```
jairus-m, line 29: *Edit: Testing this tool call on a real dbt project!

On a fact model in internal analytics with 13 direct parents and 90+ downstream nodes, calling with limit=14, depth=1, direction="both" returned:

- 14 parents
- 0 children
- truncated = true

All 13 parents fill the node limit before any children are able to be counted even tho direction="both" being explicitly requested. However, the same limit=14 with direction="downstream" is able to return 14 children.

Given this, it feels like returning truncated=true is misleading as it does say _something_ was cut, but not _what_ (like in this instance, all the kiddos). However this case is probably rare, esp with default limit=100 needing a model to have 99 parents at depth=1 to come accross.

But maybe it is worth noting in `get_lineage.md` that two separate "directional" calls give a more accurate picture when you have a model with lots of parents that exceed the limit param. OR generally, any large graph..!

wiggzz: Yea. I think this is pretty rare. We could note how many nodes and what kinds of nodes were cut since we have all of them.

src/dbt_mcp/discovery/tools.py
wiggzz: yea looks like we fetch one more than required in order to tell if there was any truncation.

src/dbt_mcp/prompts/discovery/get_lineage.md
wiggzz: Addressed in 48825e4. The boolean is now a structured `truncation` value with exact `omitted_node_count` and `omitted_resource_type_counts`; it is `null` when nothing was omitted. Because Discovery already returns the complete lineage payload, the BFS now finishes the connected walk and summarizes everything after the output limit without another API call. `task check` and all 861 unit tests pass.

_Posted by GPT-5 Codex under the direction of @wiggzz._

wiggzz: Follow-up in 47aba8f: I initially captured resource kinds but missed the core parent-first starvation signal. `truncation.omitted_direction_counts` now reports exact `upstream` and `downstream` counts. The regression reproduces the root + parents consuming the limit and asserts `{"upstream": 0, "downstream": 2}` for the omitted children. `task check` and all 861 unit tests pass.

_Posted by GPT-5 Codex under the direction of @wiggzz._

wiggzz: Simplified in d37de11 after another pass: the full directional traversal was more machinery than this signal warranted. The response now reports `omitted_immediate_node_counts: {parents, children}` for bidirectional calls, using direct set intersections with the root relationships. This still captures the parent-first starvation case while removing 27 lines of traversal logic. Full checks and all 861 unit tests pass.

_Posted by GPT-5 Codex under the direction of @wiggzz._

DevonFulcher (approved):

jairus-m:
Just curious: Since this is an additive change (tool parameter wise) and just changes the default behavior would a enhancement >> bug fix changie entry be more appropriate? 

Also, I made one recent fix in #880 that id like to release before tomorrow, 9/10 . Shall we add this PR too?

src/dbt_mcp/prompts/discovery/get_lineage.md
wiggzz: Coming back to this, I think I'm going to just move this to a simpler `omitted_node_count`.

PR #705, opened 2026-04-08
Auto-fetch DBT_HOST_PREFIX from dbt Platform for SL and Discovery API URL construction

jairus-m:
## Summary

Add step to fetch DBT_HOST_PREFIX from platform (for SL/Discovery API url construction) for env var config (similar to what happens with OAuth flow). This is purely additive and backwards compatible.

## Related Issues

Closes #583 

## Checklist
- [x] I have performed a self-review of my code
- [x] I have made corresponding changes to the documentation (in https://github.com/dbt-labs/docs.getdbt.com) if required -- Mention it here
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Validation:

Ran `task run` and inspected the setting config logs with the following settings:
```bash
DBT_HOST=https://<redacted-prefix>.us1.dbt.com
DBT_ACCOUNT_ID=
DBT_PROD_ENV_ID=
DBT_TOKEN=
DBT_PROJECT_DIR=
DBT_PATH=
```
On the main branch:
```bash
INFO [dbt_mcp.mcp.server] Multi-project mode disabled -> Env-var mode
INFO [dbt_mcp.mcp.server] Registering product docs tools
INFO [dbt_mcp.mcp.server] Registering MCP server tools
INFO [dbt_mcp.mcp.server] Registering semantic layer tools
INFO [dbt_mcp.mcp.server] Registering discovery tools
INFO [dbt_mcp.mcp.server] Registering dbt cli tools
INFO [dbt_mcp.mcp.server] Registering dbt codegen tools
INFO [dbt_mcp.mcp.server] Registering dbt admin API tools
INFO [dbt_mcp.mcp.server] Registering LSP tools
INFO [dbt_mcp.mcp.server] Starting MCP server
INFO [dbt_mcp.mcp.server] Registering proxied tools
INFO [httpx] HTTP Request: GET https://vu491.us1.dbt.com/api/v2/accounts/1/ "HTTP/1.1 200 OK"
INFO [dbt_mcp.config.credentials] Fetched prefix <PREFIX> from dbt Platform.
[...]

PR #880, opened 2026-09-09
Fix: Infer account prefix from DBT_HOST when platform API returns no subdomain fields

jairus-m:
## Summary

Fixes a DNS error on discovery and semantic layer tools discovered while working with a live customer project. Affects multi-cell accounts where `DBT_HOST` embeds the account prefix (ie `ab123.us1.dbt.com`) but the dbt Platform API returns null for both `static_subdomain` and `vanity_subdomain`.

## What Changed

- Added `_infer_prefix_from_host` helper in `credentials.py` to extract the account prefix from a 4-label `.dbt.com` host as a last-resort fallback when the platform fetch returns nothing.
- Added fallback inference in the env-var credentials path: if the auto-fetch returns nothing and `DBT_HOST` is a 4-label host, the prefix is inferred and a warning is logged. If the host is 3-label and no inference is possible, a separate observability warning is logged.

## Why

When the platform API returns null subdomain fields, the auto-fetch introduced in #705 returns `None` and the prefix is never set. The discovery URL is then built as `https://metadata.ab123.us1.dbt.com/graphql` instead of `https://ab123.metadata.us1.dbt.com/graphql`, producing `[Errno 8] nodename nor servname provided, or not known` on any discovery or semantic layer tool call. Admin API calls succeed (the admin URL uses `DBT_HOST` directly), which masks the misconfiguration until a discovery/SL tool is used.

Setting `DBT_HOST_PREFIX` explicitly continues to take precedence and remains the recommended override if inference is incorrect.

## Related Issues

Closes #
Related to #705

## Checklist

## Additional Notes

- Manual validation: set `DBT_HOST=<prefix>.us1.dbt.com` + `DBT_TOKEN` + `DBT_ACCOUNT_ID` with no `DBT_HOST_PREFIX`; startup logs should show `"Inferred prefix '<prefix>' from DBT_HOST"` and discovery tools should return results.
- Also, tested the "before" state - getting DNS error - and reconfirmed on the development branch with the "after" state which no longer led to the DNS error

PR created with AI-assistance 🤖 !


DevonFulcher (approved):
Very nice!

PR #885, opened 2026-09-10
[DI-5247] Fix get_job_run_artifacts returning 500 server errors instead of 400 client errors when jq filter fails

jairus-m:
## Summary

Convert generic `ValueError` to `InvalidParameterError` to classify `jq` tool call errors as a 400 client error instead of internal server errors (500) for callers.

## What Changed

- `ValueError` --> `InvalidParameterError`

## Why

MCP client errors hitting errors with `jq` filtering in `get_job_run_artifacts` lead to 500 internal server errors. These errors can be caused by:
- validation errors (not JSON file)
- invalid `jq` syntax
- timeouts
- memory exceed 500kb 
- etc

These are fundamentally client errors and should be classified as such as. Today, `jq` tool call errors are returned as generic `ValueError` and don't get classified as either a `ServerToolCallError` (500) OR `ClientToolCallError` (400), it defaults to a generic 500 since `ValueError` is unknown and not classified in `src/dbt_mcp/errors/__init__.py` .

## Related Issues

Closes #
Related to #

## Checklist

## Additional Notes

`InvalidParameterError` is an existing, generic, and cross-cutting error type. For the 4 cases above, this type fits well for using wrong artifact type for `jq` (validation error) and invalid `jq` syntax. However, for timeouts and memory exceptions, this is a bit of stretch. 

For now, keeping this generic bucket error but if folks prefer more specificity, pls do call it out!


jairus-m:
@copilot in regards to your last comment: https://github.com/dbt-labs/dbt-mcp/pull/885#pullrequestreview-5173981935

im not sure if that is worth addressing - 

1st, the core `jq` errors that would be regularly caught in the majority of this tool's use are returned as 400's as `InvalidParameterError`. this was the main issue and that's now addressed.

2nd, opening up the final exception to ALL exceptions and raising `InvalidParameterError` is way too broad and doesn't make sense to do. Because all of a sudden, everything would get raised as an "InvalidParameterError`. 

3rd, trying to enumerate all the failure modes and catching every single edge here is not sustainable and is a bit smelly

b-per (approved):

b-per:
The PR description talsk about

> memory exceed 500kb

But there is nothing in the PR about it, is it?

DevonFulcher (approved):
Thanks!

jairus-m:
@b-per 
>  - re: memory exceeds 500kb
> - But there is nothing in the PR about it, is it?

It is in the code. Lines 325 - 328. This is actually the exact error message that comes up in datadog that this Jira ticket is baed on. The 500kb constant, `INLINE_CONTENT_LIMIT`,  is defined at the top of the file

Screenshot 2026-09-11 at 7 56 27 AM

PR #892, opened 2026-09-21
[Bug]: OAuth: multi-project context never passes _is_context_complete(), forcing a full browser login on every server start

### LINKED ISSUE #856: [Bug]: OAuth: multi-project context never passes _is_context_complete(), forcing a full browser login on every server start
### What happened?

When the OAuth login is completed with a **project selection** (multi-project mode), the context written to `~/.dbt/mcp.yml` deliberately stores `prod_environment: null`. However, `_is_context_complete()` in `src/dbt_mcp/config/credentials.py` unconditionally requires `prod_environment`, so the cached context is rejected on every startup. The server then falls straight through to the full `login()` flow: the cached access token and the refresh token (granted via `offline_access` / "Keep session alive") are **never consulted**.

Practical impact: every start of the MCP server opens the browser authorize page plus the project picker again. With more than one MCP host on the machine (Claude Desktop, Claude Code CLI, VS Code), that is easily 10+ interactive logins per day, even though a valid token and refresh token sit in `mcp.yml` the whole time.

**Expected:** the cached context is accepted; the still-valid access token is used, or an expired one is silently renewed via the refresh token. No browser interaction on restart.

**Actual:** the browser authorize page and the project picker open again on every start.

## Root cause

`_is_context_complete()` requires `prod_environment`:

```python
return bool(
    dbt_ctx
    and dbt_ctx.account_id
    and dbt_ctx.host_prefix
    and dbt_ctx.prod_environment
    and dbt_ctx.decoded_access_token
)
```

But a multi-project login never sets it: since the fix "stored prod/dev environments are no longer inherited when selected_project_ids is set" (changelog), `prod_environment: null` is the *intended* on-disk state for multi-project contexts. The two pieces contradict each other, so multi-project contexts can never be reused. In `get_dbt_platform_context()` the incomplete context skips both the token-validity check and `_try_refresh_token()` and goes directly to `login()`.

This is the same class of bug as #547 ("make development environment optional for OAuth context reuse"), which relaxed the check for `dev_environment` but left `prod_environment` required.

## Suggested fix

Treat `prod_environment` as required only for single-project contexts, e.g.:

```python
and (dbt_ctx.prod_environment or dbt_ctx.selected_project_ids)
```

Multi-project mode is dispatched on `selected_project_ids` anyway, and `register_multi_project_dbt_mcp()` does not need `prod_environment`. I have been running this one-line patch locally; restarts now reuse the cached token and refresh silently, and the admin API tools work unchanged.

Happy to open a PR if that direction is acceptable.

### Steps to Reproduce

1. Configure the server with `DBT_HOST` only (no `DBT_TOKEN`) so the OAuth path is used, with a platform toolset enabled (e.g. admin API).
2. Start the server and complete the browser login. Tick "Keep session alive" and select one or more **specific projects** in the project picker.
3. Confirm `~/.dbt/mcp.yml` now contains `selected_project_ids: [...]`, a valid `decoded_access_token` including `refresh_token`, and `prod_environment: null`.
4. Restart the MCP server (or simply restart the MCP host).
5. The browser authorize page opens again instead of reusing the cached/refreshable token.

### Deployment

Local MCP server

### Environment

- OS: Windows 11
- dbt-mcp: 1.22.1 (bug also present on `main`), Python 3.13, stdio transport
- dbt CLI executable (for CLI tools): dbt Core (CLI toolset disabled in this setup; only admin API enabled)
- MCP client: Claude Desktop, Claude Code CLI, VS Code (all show the same behavior)

### Relevant log output

```shell
Sanitized ~/.dbt/mcp.yml after a completed multi-project login:

account_id: 12345
dbt_host: us1.dbt.com
decoded_access_token:
  access_token_response:
    access_token: '***'
    refresh_token: '***'    # present, "Keep session alive" was ticked
    expires_at: <valid, ~60 min>
    scope: user_access offline_access
    token_type: bearer
dev_environment: null
host_prefix: '***'
prod_environment: null      # <- intended for multi-project, but fails _is_context_complete()
selected_project_ids: [111, 222, 333, 444]
```

aBiR1D:
Fixes #856

Allow multi-project OAuth contexts to reuse cached tokens

Multi-project OAuth contexts store prod_environment as null by design, but _is_context_complete() required it, forcing a full browser login on every server restart despite valid cached tokens. Changed the check to accept either prod_environment (single-project) or selected_project_ids (multi-project).

Changes
src/dbt_mcp/config/credentials.py: relaxed _is_context_complete() to accept prod_environment OR selected_project_ids
tests/unit/oauth/test_token_refresh_at_startup.py: added test for multi-project context without prod_environment


DevonFulcher (approved):
Thank you!

DevonFulcher:
Can you please add a changelog entry. Thanks!

aBiR1D:
Done! Added the changelog entry in the latest commit. Thanks for the review!

jairus-m (approved):
Approving to see if that unblocks integration test error

PR #893, opened 2026-09-21
Experiment: semantic relevance filtering for list_metrics

b-per:
Following the `jev` hype of the past few days 	🤣

> **Draft / experiment.** Off by default; behaviour is unchanged unless explicitly enabled. Opening this for discussion of the approach rather than to merge as-is.

## What this does

Adds an optional `question` parameter to `list_metrics`. When set, the catalog is ranked by semantic relevance to the user's natural-language question, and the response contains only the best-matching metrics — with their full descriptions and a relevance score — plus the dimensions that matter for them.

Ranking uses [Jev](https://docs.typesafe.ai) (TypeSafe System One), a small model that returns typed judgments rather than text.

## Why

Two problems, measured on a production catalog of 425 metrics:

| Step | Cost |
|---|---|
| `list_metrics()` | ~7,600 tokens |
| `get_dimensions([metric])` | ~7,800 tokens, 134 dimensions |
| **Before the first query** | **~15,400 tokens, 2 round trips** |

1. **The size budget throws away the wrong thing.** `metrics_to_csv` drops `description` first when a listing is too large. On this catalog that means the agent receives 425 bare names — but descriptions are exactly what distinguishes near-duplicate metrics whose names differ only by a suffix (a combined total, an enterprise-only variant, a self-serve-only variant, an alternative calculation). Ranking trims the *number of items* instead and keeps every description.

2. **`search` is a substring match.** A question phrased in business vocabulary finds nothing: `search="product tier"` returns **zero** metrics on a catalog that has a subscription-tier dimension. Ranking found it at 0.90 — there is no substring to match, so no phrasing of `search` would have worked.

Because the survivors' dimensions come back in the same response, the follow-up `get_dimensions` call usually disappears too.

## Measured effect

| | Before | After |
|---|---|---|
| Tokens before first query | ~15,400 | ~840–1,500 |
| Round trips | 2 | 1 |
| Descriptions available | no | yes |

Roughly $0.027 per question on Sonnet 5, $0.070 on Opus 5 — conservative, since it counts the catalog payload once while an agent loop re-sends it every turn. Jev's own cost is ~$0.002 per question.

**Accuracy**, on the hardest case — six questions over a family of near-duplicate metrics distinguished only by description, scoring the rank of the true metric among all 425: **recall@5 6/6, recall@1 5/6**. Ablating descriptions drops recall@1 to 4/6, which is the direct argument against the current trimming order.

**Latency** is the honest weak spot. Ranking costs 1.7–2.6s. But it replaces model inference over ~14,000 more tokens plus an extra round trip, so it is not pure overhead — and **which is faster end to end has not been measured.** Settling that needs an A/B through a real agent loop measuring time to a correct answer. Treat the cost as known and the offset as plausible but unproven.

## Enabling it

Requires **both**, so that holding a TypeSafe key for unrelated reasons cannot silently change the tool schema or start sending catalog text off-host:

```bash
DBT_MCP_ENABLE_JEV=true
TYPESAFE_API_KEY=<key>
```

Install with `uv sync --extra jev`. When disabled, the `question` parameter is absent from the tool **schema**, not merely inert, and output is byte-identical to today's.

## Review notes

- `Noul` rather than `Choice`: `Choice` hard-caps at 255 options (verified: `400` at 256), which this catalog exceeds, and it forces exactly one winner while a question may need several metrics.
- Top-k with a low floor rather than an absolute cutoff — scores vary ±0.02 between runs, and a question the catalog cannot answer should return its closest candidates with visibly low scores rather than nothing.
- Every failure path (timeout, error, nothing above the floor) falls back to today's output. Ranking must never fail a tool call.
- Dimensions are fetched per metric because the API intersects them across metrics; `search` stays `None` so the question never enters the fetcher's cache key.
- A dimension block is emitted only for metrics within 80% of the top score — metrics in one family expose nearly the same dimensions, so a block for a weak runner-up is near-duplicate filler.

**Things worth pushing back on:**

- `typesafe-sdk` is added to the **dev group** (so the ranker is unit-testable offline) and given an `exclude-newer-package` exemption, since the validated version postdates the repo's dependency cutoff. That is a real change to the dependency posture for an experiment.
- `MetricToolResponse.relevance` is a new field on a type shared with the multi-project path. `None`-defaulted, so the CSV column is omitted when unused.
- **`tools_multiproject.py` is unchanged** — it has its own `list_metrics` with the same trimming logic.
- Data leaves the host when enabled: metric and dimension names, their descriptions, and the end user's question.

Metric and dimension names in the docs and tests are illustrative; the measurements are from a real catalog, the names are not.

## Testing

897 unit tests pass, `task check` green. 23 new tests cover ranking thresholds, cost/latency logging on success and failure, the two-part enablement gate, schema presence/absence of `question`, the zero-result widening, cache-key isolation, the dimension-block gate, and byte-identical fallback on every failure path.

Docs: `docs/jev-relevance-filtering.md`.


src/dbt_mcp/semantic_layer/param_descriptions.py
```diff
@@ -19,6 +19,18 @@
     "Filter saved queries by substring match on name, label, or description"
 )
 
+SEMANTIC_QUESTION = (
+    "The user's data question, in their own words. Pass it through verbatim - "
```
DevonFulcher, line 23: > Pass it through verbatim

Do you envision any issues with this? Often, users don't know the right questions to ask. Agents typically have greater context.

b-per: I don't know 🤣 

We could also try variations where we don't enforce that much the verbatim and actually recommend to the LLM to rephrase the original question.

To see the real impact we'd need some good set of evals thouhg.

